PR #7, opened 2024-05-03
bugfix and checks for making sure tokenizer is always padded left

ochougul:
* added fix to print stdout and stderr in case of compilation failure
* added checks for left padding of tokenizer
* updated notebooks, readme with left padding for tokenizer


anujgupt-github (approved):
LGTM. Linter warning seems to be a false alarm

PR #16, opened 2024-05-20
Enable arg trust_remote_code to use custom tokenizers

vbaddi:
Allow to use tokenizers for custom models defined on the Hub in their own modeling files.
Example: Salesforce/xgen-7b-8k-base


ochougul (approved):
Should we also add this to `QEfficient/exporter/export_hf_to_cloud_ai_100.py::line 72`

vbaddi:
> QEfficient/exporter/export_hf_to_cloud_ai_100.py::line 72

yes, nothing harm, fixed it.

PR #19, opened 2024-05-21
Add Changes for CustomOp to Enable LLMs on 1.14 SDK

vbaddi:
Couple of data points on this PR:

- Support to the run the LLMs with RMSNorm CustomOp on the 1.14 SDK.
- Use of regular CustomOp path, please refer the customop [readme ](https://github.com/quic/efficient-transformers/blob/release/v1.14/QEfficient/customop/README.md)to generate the shared library required to execute the model with regular customop on cloud ai 100.
- There will be no change in the `infer` API usage.
- The Onnxruntime execution will be skipped, due to regular customop. Since there is no implementation for customop in Onnxrt via efficient-transformers repo.
- **Sample Infer Test Command:** `python -m QEfficient.cloud.infer --model-name meta-llama/Llama-2-7b-hf  --cache-dir models --batch-size 1 --prompt-len 32 --ctx-len 128 --mxfp6 --num-cores 16 --device-group "[0]" --prompt "Top one must do thing in Bengaluru" --aic_enable_depth_first --mos 1 --hf-token ***********`


QEfficient/cloud/compile.py
```diff
@@ -62,10 +63,15 @@ def main(
         batch_size=batch_size, prompt_len=prompt_len, ctx_len=ctx_len, path=specialization_json_path
     )
     custom_io_file_path = os.path.join(os.path.dirname(onnx_path), "custom_io.yaml")
-    
+
     if not os.path.isfile(custom_io_file_path):
         raise FileNotFoundError(f"file {custom_io_file_path} needs to exist in the same directory as onnx model files.")
 
+    # todo: vbaddi: Identify better way to get the config path. fix it in tag/1.14 in main repo.
+    custom_rms_op_config_path = os.path.join(
```
ochougul, line 71:
```
from utils.constants import QEFF_DIR
custom_rms_config_path  = os.path.join(QEFF_DIR, "customop", "custom_rms_op_config.yaml")
```

QEfficient/cloud/compile.py
```diff
@@ -120,7 +127,8 @@ def main(
         help="Cloud AI 100 device ids (comma-separated) e.g. [0] ",
     )
     parser.add_argument(
-        "--aic_enable_depth_first", "--aic-enable-depth-first",
+        "--aic_enable_depth_first",
```
ochougul, line 130: Not sure if `mos` and `aic_enable_depth_first` in tested for 1.14?

QEfficient/cloud/compile.py
vbaddi: This looks better, Thanks. fixed it.

QEfficient/cloud/compile.py
vbaddi: Thats why default for this branch will be set to False.

ochougul: Okay, so with SDK 1.14, if I pass these options the command won't fail but might not behave correctly?
If the command will fail for SDK 1.14 we should just remove them? You can just write assert statement for it to fail in infer, compile APIs?

vbaddi: No, it won't fail. This is tested.

ochougul (approved):
LGTM

PR #23, opened 2024-05-24
Clean and Optimize Infer Api() code

vbaddi:
- Add individual methods to keep the code base neat
- Also easy for `Api(`) documentation creation.
- Add comments and missing docstring.
- If user intends to just `execute` from `infer`, without `cache-dir`, now he can do it, he will just need tokenizer to be downloaded and doesn't require entire `cache-dir` to be present and can be deleted once downloaded and exporting work is done.
- `[TODO]: Customers should be able to use single infer api to do --export-only, --compile-only`

### DIFF
--- a/QEfficient/cloud/infer.py
+++ b/QEfficient/cloud/infer.py
@@ -5,6 +5,13 @@
 #
 # -----------------------------------------------------------------------------
 
+"""
+1. Check if compiled qpc for given config already exists, if it does jump to execute, else
+2. Check if exported ONNX file already exists, if true, jump to compilation -> execution, else
+3. Check if HF model exists in cache, if true, start transform -> export -> compilation -> execution, else,
+4. Download HF model -> transform -> export -> compile -> execute
+"""
+
 import argparse
 import os
 from typing import List
@@ -21,14 +28,6 @@ from QEfficient.generation.text_generation_inference import (
 )
 from QEfficient.utils import hf_download, onnx_exists, qpc_exists
 from QEfficient.utils.constants import Constants
-from QEfficient.utils.logging_utils import logger
-
-"""
-1. Check if compiled qpc for given config already exists, if it does jump to execute, else
-2. Check if exported ONNX file already exists, if true, jump to compilation -> execution, else
-3. Check if HF model exists in cache, if true, start transform -> export -> compilation -> execution, else,
-4. Download HF model -> transform -> export -> compile -> execute
-"""
 
 
 def main(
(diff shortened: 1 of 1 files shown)


QEfficient/cloud/infer.py
```diff
@@ -45,102 +44,166 @@ def main(
...
-    )
-    tokenizer = AutoTokenizer.from_pretrained(
-        model_hf_path, use_cache=True, padding_side="left", trust_remote_code=True
-    )
 
+    # Check if pre-compiled QPC/Onnx exists and execute if so
     qpc_path_exists, qpc_dir_path = qpc_exists(model_name, qpc_base_dir_name)
+    onnx_path_exists, onnx_dir_path, onnx_model_path = onnx_exists(model_name)
+
     if qpc_path_exists:
```
ochougul, line 83: Why this condition to decide if we need to download the tokenizer files?
We always need tokenizer irrespective of the path of the code execution, i.e. for exec_kv function. Which will always be called.

ochougul (changes requested):
Can you please explain why is the need of creating one liner function wrappers?
If we use Low level APIs in this file, it provides user an example of how to use the LL APIs, are wrappers required?

QEfficient/cloud/infer.py
```diff
@@ -45,102 +44,166 @@ def main(
...
+    """
+    Create the base directory name for the QPC based on the parameters.
+
+    Returns:
+    - A string representing the base directory name.
+    """
+    return f"qpc_{num_cores}cores_{batch_size}BS_{prompt_len}PL_{ctx_len}CL_{mos}MOS_{device_count}devices{'_mxfp6_mxint8' if mxfp6 and mxint8 else '_mxfp6' if mxfp6 else '_fp16_mxint8' if mxint8 else '_fp16'}"
+
+
+def process_prompt(prompt: str, prompts_txt_file_path: str, batch_size: int) -> str:
```
ochougul, line 158: just one liner wrapper?

QEfficient/cloud/infer.py
```diff
@@ -45,102 +44,166 @@ def main(
...
+    """
+    Process the prompt for text generation.
+
+    Returns:
+    - The processed prompt.
+    """
+    return check_batch_size_and_num_prompts(prompt, prompts_txt_file_path, batch_size)
+
+
+def load_tokenizer(model_hf_path: str) -> AutoTokenizer:
```
ochougul, line 168: same

QEfficient/cloud/infer.py
```diff
@@ -45,102 +44,166 @@ def main(
...
+    """
+    Load the tokenizer from the downloaded model.
+
+    Returns:
+    - An instance of the tokenizer.
+    """
+    return AutoTokenizer.from_pretrained(model_hf_path, use_cache=True, padding_side="left", trust_remote_code=True)
+
+
+def execute_with_qpc(
```
ochougul, line 178: same?

QEfficient/cloud/infer.py
```diff
@@ -45,102 +44,166 @@ def main(
...
     )
-    assert (
-        generated_onnx_path == onnx_model_path
-    ), f"ONNX files were generated at an unusual location, expected {onnx_model_path}, got {generated_onnx_path}"
-    logger.info(f"Base Path is {base_path} and Onnx Model Path is : {generated_onnx_path}")
 
-    # Compile
-    # We need to pass parent directory of qpc_dir_path, as the compile function handles the qpcs directory creation
+
+def compile_and_execute_onnx(
```
ochougul, line 189: do you think this function will be used by other files in future?

QEfficient/cloud/infer.py
```diff
@@ -155,24 +218,43 @@ def main(
...
+        generated_qpc_path == qpc_dir_path
     ), f"QPC files were generated at an unusual location, expected {qpc_dir_path}; got {generated_qpc_path}"
-    logger.info(f"Compiled qpc files can be found at : {generated_qpc_path}")
+    execute_with_qpc(tokenizer, batch_size, qpc_dir_path, device_group, prompt)
 
-    # Execute
-    cloud_ai_100_exec_kv(
-        batch_size,
+
+def export_to_onnx(
```
ochougul, line 226: will this be used in future?

QEfficient/cloud/infer.py
vbaddi: Yes, just need tokenizer for the execution, so if user wants to delete cache and run the infer for execution, this will download tokenizer only and go ahead with execution.

QEfficient/cloud/infer.py
vbaddi: yes, if user wants to have just `onnx `or compilation from `infer`, he can do it. instead of rotating around compile/export/ apis()

QEfficient/cloud/infer.py
vbaddi: same

QEfficient/cloud/infer.py
vbaddi: yes, agree, `load_tokenizer`() should always be on.

QEfficient/cloud/infer.py
vbaddi: Just for neatness, keeping main simple and neat. (better readability)

QEfficient/cloud/infer.py
ochougul: what do you mean by "from `infer`"? Can we move these and `export_to_onnx` functions to `utils` file? and not keep it here? if it is going to be used by other files?

QEfficient/cloud/infer.py
vbaddi: yes, this works too.

vbaddi:
Issues addressed in #31, Closing this

PR #26, opened 2024-05-24
Add Json Based Pytest Pipeline for Scalability

vbaddi:
- Json based test pipeline, tested for gpt2 and codegen model categories.
- `TODO:` Look for ways to keep config params constants to scale it to other/any model category.
- Also change the PL and CL to 8 and 32 resp. for CPU execution to avoid long runs.  
- Sample: `pytest -v tests/transformers/test_models.py -s`


tests/transformers/models/test_models.py
```diff
@@ -0,0 +1,93 @@
+# -----------------------------------------------------------------------------
+#
```
anujgupt-github, line 2: This test infra will work for Causal LM models only.
Better to rename the file to reflect that

tests/transformers/models/test_models.py
```diff
@@ -0,0 +1,93 @@
...
+        d_head = model_config["d_model"] // n_heads
+        model_config["model_class"] = AutoModelForCausalLM
+        model_config["padding_shape"] = [1, n_heads, Constants.CTX_LEN, d_head]
+    else:
+        raise ValueError("Invalid model configuration: n_head/n_heads or num_key_value_heads not found.")
+
+    return model_config
+
+
+class TestQEfficientModels(unittest.TestCase):
```
anujgupt-github, line 47: Pls add docstring stylized headers, for inclusion in documentation

tests/config.json
```diff
@@ -0,0 +1,33 @@
+{
+  "models": [
+    {
+      "model_name": "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
+      "model_class": "LlamaForCausalLM",
+      "num_key_value_heads": 4,
+      "num_attention_heads": 32,
+      "hidden_size": 2048,
+      "n_layer": 22
```
anujgupt-github, line 9: while loading tests, and downloading model based on model card, can we not fetch the rest of the config elements from the json?

vbaddi: yes, was thinking about, fixed it. only requirement would be `model_name `and class in the json for new test addition now unless there is new config params.

quic-mamta (approved):
LGTM

PR #28, opened 2024-05-28
Update text generation interface to remove slash from qpc path

quic-mamta:
Update text generation interface to remove slash("/') from qpc path

'/' in the end after qpcs makes execute API look for specializations.json in wrong folder.
 
This works:

python -m QEfficient.cloud.execute --model_name gpt2 --qpc_path qeff_models/gpt2/qpc_16cores_1BS_32PL_128CL_1devices_mxfp6/qpcs --prompt "Once upon a time in" --device_group [0]  
 
This fails:

python -m QEfficient.cloud.execute --model_name gpt2 --qpc_path qeff_models/gpt2/qpc_16cores_1BS_32PL_128CL_1devices_mxfp6/qpcs/ --prompt "Once upon a time in" --device_group [0]

### DIFF
--- a/QEfficient/generation/text_generation_inference.py
+++ b/QEfficient/generation/text_generation_inference.py
@@ -98,7 +98,10 @@ def latency_stats_bertstyle(
 
 
 def get_compilation_batch_size(qpc_path: str):
-    qpc_base_path = os.path.dirname(qpc_path.rstrip("/"))
+    if qpc_path.endswith("/"):
+        logger.warning("Removing the slash('/')from given qpc path and looking for specializations.json in parent folder")
+        qpc_path = qpc_path.rstrip("/")
+    qpc_base_path = os.path.dirname(qpc_path)
     specialization_file_path = os.path.join(qpc_base_path, "specializations.json")
     with open(specialization_file_path, "r") as file:
         data = json.load(file)
--- a/QEfficient/generation/text_generation_inference.py
+++ b/QEfficient/generation/text_generation_inference.py
@@ -98,11 +98,9 @@ def latency_stats_bertstyle(
 
 
 def get_compilation_batch_size(qpc_path: str):
-    if qpc_path.endswith("/"):
-        logger.warning("Removing the slash('/')from given qpc path and looking for specializations.json in parent folder")
-        qpc_path = qpc_path.rstrip("/")
-    qpc_base_path = os.path.dirname(qpc_path)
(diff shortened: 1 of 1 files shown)


QEfficient/generation/text_generation_inference.py
```diff
@@ -98,7 +98,7 @@ def latency_stats_bertstyle(
 
 
 def get_compilation_batch_size(qpc_path: str):
-    qpc_base_path = os.path.dirname(qpc_path)
+    qpc_base_path = os.path.dirname(qpc_path.rstrip("/"))
```
vbaddi, line 101: Now, does it work for both with "/" and without?

anujgupt-github:
Put as a warning also, so that the user knows which specialization is being picked. 
What if user wanted the tool to pick up the one after '/' (user has 2 specialization files)?

QEfficient/generation/text_generation_inference.py
quic-mamta: yes

QEfficient/generation/text_generation_inference.py
```diff
@@ -98,6 +98,9 @@ def latency_stats_bertstyle(
 
 
 def get_compilation_batch_size(qpc_path: str):
+    if qpc_path.endswith("/"):
+        logger.warning("Removing the slash('/')from given qpc path and looking for specializations.json in parent folder")
+        qpc_path = qpc_path.rstrip("/")
     qpc_base_path = os.path.dirname(qpc_path)
```
vbaddi, line 104: Instead of this, can we just do: `qpc_base_path = os.path.dirname(os.path.normpath(qpc_path))` works for "/" and without. test it once pls

quic-mamta: yes, it works fine

quic-mamta:
> Put as a warning also, so that the user knows which specialization is being picked. What if user wanted the tool to pick up the one after '/' (user has 2 specialization files)?

if model is compiled with compile low level API or ran using infer API, then qpc and speializations.json are stored in following directory structure:

qeff_models/gpt2/qpc_16cores_1BS_32PL_128CL_-1MOS_1devices_fp16/qpcs/programqpc.bin
qeff_models/gpt2/qpc_16cores_1BS_32PL_128CL_-1MOS_1devices_fp16/specializations.bin

so there will be just one specializations,json

vbaddi (approved):
LGTM

vbaddi (approved):

ochougul (approved):

PR #37, opened 2024-06-05
Fixed input_len to work with prompt chunking

quic-morteza:
Without the change, prompt chunking doesn't work and there is a shape error for runs like: python -m QEfficient.cloud.infer --model_name gpt2 --prompt_len 2 --prompt "Hi, my name is Morteza and"

With the new change, the correct input_len is picked up from the inserted prompts.


quic-morteza:
> Do we have any Unit test with chunking? If not, @quic-morteza can you add a unit test case?
Sure.

Before - 
before

After -
after

The change also takes care of the max length in the batch mode
after2

PR #48, opened 2024-06-20
enabling `export_and_compile` for `QEFFAutoModelForCausalLM`

ochougul:
* Introduced `Runtime`(`CPU_PT`, `CPU_ORT`, `AI_100`)
* enabled export, compile and export_and_compile for `QEFFAutoModelForCausalLM`


QEfficient/src/common.py
```diff
@@ -83,4 +83,4 @@ def from_pretrained(cls, pretrained_model_name_or_path: str, *args, **kwargs) ->
         qeff_auto_model_class = MODEL_TYPE_TO_QEFF_AUTO_MODEL_MAP[model_type]
         assert issubclass(qeff_auto_model_class, QEFFBaseModel), f"Expected class that inherits {QEFFBaseModel}, got {type(qeff_auto_model_class)}"
 
-        return qeff_auto_model_class.from_pretrained(pretrained_model_name_or_path=pretrained_model_name_or_path)
+        return qeff_auto_model_class.from_pretrained(pretrained_model_name_or_path=pretrained_model_name_or_path, **kwargs)
```
vbaddi, line 86: Do we want to pass `**kwargs` to the `AutoModelForCausalLM `__class__, might lead to hf_token issue in the hugginface.

QEfficient/utils/_utils.py
```diff
@@ -119,13 +119,13 @@ def onnx_exists(model_name: str) -> Tuple[bool, str, str]:
     return onnx_exists_bool, onnx_dir_path, onnx_model_path
 
 
-def load_hf_tokenizer(model_name: str, cache_dir: Optional[str] = None, hf_token: Optional[str] = None, padding_side:str = "right", **kwargs) -> Union[PreTrainedTokenizerFast, PreTrainedTokenizer]:
-    logger.info(f"Loading Tokenizer for {model_name}")
+def load_hf_tokenizer(pretrained_model_name_or_path: str, cache_dir: Optional[str] = None, hf_token: Optional[str] = None, padding_side:str = "left", **kwargs) -> Union[PreTrainedTokenizerFast, PreTrainedTokenizer]:
```
vbaddi, line 122: okay to keep default padding to "`right`"?

QEfficient/utils/_utils.py
```diff
@@ -119,13 +119,13 @@ def onnx_exists(model_name: str) -> Tuple[bool, str, str]:
...
+    logger.info(f"Loading Tokenizer for {pretrained_model_name_or_path}")
     if hf_token is not None:
         login(hf_token)
 
     # Download tokenizer along with model if it doesn't exist
-    model_hf_path = hf_download(repo_id=model_name, cache_dir=cache_dir, allow_patterns=["*.json", "*.py", "*token*", "*.txt"])
+    model_hf_path = pretrained_model_name_or_path if os.path.isdir(pretrained_model_name_or_path) else hf_download(repo_id=pretrained_model_name_or_path, cache_dir=cache_dir, allow_patterns=["*.json", "*.py", "*token*"])
```
vbaddi, line 128: Need `*.txt`, to get tokenizers from models like `GPTJ `which are not in json formats.

QEfficient/exporter/export_hf_to_cloud_ai_100.py
```diff
@@ -383,6 +384,8 @@ def export_lm_model_for_cloud(model_name:str, qeff_model: QEFFAutoModelForCausal
             onnx_dir_path=onnx_dir_path,
             seq_len=seq_length) # type: ignore
 
+    cpu_ort_args = QEFFAutoModelForCausalLMCPUORTRuntimeArgs(onnx_model_path=os.path.join(onnx_dir_path, f"{fp16_model_name}.onnx"))
+    qeff_model.set_runtime(runtime=Runtime.CPU_ORT, runtime_args=cpu_ort_args)
```
vbaddi, line 388: lets discuss this `runtime `inclusion once.

QEfficient/utils/_utils.py
ochougul: will update, these changes came in, as I was reviving this old branch, and main was too ahead of this.

QEfficient/utils/_utils.py
ochougul: will update. Thanks.

QEfficient/src/common.py
ochougul: Good catch. Fixed

QEfficient/src/_transformers/runtime_args.py
```diff
@@ -0,0 +1,33 @@
...
+
+import os
+from dataclasses import dataclass, field
+from typing import List
+
+from QEfficient.utils.constants import Constants
+
+
+@dataclass
+class QEFFAutoModelForCausalLMAI100RuntimeArgs:
```
irajagop, line 16: Do we need a new dataclass for runtime args? A simple dict won't work?

ochougul: 1. Dataclasses automatically does type restriction of attributes
2. We are doing some check on them in __post_init__ which we will  have to write everywhere they are being initialized if we use dicts
3. Also, I am thinking of freezing this dataclasses so that, user can't assign arbitrary values to them
4. So, because of 2 and 3 we are always sure that once runtime-args are populated we can always allowed to execute in respective runtime, making it robust from failures.

I understand that dicts are more efficient for initialization and access, but I feel the gain is negligible as compared to the above advantages we are getting by using dataclass

ref: 
https://stackoverflow.com/questions/55248416/python-3-which-one-is-faster-for-accessing-data-dataclasses-or-dictionaries
https://stackoverflow.com/questions/74117873/data-classes-vs-dictionaries

QEfficient/src/_transformers/auto.py
```diff
@@ -74,11 +100,46 @@ class QEFFAutoModelForCausalLM(QEFFTransformersBase):
...
+    def set_runtime(self, runtime: Runtime, runtime_args: Union[QEFFAutoModelForCausalLMAI100RuntimeArgs, QEFFAutoModelForCausalLMCPUORTRuntimeArgs]) -> None:
+        assert runtime != Runtime.CPU_PT, "Please use self.runtime=Runtime.CPU_PT for updating runtime to CPU_PT"
+
+        if runtime == Runtime.CPU_ORT:
+            assert isinstance(runtime_args, QEFFAutoModelForCausalLMCPUORTRuntimeArgs), f"Expected runtime_args of type {QEFFAutoModelForCausalLMCPUORTRuntimeArgs.__class__}, got {type(runtime_args)}"
+            self.ort_runtime_args = runtime_args
+        elif runtime == Runtime.AI_100:
+            assert isinstance(runtime_args, QEFFAutoModelForCausalLMAI100RuntimeArgs),  f"Expected runtime_args of type {QEFFAutoModelForCausalLMAI100RuntimeArgs.__class__}, got {type(runtime_args)}"
+            self.cloud_ai_100_runtime_args = runtime_args
+        self.runtime = runtime
```
irajagop, line 112: Why do we need to have only one runtime at a time? We could have something like `execute_pytorch`, `execute_onnxrt` and `execute_aic`. We could still have `self.runtime` set to something as default, that will direct which `execute_` method to be called.

irajagop: Also, we could manage the args internally no? For example, after export, we could directly update the `self.onnx_model_path` instead of needing to set it from outside.

QEfficient/src/_transformers/runtime_args.py
irajagop: As suggested above, if we use `execute_` method with correct args or kwargs, that will work as validation. We're going to use it at only one place, for execution. `dataclass` makes more sense when we're re-using in multiple places, serializing or sending over network etc.

QEfficient/src/_transformers/auto.py
ochougul: If we expose `self.onnx_model_path` as model variable we allow users to set it any value they want, we can pass `forzen=True` in `dataclasses` to not allow such usage.

README.md
```diff
@@ -223,60 +223,38 @@ qeff_model = AutoModelForCausalLM.from_pretrained(model_name)
...
 
 ```Python
-import QEfficient
-from QEfficient.utils import load_hf_tokenizer
 # We can now export the modified models to Onnx framework
 # This will generate single Onnx Model for both Prefill and Decode Variations which are optimized for
 # Cloud AI 100 Platform.
 
-# This will generate Onnx model, clip the overflow constants to fp16
+# While generaing the onnx model, this will clip the overflow constants to fp16
```
irajagop, line 235: typo

ochougul: updated.

QEfficient/generation/text_generation_inference.py
```diff
@@ -139,8 +139,8 @@ def cloud_ai_100_exec_kv_helper(
     tokenizer: Union[PreTrainedTokenizer, PreTrainedTokenizerFast],
     qpc: str,
     prompt: List[str],
-    ctx_len: Optional[int] = None,
-    generation_len: Optional[int] = None,
+    ctx_len: int,
+    generation_len: int = None,
```
irajagop, line 143: `Optional[int]` should be there for generation_len.

irajagop (approved):

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -16,6 +15,8 @@
 from QEfficient.utils.device_utils import get_available_device_id
 from tests.utils import get_cloud_ai_100_tokens, set_up
 
+TEST_CONFIG_FILE_PATH = os.path.join(ROOT_DIR, "tests", "config.json")
```
vbaddi, line 18: Can you remove the `phi3` model card from the `config.json`, to test the CI pipeline? todo @quic-mamta for phi3 modeling debug.

vbaddi: this is taken care in #63. Thanks

vbaddi (approved):
LGTM, Once CI Passes, pls merge. Thanks

PR #49, opened 2024-06-21
Use ctx_len from specializations.json

vbaddi:
- Fix the `Prefill `aka `TTFT `time in the generation script.
- Pass `ctx_len `as input to the generation api. (to address the `past_key.0`) issues in MQ runs, use `session.allowed_shapes api` to capture the batch_size.
- modify `get_compilation_batch_size` api to return `ctx_len `along with `batch_size`, rename to `get_compilation_dims`


QEfficient/cloud/execute.py
```diff
@@ -48,6 +48,7 @@ def main(
         qpc_path=qpc_path,
         device_id=device_group,
         prompt=prompt,
+        generation_len=ctx_len,
```
irajagop, line 51: generation_len and ctx_len are different. In the latest runner, we're using generation_len = ctx_len - input_len, but not reflected here.

Ideal way should be to add ctx_len also as parameter and pass there.

QEfficient/cloud/execute.py
```diff
@@ -38,7 +38,7 @@ def main(
 
     tokenizer = load_hf_tokenizer(model_name, cache_dir, hf_token)
 
-    batch_size = get_compilation_batch_size(qpc_path)
+    batch_size, ctx_len = get_compilation_batch_size(qpc_path)
```
irajagop, line 41: Since we're returning multiple results and not just batch_size, we should rename this to something generic like `get_compilation_dims`

QEfficient/generation/text_generation_inference.py
```diff
@@ -171,11 +172,11 @@ def cloud_ai_100_exec_kv_helper(
     inputs = tokenizer(prompt, return_tensors="np", padding=True)
     position_ids_update = inputs["attention_mask"].sum(1, keepdims=True)
     if generation_len is None:
-        generation_len = ctx_len
+        assert generation_len > 0, "generation length should be greater than zero, please pass valid generation length"
```
irajagop, line 175: Here, we should be using generation_len = ctx_len - input_len

QEfficient/generation/text_generation_inference.py
```diff
@@ -157,8 +158,8 @@ def cloud_ai_100_exec_kv_helper(
     # Skip inputs/outputs
     session.skip_buffers([x for x in session.input_names + session.output_names if x.startswith("past_")])
 
-    # Read prompt and ctx len from session
-    batch_size, _, ctx_len, _ = session.bindings[session.binding_index_map["past_key.0"]].dims
+    # Read prompt and batch_size from session
+    batch_size, _ = session.bindings[session.binding_index_map["input_ids"]].dims
```
irajagop, line 162: This may fail to get proper batch_size, since input_ids is specialized, we use the allowed_shapes instead as mentioned below for prefill_seq_len. Will try to fix this in a new commit.

irajagop (changes requested):

QEfficient/generation/text_generation_inference.py
irajagop: The logic is incorrect here. This assertion needs to be done `if generation_len is not None`, but it's done `if generation_len is None`.

QEfficient/cloud/execute.py
vbaddi: Agree.

irajagop (approved):

QEfficient/cloud/execute.py
```diff
@@ -48,6 +48,7 @@ def main(
         qpc_path=qpc_path,
         device_id=device_group,
         prompt=prompt,
+        ctx_len=ctx_len,
```
quic-mamta, line 51: we don't need to pass this, as we can get this info from session.

quic-mamta (approved):

README.md
```diff
@@ -270,12 +271,12 @@ generated_qpc_path = QEfficient.compile(
...
 
 ```Python
-from QEfficient.generation.text_generation_inference import get_compilation_batch_size
+from QEfficient.generation.text_generation_inference import get_compilation_dims
 
 # post compilation, we can print the latency stats for the kv models, We provide API to print token and Latency stats on AI 100
 # We need the compiled prefill and decode qpc to compute the token generated, This is based on Greedy Sampling Approach
-batch_size = get_compilation_batch_size(generated_qpc_path)
-QEfficient.cloud_ai_100_exec_kv(batch_size=batch_size, tokenizer=tokenizer, qpc_path=generated_qpc_path, device_id=[0], prompt=["My name is"])
+batch_size, ctx_len = get_compilation_dims(generated_qpc_path)
```
ochougul, line 278: can we move this call inside `cloud_ai_100_exec_kv`?
so, we only ask users for `tokenizer`, `qpc_path` and `device_id`

ochougul (changes requested):

README.md
quic-mamta: If we move this inside cloud_ai_100_exec_kv then we need to move the check_num_prompts_and_batch_size also inside cloud_ai_100_exec_kv, (for execute API)

Then in case of infer API, it will cause check_num_prompts_and_batch_size being called twice, if we remove the first call of this function from infer API then it will export and compile but fail at last step of execution if there is mismatch in num_prompts and batch_size.

This is what we discussed last time.

ochougul: I don't understand, why do you say we need to move `check_num_prompts_and_batch_size` for this?

quic-mamta: because input of this function is compilation_batch_size which is output of get_compilation_dims

ochougul: As discussed, let's separate 
1. checking `batch_size` with `num_of_prompts` 
2. figuring out prompts from `prompt` and `prompt_txt_file`

So, that we can call `func` 1 twice, and it won't lead to increased runtime, and we can only expose users to the HL API which is exec as explained in first comment.

quic-mamta: Added the suggensted changes to PR#47.

QEfficient/generation/text_generation_inference.py
```diff
@@ -157,24 +158,26 @@ def cloud_ai_100_exec_kv_helper(
...
     position_ids_update = inputs["attention_mask"].sum(1, keepdims=True)
-    if generation_len is None:
-        generation_len = ctx_len
     padded_len = inputs["input_ids"].shape[1]
     num_chunks = -(padded_len // -prefill_seq_len)  # ceil divide without float
     padded_len = num_chunks * prefill_seq_len  # Convert to a multiple of prompt_len
+    if generation_len is None:
+        if ctx_len is None:
+            raise ValueError("At least one of ctx_len or generation_len is needed")
+        generation_len = ctx_len - position_ids_update.max()
```
ochougul, line 180: Shouldn't we also check when `generation_len` is provided by user,
`asser generation_len> ctx_len-position_ids_update.max()` ?

irajagop: That case is captured in the if condition lineno:177 as above. We compute generation length only if user has not provided it.

irajagop: If there is space for 200 tokens, but the user wants to generate only 20 tokens, then the above assertion doesn't make sense. We only assert that the user wants to generate at least 1 token, which is what we're already checking, in the next line (181).

ochougul: As discussed, let's put a warning for this case.
And should we expose this `generation_len` variable for `infer`, `execute` APIs?

irajagop: I've added a warning for the generation_len. Also exposed `--generation_len`, in infer and execute APIs.

ochougul (approved):
Please verify before merging.
1. notebooks
2. `--generation_len` variable via infer/execute

Everything else LGTM.

PR #54, opened 2024-06-27
Use eager mode attention implementation in Infer API

quic-mamta:
Update from_pretrained method to always use eager mode attention implementation
Fixed number of layers in falcon transformed config

### DIFF
--- a/QEfficient/exporter/export_hf_to_cloud_ai_100.py
+++ b/QEfficient/exporter/export_hf_to_cloud_ai_100.py
@@ -1,6 +1,6 @@
 # -----------------------------------------------------------------------------
 #
-# Copyright (c)  2023-2024 Qualcomm Innovation Center, Inc. All rights reserved.
+# Copyright (c) 2024 Qualcomm Innovation Center, Inc. All rights reserved.
 # SPDX-License-Identifier: BSD-3-Clause
 #
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


vbaddi:
cc @irajagop

vbaddi (approved):
LGTM

irajagop (approved):

QEfficient/exporter/export_hf_to_cloud_ai_100.py
```diff
@@ -212,12 +220,10 @@ def export_kvstyle_transformed_model_to_onnx(model_name: str,
         multi_query_value = getattr(config, "multi_query")
         if multi_query_value:
             n_heads = 1  # MQA
-            d_head = config.hidden_size // config.num_attention_heads
-            n_layer = 1  # Due to multi query
```
vbaddi, line 216: can we just change this to `n_layer=config.num_hidden_layers`? for MQA

vbaddi (changes requested):

QEfficient/exporter/export_hf_to_cloud_ai_100.py
quic-mamta: Only n_heads is different in case of MQA which is inside if else condition, rest 2 lines are common so i did this to avoid duplication.

irajagop (approved):

vbaddi (approved):
LGTM, thanks

PR #55, opened 2024-06-27
Refactor custom ops into proper file locations

irajagop:
Create proper files for custom ops instead of putting them all in `__init__.py`. Added import in `__init__.py` so existing code is not affected.


QEfficient/customop/rms_norm.py
```diff
@@ -1,61 +1,51 @@
...
-    def __init__(self, normalized_shape, eps=1e-5, device=None, dtype=None, **kwargs):
-        super().__init__()
-        self.weight = torch.nn.Parameter(torch.empty(normalized_shape, device=device, dtype=dtype))
-        self.eps = eps
+from torch import nn
+
+ops = onnxscript.opset13
+
+
+@onnxscript.script(onnxscript.values.Opset(domain="com.qti.aisw.onnx", version=1))
```
ochougul, line 15: why are we using different domain here?

QEfficient/customop/ctx_scatter_gather.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@onnxscript.script(onnxscript.values.Opset("com.qualcomm.cloud", 1))
+def CtxGather(data: onnxscript.FLOAT, ctx_indices: onnxscript.INT32) -> onnxscript.FLOAT:
+    ctx_indices = ops.Expand(ctx_indices, ops.Slice(ops.Shape(data), starts=[0], ends=[3], axes=[0]))
+    ctx_indices = ops.Unsqueeze(ctx_indices, [-1])
+    return ops.GatherND(data, ctx_indices, batch_dims=2)
+
+
+class CtxGatherFunc(torch.autograd.Function):
```
ochougul, line 64: Should we create a `pytorch` module and apply this in the forward method similar to `RMSNorm`.
And use the `pytorch` module in `Cache_utils.py` ?
Is that the standard practice?

QEfficient/customop/rms_norm.py
ochougul: Oh okay! cool.

QEfficient/customop/ctx_scatter_gather.py
```diff
@@ -0,0 +1,81 @@
...
+#
+# Copyright (c) 2023-2024 Qualcomm Innovation Center, Inc. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# -----------------------------------------------------------------------------
+
+import onnxscript
+import torch
+
+ops = onnxscript.opset13
```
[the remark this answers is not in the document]
irajagop: Currently I'm not changing the opset, as it's not relevant here. We can plan for upgrade of opset. We definitely have to update the opset version for finetuning. One path we could take is to use different opset versions for inference and finetuning.

QEfficient/customop/ctx_scatter_gather.py
irajagop: For RMSNorm, we're replacing the whole RMSNorm torch module with our version that applies the custom op. RMSNorm is a replacement module in the mapping. For CtxGather and CtxScatter, there is no such requirement. And having this way is simpler.

vbaddi (approved):
LGTM. Thanks

ochougul (approved):

ochougul:
Please rebase with main.

PR #58, opened 2024-07-01
Continuous batching

quic-rishinr:
Currently we can run models with continuous batching and non-continuous batching with these changes. 
- For enabling continuous batching pass --full_batch_size in infer API. By default infer will get executed in Non continuous batching mode. 
- --full_batch_size with batchsize > 1 is not supported. 
Refactored text_generation_inference.py module for switching between continuous and non-continuous batch execution. 
Added continuous batching modeling changes for Llama, Mistral and Mixtral models. 

Identified enhancements or issues
1. <Important> There is some additional performance improvement observed in the latency calculation. Need some debugging.  
2.  Code is failing if the infer is called without FBS and then with FBS. Need to update onnx path with FBS
3.  Check if ctx is none and fetch if its None
4. <Future improvement> Define a data class for storing the input and output values. 
5.  Testing is ongoing for Mixtral in continuous batching mode. (Havent added Mixtral in get_lists_of_cb_qeff_models in modeling utils)


ochougul:
Closing as being addressed in #73

PR #68, opened 2024-07-10
Fix packaging

irajagop:
- Fixed setuptools not including subpackages by default
- Removed setup.py for reading version using regexp
- Moved to setuptools native dynamic attribute
- Added missing `__init__.py` files
- Install QEfficient as package instead of editable install in CI


pyproject.toml
```diff
@@ -42,8 +42,9 @@ quality = ["black", "ruff", "hf_doc_builder@git+https://github.com/huggingface/d
 requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
 
-[tool.setuptools]
-packages = ["QEfficient"]
+[tool.setuptools.packages.find]
```
ochougul, line 45: What is the name of the package that get's installed? Is it `QEfficient`?
The name that comes up in `pip3 freeze` ?

vbaddi:
@irajagop need a PR directing to the release/v1.16 branch too.

cc @anujgupt-github @quic-rishinr

pyproject.toml
irajagop: Yeah, the name QEfficient comes up in the `pip freeze`. The name listed there is not related here. It will use the `project.name` value in the file.

carlstreeter-quic (approved):
tested in qaic-docker

setup.py
```diff
@@ -1,17 +0,0 @@
-# -----------------------------------------------------------------------------
-#
```
vbaddi, line 2: Can we keep this to have backward compatibility going? `pip install -e .` or `python setup.py install`

tests/cloud/conftest.py
```diff
@@ -13,7 +13,7 @@
 
 from QEfficient.generation.text_generation_inference import check_batch_size_and_num_prompts
 from QEfficient.utils import get_qpc_dir_path
-from QEfficient.utils.constants import QEFF_MODELS_DIR, ROOT_DIR, Constants
```
vbaddi, line 16: What would be defult location to the tests files then? if not $PWD of package in the Jenkins.
Maybe we export the env QEFF_HOME in Jenkins and then this makes sense.

vbaddi (changes requested):

setup.py
irajagop: With this minimal setup.py, it installs with package name "UNKNOWN". Otherwise, we have to repeat all the info in `pyproject.toml` in this file also. As per: [setup.py deprecation](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/), it's not recommended to run `python setup.py` commands anymore. Not having setup.py also ensures this.

tests/cloud/conftest.py
irajagop: These lines are removed because the test files were located using the ROOT_DIR. tests are not included in the QEfficient package, hence they'll be pointing to incorrect path.

While calling QEfficient methods from test, the ROOT_DIR is still used in the package code, so it will be continued to be used.

setup.py
vbaddi: removing setup.py, doesn't ensure `pip install -e .` works right?

irajagop: setup.py has nothing to do with pip. Even after removing setup.py, I'm able to do editable install with `pip install -e .`.

vbaddi: Okay, i was facing the issue to do it, let me retest it once.

irajagop: I've added minimum version for setuptools to support pyproject.toml only project. `pip install .` or `pip install git+<repo_url>` should work without issues when pip is reasonably up-to-date.

vbaddi (approved):
LGTM, Thanks

PR #69, opened 2024-07-11
fix setup issues for the release/v1.16 branch

vbaddi:
- Fix the setup issues, to be inline with the deployment.
- Over-ride the models/qpc directory using the env variable. -- `export QEFF_HOME`


vbaddi:
cc @quic-rishinr @carlstreeter-quic @quic-jhugo

carlstreeter-quic (approved):
tested in qaic-docker

QEfficient/utils/constants.py
```diff
@@ -10,7 +10,41 @@
...
+
+    Sets QEFF_MODELS_DIR environment variable if not already set.
+
+    Returns:
+        str: Path to the QEFF models directory.
+    """
+    qeff_cache_home = os.environ.get("QEFF_HOME")
+    # Check if XDG_CACHE_HOME is set
+    xdg_cache_home = os.environ.get("XDG_CACHE_HOME")
+    if xdg_cache_home:
```
irajagop, line 33: It should be other way around, first check for QEFF_HOME, only if not present you have to check for XDG_CACHE_HOME.

QEfficient/utils/constants.py
```diff
@@ -10,7 +10,41 @@
...
+        qeff_models_dir = os.path.join(xdg_cache_home, "qeff_models")
+    # Check if QEFF_MODELS_DIR is set
+    elif qeff_cache_home:
+        qeff_models_dir = os.path.join(qeff_cache_home, "qeff_models")
```
irajagop, line 37: We should directly assign `qeff_models_dir = qeff_cache_home`

QEfficient/utils/constants.py
```diff
@@ -10,7 +10,41 @@
...
+    else:
+        # Use ~/.cache/qeff_models as the default
+        qeff_models_dir = os.path.join(os.path.expanduser("~"), ".cache", "qeff_models")
+
+    # Set QEFF_MODELS_DIR environment variable
+    os.environ["QEFF_HOME"] = qeff_models_dir
```
irajagop, line 43: We shouldn't set the environment variable. This is available only in python env, and will be reset when python exits.

QEfficient/utils/constants.py
```diff
@@ -10,7 +10,41 @@
 UTILS_DIR = os.path.dirname(os.path.abspath(__file__))
 QEFF_DIR = os.path.dirname(UTILS_DIR)
 ROOT_DIR = os.path.dirname(QEFF_DIR)
```
irajagop, line 12: We should remove these too, as these again create files in the installation path.

QEfficient/utils/constants.py
vbaddi: Done

QEfficient/utils/constants.py
vbaddi: Agree, Removed

QEfficient/utils/constants.py
vbaddi: There is some dependency on other scripts for ROOT_DIR, would like to have this until fully tested.

QEfficient/utils/constants.py
vbaddi: The default path for all file save should be `qeff_models`, along with QEFF_HOME/XDG_CACHE_HOME/~.cache.

QEfficient/utils/constants.py
vbaddi: as mentioned in #68 same goes here too, this requires changes in many places making the review difficult, will take this up once this is closed. thanks, resolving the chat.

carlstreeter-quic (approved):
Tested QEfficient.cloud.infer in docker container with locked down venv, and that works now, too.

vbaddi:
Priority for saving the export onnx/qpc paths:

- Use QEFF_HOME if set in environment.
- Use $XDG_CACHE_HOME/qeff_models if XDG_CACHE_HOME is set.
- Default to ~/.cache/qeff_models.

PR #71, opened 2024-07-15
Returning past_key_values after scattering as output

quic-mamta:
Fixing the issue in some of the models with range-based gather. 

We should be returning the past key values after scattering as output. The gathered values should used for only that iteration. The output of the model should be directly after scattering.
 
If we return the cache after gathering, there will be performance hit.


irajagop (approved):
LGTM. I hope other models are also verified for the same.

vbaddi:
Do we need to change for `llama/mistral/starcoder2` or they are handled already?

irajagop:
> Do we need to change for `llama/mistral/starcoder2` or they are handled already?

Models that use `cache_utils.DynamicCache` directly already have this handled. The above models use DynamicCache, hence the fix is not needed.

ochougul (approved):

PR #73, opened 2024-07-16
Continuous batching 

quic-rishinr:
- Added support for transforming, compiling, and executing models using continuous batching (CB) mode.
- Supported architectures for CB execution include Llama, Mistral, Mixtral, and Starcoder.
- Integrated continuous batching transform under PyTorch transform.
- Added support for setting QEFF_HOME and XDG_CACHE_HOME directory paths.
- Refactored text_generation_inference module to accommodate CB mode execution and enable switching between regular and CB execution.
- Updated README and documentation to reflect the new changes.
- Added tests for continuous batching mode.

PR #75, opened 2024-07-17
Existing qpc and onnx file check

quic-amitraj:
Add qpc and onnx file check in low level APIs- export, compile and generate.

### DIFF
--- a/QEfficient/compile/compile_helper.py
+++ b/QEfficient/compile/compile_helper.py
@@ -11,6 +11,7 @@ import shutil
 import subprocess
 from typing import List, Tuple
 
+from QEfficient.utils import qpc_exists
 from QEfficient.utils.logging_utils import logger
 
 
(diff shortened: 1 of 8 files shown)


QEfficient/compile/compile_helper.py
```diff
@@ -127,41 +128,43 @@ def compile(
     :mxfp6: bool. Enable compilation for MXFP6 precision
     :mxint8: Compress Present/Past KV to MXINT8 using CustomIO config, default is False.
     """
-    if qpc_exists(qpc_path+"/qpcs"):
```
vbaddi: use `os.path.join()`

QEfficient/exporter/export_hf_to_cloud_ai_100.py
```diff
@@ -443,7 +443,7 @@ def qualcomm_efficient_converter(
     #Checking if onnx file already exist.
     if os.path.isfile(onnx_model_path) and os.path.isfile(os.path.join(os.path.dirname(onnx_model_path), "custom_io_fp16.yaml")):
         print("Pre-exported ONNX files found at {onnx_dir_path}")
-        return onnx_dir_path, onnx_model_path
+        return onnx_dir_path, onnx_mpwdodel_path
```
vbaddi, line 448: nit: `onnx_mpwdodel_path`

QEfficient/generation/text_generation_inference.py
```diff
@@ -290,6 +290,10 @@ def cloud_ai_100_exec_kv(
     write_io_dir: Optional[str] = None,
     automation=False,
 ):
+    #Check of qpc file exist on path
+    if not qpc_exists(qpc_path): 
+        raise FileNotFoundError("The specified qpc file not found at {qpc_path}.")
```
vbaddi, line 295: Ask to run, infer/export.

vbaddi (commented):
- pre-commit checks are failing, pls rerun pre-commit.

QEfficient/generation/text_generation_inference.py
quic-amitraj: Isn't it should be infer/compile?

vbaddi: things are somewhat being handled in #73 , we need to check once. if there is duplicity lets avoid it.

QEfficient/compile/compile_helper.py
```diff
@@ -126,6 +127,11 @@ def compile(
     :mxfp6: bool. Enable compilation for MXFP6 precision
     :mxint8: Compress Present/Past KV to MXINT8 using CustomIO config, default is False.
     """
+    # Checking if qpc file already exist
+    if qpc_exists(os.path.join(qpc_path, "qpcs")):
+        print(f"Pre-compiled qpc found at {qpc_path}")
+        return qpc_path
+
```
irajagop, line 134: We simply check if the qpc is present or not. But we should check if the qpc exists for the same compilation commands. Eg: First call could be `compile(num_cores=1)`, but next when the user calls `compile(num_cores=16)`, she won't see the improvement in runtime. We should use some intelligent caching behaviour.

ochougul:
Not required now! being addressed in #116

PR #79, opened 2024-07-23
fix: FP16 clipping applied to INT types

irajagop:
- Create FP16Clip transform which applies correct logic or FP16 clipping
- Transforms now return flag indicating transformation
- FP16 clipping should be applied only for FLOAT types
- Added tests for FP16Clip transform
    - Tests for INT clipping
- Avoid saving onnx file if no tensors are clipped in FP16 range.


tests/transformers/models/test_causal_lm_models.py
```diff
@@ -5,25 +5,35 @@
...
-
 import pytest
 from transformers import AutoConfig, AutoModelForCausalLM
 
 from QEfficient.utils.constants import Constants
 from QEfficient.utils.device_utils import get_available_device_id
 from tests.utils import get_cloud_ai_100_tokens, set_up
 
-TEST_CONFIG_FILE_PATH = "tests/config.json"
+test_params = [
```
ochougul, line 15: We are introducing num_layers as parameter in the config file. This is to fix pytorch vs onnx value mismatch in single layer phi model.

irajagop: - Main reason is because there is no reason for it to be in a separate file - It is not used anywhere else than here.
- Simplifying the configuration for this test. It would be best if the params for this test resides in this file, no need to hunt for another file somewhere in the repo.

Did we debug why single layer model is failing for phi model? We can have mismatch between ONNX and AIC, but PyTorch vs ONNX shouldn't have an issue.

quic-mamta: yes Pytorch and ONNX outputs are matching, AIC output is different.
image

QEfficient/base/pytorch_transforms.py
```diff
@@ -19,12 +19,13 @@ def __init__(self):
         raise TypeError("Transform classes are not to be instantiated. Directly use the `apply` method.")
 
     @classmethod
-    def apply(cls, model: nn.Module) -> nn.Module:
+    def apply(cls, model: nn.Module) -> [nn.Module, bool]:
```
ochougul, line 22: return type

ochougul (changes requested):

ochougul (approved):

PR #81, opened 2024-08-02
moved from transform_lm to PyTorchTransforms; Added transform specifi…

ochougul:
* moved `transfom_lm` to `PyTorchTransforms` 
* Added `PyTorchTransforms` sample tests
* Updated `causal_lm` tests to use `QEffAUtoModel` functions instead of `QEfficient`


QEfficient/transformers/transforms.py
irajagop, line 1: Since these are transformers related changes, I would prefer this to be in transformers directory only. Similarly, I'm adding peft related changes in peft directory.

It's better to do this way instead of putting all the transforms in the same file.

irajagop (changes requested):

QEfficient/transformers/transforms.py
ochougul: * Moved KVcache and CustomOPs transforms  to transformers dir.
* Tests for other model architectures will be added by @abukhoy 
* It's easier to keep tests for different models separate, because we can parametrize based on model class in different functions, it's getting complicated when trying to put everything in single test

QEfficient/transformers/pytorch_transforms.py
```diff
@@ -46,46 +38,57 @@
...
+from QEfficient.transformers.models.qwen2.modeling_qwen2 import QEffQwen2Attention, QEffQwen2ForCausalLM, QEffQwen2Model
+from QEfficient.transformers.models.starcoder2.modeling_starcoder2 import (
     QEffStarcoder2Attention,
     QEffStarcoder2ForCausalLM,
     QEffStarcoder2Model,
 )
 
 
-class CustomOps(ModuleMapping):
+class CustomOpsTransform(ModuleMapping):
```
irajagop, line 91: Do we need the `Transform` suffix? If so we should add it to `ModuleMapping` also. I'm thinking we don't add it to every transform.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -82,7 +73,7 @@ def from_pretrained(cls, pretrained_model_name_or_path: str, *args, **kwargs):
         )  # Remove model_card_name from kwargs for transformers APIs
 
         model = QEFFAutoModelToTransformersAutoModelMap[cls.__name__].from_pretrained(
-            pretrained_model_name_or_path, *args, **kwargs
+            pretrained_model_name_or_path, num_hidden_layers=1, attn_implementation="eager", *args, **kwargs
```
irajagop, line 76: Why this num_hidden_layers=1? I think this came while you were testing it out. Please review before pushing.

Also, while setting attn_implementation="eager", we also need to check if kwargs already has that. Otherwise, it will trigger a bug.

QEfficient/transformers/pytorch_transforms.py
```diff
@@ -1,12 +1,4 @@
-# -----------------------------------------------------------------------------
-#
-# Copyright (c) 2024 Qualcomm Innovation Center, Inc. All rights reserved.
-# SPDX-License-Identifier: BSD-3-Clause
-#
-# ----------------------------------------------------------------------------
```
irajagop, line 6: Licence clause need not be removed

QEfficient/transformers/pytorch_transforms.py
```diff
@@ -151,5 +154,8 @@ class KVCache(ModuleMapping):
     }
 
     @classmethod
-    def apply(cls, model: nn.Module) -> Tuple[nn.Module, bool]:
-        return super().apply(model)
+    def apply(cls, model: nn.Module):
```
irajagop, line 157: The output signature need node be removed.

tests/base/test_pytorch_transforms.py
```diff
@@ -40,3 +138,387 @@ def forward(self, x):
...
+
+
+@pytest.mark.parametrize("input_len", [8], ids=lambda x: "input_len=" + str(x))
+@pytest.mark.parametrize("hidden_size", [128], ids=lambda x: "hidden_size=" + str(x))
+@pytest.mark.parametrize("intermediate_size", [512], ids=lambda x: "intermediate_size=" + str(x))
+@pytest.mark.parametrize("num_key_value_heads", [8, 32], ids=lambda x: "num_key_value_heads=" + str(x))
+@pytest.mark.parametrize("num_attention_heads", [32], ids=lambda x: "num_attention_heads=" + str(x))
+@pytest.mark.parametrize("ctx_len", [32], ids=lambda x: "ctx_len=" + str(x))
+@pytest.mark.parametrize("num_hidden_layers", [1, 3], ids=lambda x: "num_hidden_layers=" + str(x))
+def test_kv_cache_transform_llama(
```
irajagop, line 175: There should be a better way to do this than copy paste this function and just change the names. You should be able to use `transformers.AutoConfig` along with `transformers.AutoModelForCausalLM`.

irajagop (changes requested):

QEfficient/transformers/models/modeling_auto.py
ochougul: We don't support any other `attn_implementation` right?
I will raise an error if any other is provided via `kwargs`
Will remove `num_hidden_layers`

tests/base/test_pytorch_transforms.py
ochougul: We are not directly copying the code.
The names of input arguments are different for different models.
and we are changing the arguments to be of smaller size so that the runtime reduces.
If we had to combine all these tests into one tests, we still will have to write these parameters names and values somewhere.
It's easier that pytorch handles this than us doing it.

QEfficient/transformers/pytorch_transforms.py
ochougul: added back

QEfficient/transformers/pytorch_transforms.py
ochougul: added back

QEfficient/transformers/pytorch_transforms.py
irajagop: Let's be consistent and use `Transform` suffix everywhere.

ochougul: Okay, will do.

tests/base/test_pytorch_transforms.py
```diff
@@ -40,3 +138,387 @@ def forward(self, x):
     assert transformed
     y2 = model(x)
     assert torch.all(y2 == x)
+
+
+@pytest.mark.parametrize("input_size", [random.randint(0, 10)], ids=lambda x: "input_size=" + str(x))
+@pytest.mark.parametrize("hidden_size", random.sample([4, 64, 2048, 4096], 2), ids=lambda x: "hidden_size=" + str(x))
```
irajagop, line 144: We should not be picking hidden_size at random for each run. We should have a fixed set of parameters. Because when we have a test fail once, we'll not be reproduce when we run the next time.

And I think we should not be adding a common test for all custom ops like this. We should be adding tests for each custom op separately.

irajagop (changes requested):

tests/transformers/test_pytorch_transforms.py
```diff
@@ -0,0 +1,183 @@
...
+    ("qwen2", 3, 32, 128, {"num_key_value_heads": 8, "intermediate_size": 512}, 0.8),
+    ("qwen2", 3, 32, 128, {"num_key_value_heads": 32, "intermediate_size": 512}, 0.8),
+    ("starcoder2", 3, 24, 192, {"num_key_value_heads": 2, "intermediate_size": 512}, 0.8),
+    ("starcoder2", 1, 24, 192, {"num_key_value_heads": 2, "intermediate_size": 512}, 0.8),
+    ("starcoder2", 3, 24, 192, {"num_key_value_heads": 24, "intermediate_size": 512}, 0.8),
+    ("starcoder2", 1, 24, 192, {"num_key_value_heads": 24, "intermediate_size": 512}, 0.8),
+]
+
+
+def compare_original_vs_kv_model_pt_outputs(original_val, kv_val, tolerance=1e-6) -> bool:
```
irajagop, line 53: Since we made it a single test now, can we move this code directly into the function, if there is no reuse of this code?

ochougul: This function will be useful in future for testing other pytorch transforms.
I would like to keep it.

irajagop (approved):
LGTM

PR #80, opened 2024-07-24
[WIP]: Finetune initial commit

vbaddi:

PR #85, opened 2024-08-13
Peft Lora implementation

irajagop:
- Prerequisite: #89 
- `AutoPeftModelForCausalLM` for loading LoRA models
- Better export code that can be utilized for other `auto` classes
- Hashing model cache location to avoid exporting again
- Hashing model compile location to avoid compiling again


pyproject.toml
```diff
@@ -20,6 +20,7 @@ classifiers = [
 requires-python = "==3.8.*"
 dependencies = [
     "transformers==4.41.2",
+    "peft==0.12.0",
```
ochougul, line 23: Please initialize the legal mail for this one, as soon as possible.

irajagop: @quic-jhugo @carlstreeter-quic Please review and approve using `peft==0.12.0` with QEfficient.

quic-jhugo: Actually, hold on.

carlstreeter-quic: PR80 is requesting peft 0.11.1.  Those reviews are underway.  Is this a necessary update, or is that version OK?

carlstreeter-quic: And how do these PRs interact?

irajagop: @carlstreeter-quic #80 is now closed. It was an initial commit. We're working on that feature in a separate branch (`refine`). Both this PR and the `refine` branch will use 0.12.0.

QEfficient/peft/auto.py
```diff
@@ -0,0 +1,436 @@
...
+        # absolute path for another run will result in different hash.
+        # The added complexity to resolve different paths to same location is not worth pursuing.
+        # Instead, advise the user to always provide same relative paths or absolute paths for local models.
+
+        # Compute the hash with: model_config, peft_config, transforms
+        mhash = hashlib.sha256()
+        mhash.update(to_hashable(self.model.get_base_model().config.to_diff_dict()))
+        mhash.update(to_hashable(self.model.active_peft_config.to_dict()))
+        mhash.update(to_hashable(self.transform_names()))
+        mhash = mhash.hexdigest()[:16]
```
ochougul, line 75: Should we use `SHAKE_128`? for variable length `hexdigest`? or `xxhash` -> faster

QEfficient/peft/auto.py
ochougul, line 1: We should provide a function that takes `model_name_or_path`, `adapter_name` etc. and other variables that affect the path of the saved `onnx` files. and return the path of saved `onnx` file if it exists.

Same for `qpc` file.

This should be a `classmethod`, user need not initialize the class just in order to get path of saved files.

QEfficient/peft/auto.py
```diff
@@ -0,0 +1,436 @@
...
+            raise e
+
+        finally:
+            shutil.rmtree(tmp_onnx_dir, ignore_errors=True)
+
+        self.onnx_path = onnx_path
+        return onnx_path
+
+    def export(self, export_dir: Optional[str] = None) -> str:
+        kv_cache_shape = get_padding_shape_from_config(self.model.config, 1, 32)
```
ochougul, line 178: Why are we fixing `batch_size` and `seq_len`?

QEfficient/peft/auto.py
```diff
@@ -0,0 +1,436 @@
...
+        inputs: Optional[Union[torch.Tensor, np.ndarray]] = None,
+        generation_config: Optional[GenerationConfig] = None,
+        stopping_criteria: Optional[StoppingCriteria] = None,
+        streamer: Optional[BaseStreamer] = None,
+        **kwargs,
+    ) -> np.ndarray:
+        # Initialize session
+        if self.qpc_session is None:
+            if self.qpc_path is None:
+                raise ValueError("Please compile the model with `model.compile(...)`")
```
ochougul, line 348: `FileNotFoundError` ?

QEfficient/peft/auto.py
```diff
@@ -0,0 +1,436 @@
...
+            specializations=specializations,
+            convert_to_fp16=True,
+            mxfp6_matmul=mxfp6_matmul,
+            custom_io=custom_io,
+            mdp_ts_num_devices=num_devices,
+            aic_num_cores=num_cores,
+            **compiler_options,
+        )
+
+    def generate(
```
ochougul, line 337: We provide way to generate outputs on multiple input prompts, which includes splitting inputs based on batch size as well.

The `QEffAutoModelForCausalLM` currently takes list of strings as input. We need to introduce pipelines eventually to make align our interface with transformers.

But with this PR, we have two different generate methods taking different types of inputs. So two questions.
1. How do you plan to provide way to generate inference on multiple input prompts and when?
2. Let's align this method with `QEffAutoModelForCausalLM` before the release.

If it is not possible to provide `Pipelines` before `1.18` we can make this generate align with `QEFFAutoModel` generate and provide a temporary way to generate output for multiple prompts -> which can eventually evolve into Pipelines

QEfficient/peft/onnx_transforms.py
```diff
@@ -0,0 +1,56 @@
...
+                out = onnx.ValueInfoProto(name=new_weight_name + "_RetainedState", type=type_proto)
+                model.graph.input.append(inp)
+                model.graph.output.append(out)
+
+                # Create a node that connects input -> output
+                node = onnx.helper.make_node("Identity", [inp.name], [out.name], new_weight_name + "_identity")
+                model.graph.node.append(node)
+
+                # Rename weight input
+                lora_weight_node = lora_weight_nodes[weight.name]
```
ochougul, line 44: have a map of `lora_weight_node_state = {node: True/False}` If True skip the for loop.
I think currently this will be called again in case the node inputs were already updated.

QEfficient/peft/onnx_transforms.py
```diff
@@ -0,0 +1,56 @@
...
+                for j, inp in enumerate(lora_weight_node.input):
+                    if inp.endswith(weight_suffix):
+                        lora_weight_node.input[j] = new_weight_name
```
ochougul, line 47: have a map of `lora_weight_node_state = {node: True/False}` If True skip the for loop.
I think currently this will be called again in case the node inputs were already updated.

QEfficient/peft/auto.py
```diff
@@ -0,0 +1,436 @@
...
+        raise NotImplementedError("Disabling adapters not supported currently")
+
+    @classmethod
+    def _from_pretrained(cls, pretrained_name_or_path: str, *args, **kwargs):
+        # Base class
+        model = cls._hf_auto_class.from_pretrained(pretrained_name_or_path, *args, **kwargs)
+        return cls(model)
+
+    @classmethod
+    def from_pretrained(cls, pretrained_name_or_path: str, *args, **kwargs):
```
ochougul, line 102: Does not handle Continuous batching case. If we don't want to allow this. raise error in `from_pretrained` when `full_batch_size` is passed

QEfficient/peft/auto.py
irajagop: I researched a bit and thought of using SHAKE, but it may not be supported in all platforms. hashlib uses OpenSSL system libraries to provide the support for hashing. In older versions of OpenSSL, it may not be supported. This is mentioned in python library documentation: https://docs.python.org/3/library/hashlib.html 

> Changed in version 3.9: Hashlib now uses SHA3 and SHAKE from OpenSSL if it provides it.

As for xxhash, it's adding unnecessary dependency for such a very small hash. We're not hashing GB's or TB's of data to go for xxhash to make it faster. So went with much simpler and standard way of doing things.

QEfficient/peft/auto.py
irajagop: We can have it in the constants, but it doesn't matter because the resulting onnx is going to be dynamic anyways. I'll move it to constants still.

QEfficient/peft/auto.py
```diff
@@ -0,0 +1,436 @@
...
+    ) -> str:
+        # Base class
+        onnx_path = Path(onnx_path or self.onnx_path)
+        compile_dir = Path(compile_dir or onnx_path.parent)
+        qpc_path = compile_dir / "qpc"
+        if not onnx_path.is_file():
+            raise FileNotFoundError(f"ONNX file not found at: {onnx_path}")
+
+        command = ["/opt/qti-aic/exec/qaic-exec", f"-m={onnx_path}", "-aic-hw", "-aic-hw-version=2.0"]
+        for key, value in compiler_options.items():
```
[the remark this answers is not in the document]
irajagop: Currently there are no e2e tests to test the behaviour for this class. Will try to write one or we'll have to write this after the base methods are moved to base classes.

I've done extensive testing with multiple adapter models on my side though.

QEfficient/peft/peft_model.py
```diff
@@ -0,0 +1,61 @@
...
+#
+# Copyright (c) 2024 Qualcomm Innovation Center, Inc. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# ----------------------------------------------------------------------------
+
+from peft import PeftModelForCausalLM, PeftType
+
+
+class QEffPeftModelForCausalLM(PeftModelForCausalLM):
```
ochougul, line 11: How is this class supposed to be used?

QEfficient/peft/onnx_transforms.py
irajagop: This will not be called again, because we're renaming the weight name, so it will not match the `endswith(...)` condition. If all the inputs are already renamed, it will not match any of the inputs and it is already skipping over.

QEfficient/peft/peft_model.py
ochougul: If this class is not supposed to be used directly by user, we should put errors

irajagop: This is a patched class to fix issues in onnx export, as it's not taking past_key_values as input in the original peft library. This just fixes that issue. It's used only in the pytorch transform (see pytorch_transform.py). I'll add a docstring to explain this.

QEfficient/utils/cache.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+from pathlib import Path
+
+QEFF_HOME: Path = None
+if "QEFF_HOME" in os.environ:
+    QEFF_HOME = Path(os.environ["QEFF_HOME"])
+elif "XDG_CACHE_HOME" in os.environ:
+    QEFF_HOME = Path(os.environ["XDG_CACHE_HOME"]) / "qeff_models"
+else:
+    QEFF_HOME = Path("~/.cache/qeff_models").expanduser()
```
ochougul, line 18: Where is this used? check `utils/constants/QEFF_MODELS_DIR`

QEfficient/utils/cache.py
ochougul, line 1: Move to `_utils.py` ?

tests/peft/test_peft_onnx_transforms.py
```diff
@@ -0,0 +1,63 @@
...
+        float[32, 32] layer2_{adapter_name}_weight = [ "location" : "{external_tensors_file}" ]
+    >
+    {{
+        layer1output = MatMul (input, layer1_{adapter_name}_weight)
+        output = MatMul (layer1output, layer2_{adapter_name}_weight)
+    }}
+    """)
+
+    out_onnx, transformed = AdapterWeightsToInputsTransform.apply(test_onnx, adapter_name=adapter_name)
+    assert not transformed
```
ochougul, line 35: did you want to remove these lines

QEfficient/peft/auto.py
ochougul: Can we get rid of only considering first 16 characters of `hexdigest`? That's the cause of concern

QEfficient/peft/auto.py
irajagop: I'll raise this error for now.

QEfficient/peft/auto.py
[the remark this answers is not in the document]
ochougul: Actually, you should add minimal functional test making sure the `QEffAutoPeftModelForCausalLM` works.
And above comment was specific to line `225-231` -> we need to tell user how to pass these params. and a list.
Or we can write in doc that always pass the params just the way compiler expects, so we don't need to check this.

QEfficient/utils/cache.py
irajagop: I'll have to pick the new changes introduced in the other PR. I'll refactor to move it from constants.py to this one, as this will contains all utilities related to caching.

QEfficient/utils/cache.py
irajagop: I think it's better to keep it here as the name `cache.py` provides clarity on what this file is going to be used for.

tests/peft/test_peft_onnx_transforms.py
irajagop: No, this also is a part of the transform test that it will not touch a model that doesn't match the exact criteria.

QEfficient/peft/peft_model.py
ochougul: Let's put errors in `__init__`, `from_pretrained`

ochougul:
Documentation needs to be updated.

QEfficient/peft/auto.py
irajagop: Even for large git repos (which use SHA256 hash) like linux kernel (millions of commits), prefix size of 11 is sufficient to avoid collision: https://stackoverflow.com/questions/32405922/in-my-repo-how-long-must-the-longest-hash-prefix-be-to-prevent-any-overlap

QEfficient/peft/auto.py
irajagop: Done

QEfficient/peft/peft_model.py
irajagop: This is like QEff wrapper for LlamaForCausalLM. I don't think we have to raise errors for everything that users don't have to use. There comes a time when a user with different needs wants to use this class.

QEfficient/utils/cache.py
irajagop: I'll be doing this refactor when I move the export utils to base classes, as this will return in pathlib.Path, whereas the other one returns a string.

QEfficient/peft/auto.py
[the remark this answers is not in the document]
irajagop: I've added in the docstring with example usage of how to use the class, load different adapters, run them, etc.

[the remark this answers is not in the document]
irajagop: I'll add a docstring on how to pass params to the `_compile` method.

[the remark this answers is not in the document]
irajagop: Added docstring on how to use the `_compile` method

pyproject.toml
irajagop: Removed this as all packages are added in a different PR.

QEfficient/peft/auto.py
irajagop: Since the Peft class is a new addition, we can use a different method from AutoModelForCausalLM. Users would have to read the documentation to use the same. When we migrate AutoModelForCausalLM to the same interface, we provide that as breaking change in the release.

Since this class interface aligns more with transformers class, it's simple for users to understand.

QEfficient/peft/auto.py
irajagop: Since the hash for onnx is computed in custom way using the objects of the class, we can't make this classmethod. However there are multiple ways to get the onnx_path of the model:
- Pass the `export_dir` parameter to `export()` method, this way the users can customize the directory where the onnx is saved.
- Run the `export()` method again, it will return the onnx_path without exporting the model again.

Same for qpc, `compile_dir` can be passed, as well as return value of `compile()` function can be used to determine the path.

QEfficient/__init__.py
```diff
@@ -9,6 +9,7 @@
 from QEfficient.compile.compile_helper import compile
 from QEfficient.exporter.export_hf_to_cloud_ai_100 import qualcomm_efficient_converter
 from QEfficient.generation.text_generation_inference import cloud_ai_100_exec_kv
+from QEfficient.peft import QEffAutoPeftModelForCausalLM
```
vbaddi, line 12: `Couple of points:
`
- The caching changes here are specific to PEFT or apply to overall repo? If not, then are we following two paths currently to save PEFT models and mainline models separately?
- Do we have details in the ReadMe on usage of this class? like passing adapters along with the model name to initiate this?

irajagop: 1. Caching is specific to PEFT for now. I'm creating a separate PR to have this available to `QEffAutoModelForCausalLM` class next.
2. I have added the sample usage in the docstring, and it will be added in the github.io pages when merged. See docs/source/hl_api.md file for the new section.

QEfficient/utils/cache.py
ochougul: This needs to be updated before we do code freeze. This is duplicate code.

QEfficient/peft/auto.py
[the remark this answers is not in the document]
ochougul: We should have e2e test for this in CI

PR #93, opened 2024-08-22
Added beautified error for GatedRepoError from HuggingFaceHub

ochougul:


QEfficient/utils/_utils.py
```diff
@@ -61,16 +62,21 @@ def hf_download(
...
             retry_count += 1
-
         except HTTPError as e:
             retry_count = max_retries
             if e.response.status_code == 401:
                 logger.info("You need to pass a valid `--hf_token=...` to download private checkpoints.")
             else:
                 raise e
+        except Exception as e:
+            raise e
```
irajagop, line 79: This is not needed, by default uncaught exception is always raised.

irajagop (changes requested):

QEfficient/utils/_utils.py
irajagop: Try this:

```python
# except_test.py
try:
    raise OSError("ose")
except FileNotFoundError:
    print("fnfe")
except RuntimeError:
    print("rte")
```

You'll get:

```
Traceback (most recent call last):
  File "except_test.py", line 3, in <module>
    raise OSError("ose")
OSError: ose
```

ochougul: Okay, will remove. Thanks

PR #97, opened 2024-09-03
Added support for creating model path hash when model card is not pro…

quic-rishinr:
Modified _transformers/auto.py to create a model card name as a hash of the local model path when the model path is given as input and the model card name is not specified.

### DIFF
--- a/QEfficient/src/_transformers/auto.py
+++ b/QEfficient/src/_transformers/auto.py
@@ -19,7 +19,7 @@ from QEfficient.src._transformers.runtime_args import (
 )
 from QEfficient.src.base import QEFFBaseModel, Runtime
 from QEfficient.transformers.modeling_utils import TransformersToQEffModulesDict
-from QEfficient.utils import get_qpc_dir_name_infer, load_hf_tokenizer, qpc_exists
+from QEfficient.utils import generate_sha256_hash, get_qpc_dir_name_infer, load_hf_tokenizer, qpc_exists
 from QEfficient.utils.logging_utils import logger
 
 # Dictionary that defines the interface from transformers to be used underneath the QEFF interface
@@ -70,11 +70,7 @@ class QEFFTransformersBase(QEFFBaseModel):
     def get_model_card_name(self) -> str:
         # FIXME: use getter
         if self.model_card_name is None:
-            # Handle when pretrained_model_name_or_path is a path and we don't know the model_card_name
-            assert not os.path.isdir(
-                self.pretrained_model_name_or_path
-            ), f"Please provide `model_card_name` argument as valid string, got {self.model_card_name}"
-            self.model_card_name = self.pretrained_model_name_or_path
+            self.model_card_name = generate_sha256_hash(self.pretrained_model_name_or_path)
 
         return self.model_card_name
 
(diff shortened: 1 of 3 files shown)


QEfficient/utils/_utils.py
```diff
@@ -5,6 +5,7 @@
 #
 # -----------------------------------------------------------------------------
 
+import hashlib
```
vbaddi, line 8: Do we need this package? or its preinstalled in unix?

QEfficient/utils/_utils.py
```diff
@@ -186,3 +187,17 @@ def padding_check_and_fix(tokenizer: Union[PreTrainedTokenizer, PreTrainedTokeni
             tokenizer.pad_token_id = tokenizer.eos_token_id
         else:
             tokenizer.pad_token_id = tokenizer.vocab_size - 1
+
+
+def generate_sha256_hash(data: str) -> str:
+    """
+    Generate a SHA-256 hash for the given data.
+
+    :param data: The input data to hash.
```
vbaddi, line 196: can you add more details on what this data excatly is? `model_card` name?

vbaddi (changes requested):

QEfficient/src/_transformers/auto.py
```diff
@@ -70,11 +70,7 @@ def tokenizer(self) -> Union[PreTrainedTokenizer, PreTrainedTokenizerFast]:
     def get_model_card_name(self) -> str:
         # FIXME: use getter
         if self.model_card_name is None:
-            # Handle when pretrained_model_name_or_path is a path and we don't know the model_card_name
-            assert not os.path.isdir(
-                self.pretrained_model_name_or_path
-            ), f"Please provide `model_card_name` argument as valid string, got {self.model_card_name}"
-            self.model_card_name = self.pretrained_model_name_or_path
+            self.model_card_name = generate_sha256_hash(self.pretrained_model_name_or_path)
```
irajagop, line 73: This should only be done when the provided `pretrained_model_name_or_path` is a local directory.

irajagop (changes requested):

QEfficient/utils/_utils.py
quic-rishinr: its a built-in library in Python
https://docs.python.org/3.8/library/hashlib.html

QEfficient/utils/_utils.py
quic-rishinr: Any string data can be converted to hash. I added it as a generic method under utils.

QEfficient/src/_transformers/auto.py
quic-rishinr: yes. Updated it in the latest patch.

vbaddi (approved):

irajagop (approved):

QEfficient/src/_transformers/auto.py
ochougul: The point was to force user to pass `model_card_name` at export stage. So that she/he can figure out where the exported/compiled `onnx/qpc` is saved under `QEFF_HOME`. 
In PR 85, we are moving towards hash always for PEFT models. Once this feature is added for `QEFFAutoModelForCausalLM` we don't need to do this.

PR #101, opened 2024-09-04
AWQ+GPTQ

ochougul:
* Added preprocess layer before loading quantized awq weights
* Added onnx export
* Added ScaledActivation class
* Refactoring the code to right places and added one single test for now
* Cleaned code
* Added proper tests, added decorator for updating quantizers, cleaned code
* Fixed CLI
* Added auto file for decorator

---------


QEfficient/transformers/quantizers/quant_transforms.py
```diff
@@ -0,0 +1,51 @@
...
+        original_module.weight = fp16_weight
+        new_module = QuantLinearORT(
+            original_module.w_bit,
+            original_module.group_size,
+            original_module.in_features,
+            original_module.out_features,
+            original_module.bias is not None,
+        )
+        new_module.bias = original_module.bias if original_module.bias is not None else None
+        new_module.pack(original_module, scales.T, zeros.T, original_module.g_idx)
```
irajagop, line 50: Here we have to call a method of new_module `new_module.pack`, we could simply move the `unpack_and_dequantize` also into that class and call these 2 from `__qeff_init__`.

This will make the code look simpler and not scattered across multiple files for 1 feature.

QEfficient/transformers/quantizers/quantizer_awq.py
```diff
@@ -0,0 +1,198 @@
...
+
+        model = replace_quantization_scales(model, model.config.model_type)
+        if not has_been_replaced:
+            logger.warning(
+                "You are loading an AWQ model but no linear modules were found in your model."
+                " Please double check your model architecture, or submit an issue on github if you think this is a bug."
+            )
+
+
+class ScaledActivation(nn.Module):
```
irajagop, line 81: If any of the functions or classes defined here can be reused from transformers, please do so. Ignore if already done.

tests/base/test_pytorch_transforms.py
```diff
@@ -19,24 +32,35 @@ def test_module_mapping_transform():
...
+def test_module_mutator_transform():
+    with pytest.raises(TypeError):
+        ModuleMutatorTransform()
+
+    class TestTransform(ModuleMutatorTransform):
+        _match_class = nn.Linear
+
+        @classmethod
+        def mutate(cls, original_module: nn.Module, parent_module: nn.Module):
+            return nn.Identity()
```
irajagop, line 55: In this test, we can easily see that the functionality of ModuleMutatorTransform and ModuleMappingTransform are basically the same.

irajagop (changes requested):

QEfficient/transformers/quantizers/quant_transforms.py
ochougul: The `upack_and_dequantize` code will be moved to `quantized_utils.py` as They `awq` and `gptq` share same code along with their own differences.
It is possible to move this code to a base class that both GPTQ and AWQ linear classes inherit.
I will check the best possible way to do this.

QEfficient/transformers/quantizers/quantizer_awq.py
ochougul: This class is defined in `auto-awq` package, We don't want to add extra package dependency for this.

tests/base/test_pytorch_transforms.py
ochougul: They are not same, check line 64 of this file.
`ModuleMapping` is when you don't need to initialize new class, and it is possible to directly assign `old_module.__class__=some_new_class`
`ModuleMutator` is when you must initialize a new class and replace in place of old one. Only class mapping is not enough. In AWQ/GPTQ we need to initialize a new class and then pack the unpacked weights into new class object.

QEfficient/transformers/quantizers/quantizer_awq.py
irajagop: From [HF transformer docs](https://huggingface.co/docs/transformers/main/en/quantization/awq), it seems that autoawq is a prerequisite for running AWQ transformer models. I think we can add an optional dependency for the same.

tests/base/test_pytorch_transforms.py
ochougul: I did some thinking on using `__qeff_init__`.

- Approach 1: Using `__qeff_init__`  in `WQLinearGEMM` class
    - Moving unpacking, initialization and repacking to `__qeff_init__`
        If we do this and return the new `QuantLinearORT` class as output, we still need a way to insert this class into model i.e. `setattr` call -> which is happening in `ModuleMutator.apply` method. In this case we are moving away from what `__qeff_init__` is supposed to do, it needs to change the module in-place without external help. And also, In this case `WQLinearGEMM` has the code of packing for `QuantLinearORT` which it should not care about.
- Apprach 2:  Using `__qeff_init__` in Attention clasess
   - This leads to us adding `__qeff_init__` in all models Attention classes i.e. separate for `LlamaAttention`, `MistralAttention` and so on. Not scalable.

I think `ModuleMutator` can be quiete useful when we want to

ochougul: If developer doesn't apply `KVCacheTransform` but applies `AWQToMatmulNbitsTransform` the model will give wrong output as weights are not packed yet, and we can't force user to always run `KVCacheTransform`

QEfficient/transformers/quantizers/quantizer_awq.py
ochougul: We don't want to add this because of legal review process. This will delay the awq feature release

tests/base/test_pytorch_transforms.py
irajagop: I'm not sure how this has anything to do with `KVCacheTransform`. If you see closely, we have to add the `__qeff_init__` in the `QuantLinearORT`, not in the KVCache classes. `KVCacheTransform` and `AWQToMatmulNbitsTransform` will be completely orthogonal.

irajagop: - We set `__class__` exactly to change the type of a variable at runtime based on runtime conditions. Here the runtime condition is that the user wants to run the pytorch module on QAIC and not on other devices available in pytorch.
- If we initialize a new class and then do `setattr`, we'll be duplicating the weights of the model in memory. The primary reason we use `a.__class__ = B` is to not duplicate the weights unnecessarily, we only change the behaviour of `forward` in the class.
- Again, if we want to switch to `setattr` instead of `a.__class__ = B` everywhere, we have to know parameters to be passed to `B()` initialization, which may be different for each class. Creating a new ModuleMutator for each and every class that we need to replace is not scalable.
- Using `ModuleMutator` actually scatters the related code across in multiple files. For example, here part of the implementation is in Transform and part of it is in the QuantLinearORT class.
- Rotary embedding implementation is different for different models, some use default embeddings, some use scaled embeddings. There is also rotary_dim config, which will change the way embeddings are applied. So, we can't really re-use the implementation.

ModuleMutator is almost the same as ModuleMapping:
- You could simple move the `mutate` function to `__qeff_init__` in the target class.
- This will keep the code for that specific class in there, instead of having it in a different file (transforms).
- In AWQ case, we have to make changes in weights and drop old weights. We could make all the changes and then delete the old weights simply with `del self.old_weight`

Issue with ModuleMutator design:
- You can match only one class; this works well for AWQ currently. Say tomorrow there are new modules to support conv and norm as WQConvGEMM and WQLayerNormGEMM. Now you have to create 2 more transforms just to target those classes.
- If you think about a solution for the issue, you'll naturally come to adding a special function (`__qeff_init__`) in the replacement class itself.

With all that said, you can continue with ModuleMutator class if you still think it's a good idea. I'd like to close this thread as it has gone longer that I expected.

quic-amitraj:
**TS with 4 Devices** 
1. Model (TheBloke/Mistral-7B-Instruct-v0.1-AWQ)-
image

2. Model (solidrust/Meta-Llama-3-8B-Instruct-hf-AWQ)-
image

quic-amitraj:
CB with batch size-4.
1. TheBloke/Mistral-7B-Instruct-v0.1-AWQ
image

quic-amitraj:
CB with Batch size-4

Model **solidrust/Meta-Llama-3-8B-Instruct-hf-AWQ**

image

PR #114, opened 2024-09-16
Add rope scaling to llama modeling file

vbaddi:


ochougul (commented):
Does the output remain same for models without scaling before and after this PR.
Are the model with different rope types = [default, linear, dynamic, yarn, longrope, llama3] tested?

vbaddi:
> Does the output remain same for models without scaling before and after this PR. Are the model with different rope types = [default, linear, dynamic, yarn, longrope, llama3] tested?

The various tests happen as part of the CI, but yes for llama specific models rope types are tested.

irajagop (approved):

PR #116, opened 2024-09-16
Intelligent caching behaviour

irajagop:
- Use hash of model config to compute model export path
- Use hash of compilation parameters for qpc path
- Move `_export` & `_compile` utility methods to base class
- Removed need for tokenizer for export
- Removed need for `model_card_name`
- Replace `transform()` with Transforms
- Added `QEFFAutoModelForCausalLMwithCB` for continuous batching
- `export()` called automatically by `compile()`
- Added tests for CausalLM class


QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,284 @@
...
+        if onnx_path is None and self.onnx_path is None:
+            self.export()
+
+        onnx_path = Path(onnx_path or self.onnx_path)
+        compile_dir = Path(compile_dir or onnx_path.parent)
+        qpc_path = compile_dir / "qpc"
+        if not onnx_path.is_file():
+            raise FileNotFoundError(f"ONNX file not found at: {onnx_path}")
+
+        command = ["/opt/qti-aic/exec/qaic-exec", f"-m={onnx_path}", "-aic-hw", "-aic-hw-version=2.0"]
```
quic-rishinr, line 221: wouldnt be better if we can keep the command constants under constant file and use it here?

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,284 @@
...
+        if qpc_path.is_dir():
+            if (qpc_path / "programqpc.bin").is_file():
+                self.qpc_path = qpc_path
+                return qpc_path
+            # Probably compilation failure last time, delete directory to start over
+            shutil.rmtree(qpc_path)
+
+        command.append(f"-aic-binary-dir={qpc_path}")
+        logger.info(f"Running compiler: {' '.join(command)}")
+        subprocess.run(command).check_returncode()
```
quic-rishinr, line 285: Shouldnt we add a trycatch for subprocess.run? if it fails qeff would fail with an uncaught exception right?

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,284 @@
...
-    def run_pytorch(self, inputs):
-        raise NotImplementedError("Reached too far!!")
+        # Apply the transformations
+        for transform in self._pytorch_transforms:
+            self.model, transformed = transform.apply(self.model)
+        logger.info("Pytorch transforms applied")
 
-    def run_ort(self, inputs: Dict[str, np.ndarray], /, cache_session: bool = False):
-        raise NotImplementedError("Reached too far!!")
+    @abstractproperty
```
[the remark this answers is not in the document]
irajagop: I had missed it, good catch!

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,284 @@
...
 
-    @classmethod
-    def from_pretrained(cls, pretrained_model_name_or_path: str, *args, **kwargs):
-        raise NotImplementedError("Must implement for child classes")
-
-    def run_pytorch(self, inputs):
-        raise NotImplementedError("Reached too far!!")
+        # Apply the transformations
+        for transform in self._pytorch_transforms:
+            self.model, transformed = transform.apply(self.model)
```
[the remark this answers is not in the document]
irajagop: `transformed` is not used anywhere right now. We could use this to find if none of the transforms were applied. This will be a signal that it's some unsupported architecture. Then we raise a warning/error that it's an unsupported model. I'll try adding that change here in this PR.

QEfficient/base/modeling_qeff.py
irajagop: We don't have to try/catch, because even after catching, what can we do? We could only re-raise as a different error. Better to propagate the same error to the user.

QEfficient/base/modeling_qeff.py
irajagop: Yes, better to move the 3 constant values.

QEfficient/base/modeling_qeff.py
quic-rishinr: I think it would be better if we can catch and wrap the error around the stage at which the error is observed. Otherwise, user would need to figure out from where the error is coming from.

irajagop: By wrap around, if you mean add some descriptive message to the user that the compiler failed, we can do that. I'll be adding that.

QEfficient/base/modeling_qeff.py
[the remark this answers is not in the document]
irajagop: I've raised a warning when no transformations are applied.

QEfficient/transformers/modeling_utils.py
```diff
@@ -117,6 +117,7 @@
         GPT2LMHeadModel.__name__,
         GPTJForCausalLM.__name__,
         MptForCausalLM.__name__,
+        "MPTForCausalLM",
```
quic-mamta, line 120: instead of this can we do, if full_batch_size and qeff_model.model.config.architectures[0].lower() not in [s.lower() for s in get_lists_of_cb_qeff_models.architectures]: in line 323 of QEfficient/exporter/export_hf_to_ai_100.py

ochougul (changes requested):
will review rest of the files later.

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,307 @@
...
+        # Apply the transformations
+        any_transformed = False
+        for transform in self._pytorch_transforms:
+            self.model, transformed = transform.apply(self.model)
+            any_transformed = any_transformed or transformed
 
-    def run_ort(self, inputs: Dict[str, np.ndarray], /, cache_session: bool = False):
-        raise NotImplementedError("Reached too far!!")
+        if not any_transformed:
+            warnings.warn(f"No transforms applied to model: {self.model_name}. It may be an unsupported model!")
```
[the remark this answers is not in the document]
irajagop: The issue is unavoidable by us (library). But the client code can avoid it, by using a supported model. The above answer says if the client (user of the library) application should be modified, it should raise a warning. Also, if the user wants, they can capture the warnings in logging (read the second answer from the same page above).

We can do both too. Add a logging as well as warning.

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,307 @@
...
+                model, transformed = transform.apply(model, **onnx_transform_kwargs)
+            model.metadata_props.append(
+                onnx.StringStringEntryProto(key="qeff_transforms", value=",".join(self._transform_names()))
+            )
+            logger.info("ONNX transforms applied")
+
+            onnx.save(model, onnx_path)
+            logger.info("Transformed onnx saved")
+
+        except Exception as e:
```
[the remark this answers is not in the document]
irajagop: It may be misleading from the error message "ONNX export failed". But from API point of view, "export()" always means "export and apply ONNX transforms".

Here, "export and transform" is made atomic. Even if transforms fail, the "export() API" fails and the temporary onnx files need to be deleted, as it may cause issues next time while trying export().

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,307 @@
...
+            onnx_transform_kwargs = {
+                "onnx_base_dir": str(tmp_onnx_dir),
+                "model_name": self.model_name,
+                **onnx_transform_kwargs,
+            }
+            for transform in self._onnx_transforms:
+                model, transformed = transform.apply(model, **onnx_transform_kwargs)
+            model.metadata_props.append(
+                onnx.StringStringEntryProto(key="qeff_transforms", value=",".join(self._transform_names()))
+            )
```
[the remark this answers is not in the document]
irajagop: I'll add a TODO, but we can't really track that whether transform needs to be applied or not. It will be determined by the Transform. If we add the logic of determining if it needs to be applied, we go back to same `if-else` kind of applying transforms.

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,307 @@
...
+                self.qpc_path = qpc_path
+                return qpc_path
+            # Probably compilation failure last time, delete directory to start over
+            shutil.rmtree(qpc_path)
+
+        command.append(f"-aic-binary-dir={qpc_path}")
+        logger.info(f"Running compiler: {' '.join(command)}")
+        try:
+            subprocess.run(command, capture_output=True, check=True)
+        except subprocess.CalledProcessError as e:
```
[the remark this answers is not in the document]
irajagop: Yes, the `check=True` argument will make sure if this exits with exit code != 0, it will raise exactly this error. Even if it's a different one, it will be raised directly instead of our `RuntimeError` below.

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,307 @@
...
+            command.append(f"-mdp-load-partition-config={mdp_ts_json}")
+            compile_hash.update(to_hashable({"mdp_ts_num_devices": mdp_ts_num_devices}))
+
+        # Check if already compiled
+        compile_hash = compile_hash.hexdigest()[:16]
+        qpc_path = qpc_path.with_name(qpc_path.name + "-" + compile_hash)
+        if qpc_path.is_dir():
+            if (qpc_path / "programqpc.bin").is_file():
+                self.qpc_path = qpc_path
+                return qpc_path
```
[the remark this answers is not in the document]
irajagop: We can do that but thought it will add more complexity in the method, simply checking again for `if specializations` and `if custom_io`. I think writing specializations and custom_io are not so heavy, so it's fine to be there before checking for re-compilation.

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,307 @@
...
+
+        self.onnx_path = onnx_path
+        return onnx_path
+
+    def _compile(
+        self,
+        onnx_path: Optional[str] = None,
+        compile_dir: Optional[str] = None,
+        *,
+        specializations: Optional[List[Dict[str, int]]] = None,
```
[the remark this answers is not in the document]
irajagop: I wanted to make the `_compile` a core utility. Not something tied to causal LM or any other use case. So, there is no need of change this in future.

We can make the subclasses like QEffAutoModelForCausalLM have required argument, rather than enforcing in base class.

tests/transformers/test_causal_lm.py
ochougul, line 1: Why are you changing this?

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -69,15 +106,19 @@ def test_causal_lm_pytorch_vs_kv_vs_ort_vs_ai100(model_name):
...
     assert (
         pytorch_hf_tokens == pytorch_kv_tokens
     ).all(), "Tokens don't match for HF PyTorch model output and KV PyTorch model output"
 
-    onnx_model_path = qeff_model.export()
+    _, onnx_model_path = qualcomm_efficient_converter(
+        model_name=model_name,
+        model_kv=qeff_model,
+        tokenizer=tokenizer,
+    )
```
ochougul, line 121: The idea is to test the export API? why change this?

tests/test_loader.py
ochougul, line 1: why remove this?

scripts/Jenkinsfile
```diff
@@ -36,9 +36,11 @@ pipeline
                     sh '''
                     . preflight_qeff/bin/activate
                     export TOKENIZERS_PARALLELISM=false
-                    pytest tests --ignore tests/cloud --junitxml=tests/tests_log1.xml
-                    pytest tests/cloud --junitxml=tests/tests_log2.xml
-                    junitparser merge tests/tests_log1.xml tests/tests_log2.xml tests/tests_log.xml
+                    pytest tests -m "not cli and not on_qaic" -n auto --junitxml=tests/tests_log1.xml &
+                    pytest tests -m "not cli and on_qaic" -n 4 --junitxml=tests/tests_log2.xml &
+                    wait
+                    pytest tests -m cli --junitxml=tests/tests_log3.xml
```
ochougul, line 42: this can also be run with -n 4

tests/test_loader.py
ochougul: I don't see a parallel test added

scripts/specializations.json
```diff
@@ -1,11 +1,13 @@
 {
 	"specializations": [
 		{
+			"full_batch_size": "4",
```
ochougul, line 4: does this mean we only test with full_batch_size? I think we should test both scenarios separately

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -122,7 +163,12 @@ def test_causal_lm_pytorch_vs_kv_vs_ort_vs_ai100(model_name):
     pytorch_hf_tokens = np.vstack(pytorch_hf_tokens)
 
     qeff_model = QEFFAutoModelForCausalLM(model_hf, f"{model_name}")
-    onnx_model_path = qeff_model.export()
+    _, onnx_model_path = qualcomm_efficient_converter(
+        model_name=model_name,
+        model_kv=qeff_model,
+        tokenizer=tokenizer,
+        full_batch_size=4,
+    )
```
ochougul, line 171: need to test fbs and without fbs separately

ochougul (changes requested):
please also runs spell-checker and rebase

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,307 @@
...
-    def from_pretrained(cls, pretrained_model_name_or_path: str, *args, **kwargs):
-        raise NotImplementedError("Must implement for child classes")
-
-    def run_pytorch(self, inputs):
-        raise NotImplementedError("Reached too far!!")
+        # Apply the transformations
+        any_transformed = False
+        for transform in self._pytorch_transforms:
+            self.model, transformed = transform.apply(self.model)
+            any_transformed = any_transformed or transformed
```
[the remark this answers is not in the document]
irajagop: Yes, that's the logic. Even if we have one transformation is applied, we're fine. That's also reflected in the variable name `any_transformed`. Anything else is expected here?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -6,385 +6,246 @@
...
+logger = logging.getLogger(__file__)
 
 
 class QEFFTransformersBase(QEFFBaseModel):
     """
     Parent class for models QEFF provides from transformers i.e. (AutoModel, AutoModelForCausalLM, AutoModelForAudioClassification etc.) from transformers/models/modeling_auto.py file.
     """
 
-    def __init__(self, model: nn.Module, pretrained_model_name_or_path: str, **kwargs) -> None:
+    _hf_auto_class: type
```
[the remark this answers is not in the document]
irajagop: `Type` is same as `Type[Any]`, as is `type`. There is not generic class that we can use in place of Any. I went with type as it's simpler.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -6,385 +6,246 @@
...
-        if self._tokenizer is None:
-            self._tokenizer = self.get_tokenizer()
-        return self._tokenizer
+    @property
+    def model_hash(self) -> str:
+        # NOTE: model_config.to_diff_dict() has "_name_or_path" attribute which is the model card name or path.
+        # Using same card name will result in same hash. But, using a relative path for one run and
+        # absolute path for another run will result in different hash.
+        # The added complexity to resolve different paths to same location is not worth pursuing.
+        # Instead, advise the user to always provide same relative paths or absolute paths for local models.
```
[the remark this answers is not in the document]
irajagop: We shouldn't delete the '_name_or_path'. That is the only identifier we have to recognize the model card name. If we have 2 models with same architecture and same config, but different card name, they'll result in same hash. Only this variable differentiates the 2.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -6,385 +6,246 @@
...
-            if isinstance(self.model.config.quantization_config, QEffAwqConfig):
-                self._pytorch_transforms.insert(0, AwqToMatmulNbitsTransform)
+    _hf_auto_class = AutoModelForCausalLM
+    _pytorch_transforms = [AwqToMatmulNbitsTransform, GPTQToMatmulNbitsTransform, CustomOpsTransform, KVCacheTransform]
+    _onnx_transforms = [FP16ClipTransform, SplitTensorsTransform]
 
-            if isinstance(self.model.config.quantization_config, QEffGPTQConfig):
-                self._pytorch_transforms.insert(0, GPTQToMatmulNbitsTransform)
+    def __init__(self, model: nn.Module, continuous_batching: bool = False, **kwargs):
+        if kwargs.pop("full_batch_size", None):
```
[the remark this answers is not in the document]
irajagop: This is better way to enforce deprecation, otherwise this will need to be supported in future releases also.

tests/transformers/models/test_causal_lm_models.py
irajagop: We essentially have 2 different API's now. One is `QEfficient.export()`, other is `QEFFAutoModel.export()`. The two are not interchangeable with each other. This test was essentially testing the old (first) export API. We're replacing the `model.export()` with new implementation. So replacing the old test with old function call.

tests/test_loader.py
irajagop: We already have test for CausalLM models. Testing for export location, qpc location, etc. This is a redundant test, hence removing.

QEfficient/base/modeling_qeff.py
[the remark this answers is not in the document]
irajagop: I've made it to logger warning instead of warning. I think it's still fine to provide this functionality, since we're providing clear indication that it may not work.

It may still be useful for some users. For example, we may need to use this if we have locally patched the transformers classes to QEff classes and want to use this class.

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,307 @@
...
+
+        try:
+            torch.onnx.export(
+                self.model,
+                (example_inputs,),
+                str(tmp_onnx_path),
+                input_names=input_names,
+                output_names=output_names,
+                dynamic_axes=dynamic_axes,
+                opset_version=13,
```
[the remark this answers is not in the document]
irajagop: I've set the value to 13, because 14 is the default value in torch 2.0. If we upgrade the version, we can upgrade here too. I've moved this to constants too.

scripts/Jenkinsfile
irajagop: We can run other tests that don't need qaic with value more than `-n 4`. So using the marker to filter, we're able to run much faster.

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,307 @@
...
+        Returns:
+            :str: Path of the compiled ``qpc`` package.
+        """
+
+    def _export(
+        self,
+        example_inputs: Dict[str, torch.Tensor],
+        output_names: List[str],
+        dynamic_axes: Dict[str, Dict[int, str]],
+        export_kwargs: Dict[str, any] = {},
```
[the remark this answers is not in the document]
irajagop: Good catch! Fixed.

QEfficient/base/modeling_qeff.py
```diff
@@ -5,96 +5,307 @@
...
+            :compiler_options: Pass any compiler option as input. Any flag that is supported by ``qaic-exec`` can be passed. Params are converted to flags as below:
+                - aic_num_cores=16 -> -aic-num-cores=16
+                - convert_to_fp16=True -> -convert-to-fp16
+
+        ``QEffAutoModelForCausalLM`` Args:
+            :batch_size (int): Batch size to compile for. ``Defaults to 1``.
+            :prefill_seq_len (int): Prefill sequence length to compile for. Prompt will be chunked according to this length.
+            :ctx_len (int): Context length to allocate space for KV-cache tensors.
+
+        ``QEffAutoModelForCausalLMwithCB`` Args:
```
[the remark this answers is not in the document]
irajagop: Fixed

tests/transformers/models/test_causal_lm_models.py
irajagop: We can update this in a different PR, as it is not related.

scripts/specializations.json
irajagop: This will work with non-CB also. It's not an error to have extra symbols listed in the specializations.json.

QEfficient/base/modeling_qeff.py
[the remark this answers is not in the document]
quic-rishinr: What if any of the transform fails? It will still show transform is applied right?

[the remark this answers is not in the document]
irajagop: By failing, you mean throwing an exception? Exceptions will directly be raised to the user. If a transform is not applicable, it will return False. If all the transforms return False (meaning nothing was applicable), it will raise the warning.

quic-rishinr (approved):

scripts/Jenkinsfile
ochougul: I understood that part, but we can still run the tests on ran by line 42 with 4 threads

tests/transformers/models/test_causal_lm_models.py
ochougul: But we are not testing the new implementation of export, compile. What I mean by that is we are not checking if onnx output with new API matches pytorch API and same for qpc. 
I am planning to make this test use the new APIs and add a test to make sure the new and old API produces same files for both onnx export and compilation.

ochougul:
There is no generate function is `QEffAutoModelForCausalLM`. We cannot just remove an API that might be being used. 
I am adding the old implementation of generate back to the class. We will later change it to match the input arguments of `generate` to `transformers` `AutoModelForCausalLM.generate` in follow up PR

ochougul (approved):

PR #125, opened 2024-09-24
Release/v1.18 : Update number of heads in MQA for Falcon

quic-mamta:
Update number of heads in MQA for Falcon, fixes falcon 40b export issue


quic-mamta:
Not required

PR #129, opened 2024-09-25
Enabled GPTQ test and doc fix

quic-amitraj:
1. Enabled the GPTQ model test
2. Fixed continues batching cmd


quic-amitraj:
Another PR raised to take of this- #136.

PR #133, opened 2024-09-27
Fixed installation issue in AArch64

quic-amitraj:
Added explicit packages with suitable versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ classifiers = [
 requires-python = "==3.8.*"
 dependencies = [
     "transformers==4.44.2",
-    "torch==2.0.0+cpu",
+    "torch==2.0.0",
     "peft==0.12.0",
     "datasets==2.7.0",
     "fsspec==2023.6.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -33,6 +33,8 @@ dependencies = [
     "protobuf==3.20.2",
     "onnxscript==0.1.0.dev20240327",
     "sympy",
+    "networkx==3.1",
```
vbaddi, line 36: These are not required, will be downloaded as part of torch.
also `torch==2.0.0` is not allowed, we need to explicitly pass `+cpu`

vbaddi (changes requested):

pyproject.toml
quic-amitraj: Yes, it should but throwing error-
image

vbaddi: Pls check the readme, these changes are not required, pls revert the PR.

pyproject.toml
```diff
@@ -33,6 +33,8 @@ dependencies = [
...
+    "multiprocess==0.70.14",
```
ochougul, line 37: add dill package, otherwise throws error on aarch64 platform

ochougul:
Waiting for Carls reply

PR #139, opened 2024-10-03
Remove `index_url` dependency and extend python version compatibility

ochougul:
Update torch whl per python version


quic-rishinr (approved):
looks good to me.

PR #145, opened 2024-10-07
Fix for hf_token passed via cli flag

asmigosw:


ochougul:
did you test the issue that was raised in the ticket for this?

asmigosw:
Yes, this change appears to be solving the issue.

QEfficient/base/common.py
```diff
@@ -82,6 +82,7 @@ def from_pretrained(cls, pretrained_model_name_or_path: str, *args, **kwargs) ->
             model_card_name = kwargs.pop("model_card_name", pretrained_model_name_or_path)
             kwargs.update({"model_card_name": model_card_name})
             pretrained_model_name_or_path = login_and_download_hf_lm(pretrained_model_name_or_path, *args, **kwargs)
+        kwargs.pop("hf_token")
```
ochougul, line 85: Is it required to pop it here again?
looks to me like, it will throw `keyError` as it was already popped inside `login_and_download_hf_lm` function

ochougul (changes requested):
We will need to push the finalized fix to main branch as well

QEfficient/base/common.py
asmigosw: Yes, we are passing the values to a separate method and it's getting poped inside the method. We still have hf_token inside kwarg in from_pretrained method.

quic-rishinr: @asmigosw better to do kwargs.pop("hf_token", None) as it would not throw a key error incase if hf_token is not available.

asmigosw: Changed to kwargs.pop("hf_token", None)

ochougul:
Please raise PR on mainline with cherry-pick thanks

PR #149, opened 2024-10-10
fix for token_type_ids

quic-mamta:
Fix token type ids in pytorch inputs for falcon 180B

### DIFF
--- a/QEfficient/utils/generate_inputs.py
+++ b/QEfficient/utils/generate_inputs.py
@@ -137,7 +137,7 @@ class InputHandler:
         input_ids = inputs["input_ids"]
         batch_size, input_len = input_ids.shape
         inputs.pop("attention_mask")
-        uts.pop("token_type_ids", None)
+        inputs.pop("token_type_ids", None)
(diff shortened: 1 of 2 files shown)


QEfficient/utils/generate_inputs.py
```diff
@@ -53,6 +53,7 @@ def prepare_pytorch_inputs(self):
         input_ids = inputs["input_ids"]
         batch_size, input_len = input_ids.shape
         inputs.pop("attention_mask")
+        inputs.pop("token_type_ids", None)
```
ochougul, line 56: don't we need to do this in prepare_ort_inputs and also in text_generation script?
Did you test the model that has this issue with infer CLI API?

ochougul (changes requested):

QEfficient/utils/generate_inputs.py
quic-mamta: I had tested with export command provided in the jira but 180B model export started but failed due to RAM issue. added the changes for ort and aic runs as well. and tested infer command for single layer model.

quic-rishinr (approved):

PR #151, opened 2024-10-10
Fixed test generation for prompt_len=1

quic-amitraj:
1. Fixed the repetitive inputs in the specializations in case of PL=1
2. Fixed vocab_size fetch, when session allowed_shape is empty.

### DIFF
--- a/QEfficient/compile/compile_helper.py
+++ b/QEfficient/compile/compile_helper.py
@@ -24,10 +24,15 @@ def create_and_dump_specializations(
                 "batch_size": str(batch_size),
                 "seq_len": str(prompt_len),
                 "ctx_len": str(ctx_len),
-            },
-            {"batch_size": str(batch_size), "seq_len": "1", "ctx_len": str(ctx_len)},
+            }
         ]
     }
+
+    # To handle repetative input in specializations when prompt_len is 1
+    if prompt_len != 1:
+        specializations["specializations"].append(
+            {"batch_size": str(batch_size), "seq_len": "1", "ctx_len": str(ctx_len)}
+        )
     # If continuous batching is enabled by proving full_batch_size we need to add FBS to the specialization file and update the batch size of decoder part to FBS
     if full_batch_size is not None:
         specializations["specializations"][0]["full_batch_size"] = str(full_batch_size)
--- a/QEfficient/generation/text_generation_inference.py
+++ b/QEfficient/generation/text_generation_inference.py
@@ -422,7 +422,10 @@ class TextGeneration:
         Returns:
             vocab_size: The vocabulary size fetched from the session's allowed shapes.
         """
-        return [x[self.session.binding_index_map["logits"]] for x in self.session.allowed_shapes][0][1][2]
+        if self.session.allowed_shapes:
+            return [x[self.session.binding_index_map["logits"]] for x in self.session.allowed_shapes][0][1][2]
+
+        return self.session.bindings[self.session.binding_index_map["logits"]].dims[2]
 
     def _fetch_generation_len(self, generation_len, max_gen_len):
         """


ochougul:
Please add a test for this, let's discuss how to do it.

ochougul (changes requested):
Change looks good, test is required!

QEfficient/compile/compile_helper.py
```diff
@@ -24,10 +24,15 @@ def create_and_dump_specializations(
...
                 "seq_len": str(prompt_len),
                 "ctx_len": str(ctx_len),
-            },
-            {"batch_size": str(batch_size), "seq_len": "1", "ctx_len": str(ctx_len)},
+            }
         ]
     }
+
+    # To handle repetative input in specializations when prompt_len is 1
+    if prompt_len != 1:
```
quic-rishinr, line 32: in this case it will throw out of index error in line number 39 and 40
        specializations["specializations"][1]["full_batch_size"] = str(full_batch_size)
        specializations["specializations"][1]["batch_size"] = str(full_batch_size)
I believe the better approach would be to remove it at the end, otherwise every time when we make change for FBS or SpD we need to account for this change.

quic-rishinr (changes requested):

QEfficient/compile/compile_helper.py
quic-amitraj: Agree, I have updated the script by adding if condition while updating specializations for full batch size.

quic-rishinr (approved):

quic-rishinr:
@quic-amitraj please rebase the PR with mainline

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -204,3 +206,30 @@ def test_causal_lm_export_with_deprecated_api(model_name):
...
+    if model_name == "microsoft/Phi-3-mini-4k-instruct":
+        n_layer = 2  # test only 2 layer models
+    else:
+        n_layer = 1
+
+    check_causal_lm_pytorch_vs_kv_vs_ort_vs_ai100(model_name=model_name, n_layer=n_layer)
+
+
+@pytest.mark.causal_lm_prompt_len_1
+def test_causal_lm_pl_1():
```
ochougul, line 228: Please update to
```
@pytest.mark.on_qaic
@pytest.mark.parametrize("model_name", test_models)
def test_causal_lm_pytorch_vs_kv_vs_ort_vs_ai100_pl1(model_name):
```

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -275,6 +275,9 @@ def compile(
                 {"batch_size": batch_size, "seq_len": 1, "ctx_len": ctx_len},
             ]
 
+            if prefill_seq_len == 1:
+                specializations.pop()
+
```
ochougul, line 280: Can you update this to 
```
        else:
            specializations = [
                {"batch_size": batch_size, "seq_len": prefill_seq_len, "ctx_len": ctx_len},
            ]

            if prefill_seq_len != 1:
                specializations.append({"batch_size": batch_size, "seq_len": 1, "ctx_len": ctx_len})
```
I think this is more robust, as in future we could have more things being added to specializations and developer could miss this if condition.
So, the first specialization is always there, and others can be added based on coditions.
In future for multimodal models, we will have more specializations.

quic-amitraj: Agree, Updated.

tests/transformers/models/test_causal_lm_models.py
quic-amitraj: Yes  I have updated the marker here and name of the test function. I think making it parametrize is not a good idea which will test with all the models in `test_models` which will be time consuming.

ochougul: Yeah, sure. Let's not parametrize the PL=1 test.

ochougul (approved):

PR #158, opened 2024-10-17
Consistency Check Failure is Fixed

abukhoy:
BUG: OSError: Consistency check failed: file should be of size {original size} but has size {downloaded size}.
We are sorry for the inconvenience. Please retry with `force_download=True`.
If the issue persists, please let us know by opening an issue on https://github.com/huggingface/huggingface_hub.

--> it is fixed now with this PR.


QEfficient/utils/constants.py
```diff
@@ -55,3 +55,4 @@ class Constants:
     INPUT_STR = ["My name is"]
     GB = 2**30
     MAX_QPC_LIMIT = 30
+    MAX_RETRIES = 5
```
quic-rishinr, line 58: Add a comment mentioning this constant will be used set the maximum number of retry attempts for downloading a model using huggingface_hub snapshot_download

QEfficient/utils/_utils.py
```diff
@@ -67,6 +67,15 @@ def hf_download(
             else:
                 raise e
 
+        except OSError as e:
+            logger.error(f"OSError: {e}")
+            if "Consistency check failed" in str(e):
+                logger.info(
+                    "OSError: Consistency check failed: file should not be incomplete, Resuming the downloading..."
```
quic-rishinr, line 74: rephrase this to something like "Consistency check failed during model download. The file appears to be incomplete. Resuming the download..."

quic-rishinr (changes requested):

QEfficient/utils/_utils.py
abukhoy: will do.

abukhoy:
@quic-rishinr  the changes have been made.

QEfficient/utils/_utils.py
```diff
@@ -42,7 +42,7 @@ def hf_download(
     if cache_dir is not None:
         os.makedirs(cache_dir, exist_ok=True)
 
-    max_retries = 5
+    max_retries = Constants.MAX_RETRIES
     retry_count = 0
     while retry_count < max_retries:
```
ochougul, line 47: directly use 
```retry_count<constants.MAX_RETRIES``` instead of redefining on line 45

**Review by ochougul — CHANGES_REQUESTED** [review]

File: QEfficient/utils/_utils.py (hunk 3 above)
Comment (abukhoy) [reply QEfficient/utils/_utils.py:47]: Okay, I will do that.

**Review by quic-amitraj — APPROVED** [review]
LGTM

File: QEfficient/utils/_utils.py [hunk 4]
```diff
@@ -61,12 +60,21 @@ def hf_download(
             retry_count += 1
 
         except HTTPError as e:
-            retry_count = max_retries
+            retry_count = Constants.MAX_RETRIES
```
quic-amitraj, line 63: I think its better to use  `break` statement instead of `retry_count = Constants.MAX_RETRIES`?

ochougul: We should raise the error here, if we only break, the execution will continue even when the model is not downloaded, and we will get `undefined variable model_path` as error.
Please change this to
```
if e.response.status_code == 401:
    logger.info("You need to pass a valid `--hf_token=...` to download private checkpoints.")
raise e
```

QEfficient/utils/_utils.py
```diff
@@ -42,9 +42,8 @@ def hf_download(
     if cache_dir is not None:
```
ochougul, line 42: can you change the function definition to take max_retries as input and set it's default value to `constants.MAX_RETRIES`

quic-rishinr: @ochougul this can be a configurable value. so it would be better if we can fetch from a file. If we need to make the max_retries as input we need to provide a way for user to set it.

quic-rishinr (approved):

PR #163, opened 2024-10-24
Fix for OOM issue in low host memory systems

asmigosw:


vbaddi:
Closing the PR, since the change is not generalized to all models!

PR #166, opened 2024-11-07
Fixed intermediate shape from projection

quic-meet:
- Fixed shape calculation from projection layer of gemma-7b which is different from intermediate shape


quic-rishinr (approved):

quic-rishinr:
Shouldn’t we run a test to run vLLM in continuous batching mode?

ochougul (approved):

PR #168, opened 2024-11-07
Vllm test

ochougul:


quic-rishinr:
Dont we need to have a test for running vLLM in continuous batching mode?

PR #174, opened 2024-11-12
Validate.md Fixed for Gemma-2 model

abukhoy:
As, there was a typing mistake initially in validate.md docs:  "google/gemma-2-7b" instead of "google/gemma-2-9b".
Now, it is fixed.


abukhoy:
Hi @quic-rishinr, Please check it once.

quic-rishinr (approved):

PR #177, opened 2024-11-15
Validate.md fixed for "Salesforce/codegen25-7b-mono_P", "Salesforce/xgen-7b-8k-base" model

abukhoy:
Unable to export ONNX for "Salesforce/codegen25-7b-mono_P", "Salesforce/xgen-7b-8k-base" model, so removed from the docs.


quic-amitraj (approved):
LGTM

PR #186, opened 2024-12-02
Jenkins Fix

quic-amitraj:


scripts/Jenkinsfile
```diff
@@ -4,6 +4,9 @@ pipeline {
            label 'qeff_node'
        }
    }
+   options {
+        disableConcurrentBuilds()
```
quic-rishinr, line 8: in this case if there are multiple commits pushed it will execute it sequentially right?

quic-rishinr (approved):

scripts/Jenkinsfile
quic-amitraj: Yes

PR #194, opened 2024-12-10
onnx_defer_loading parameter addition in QNN Convertor Command

shubhagr-quic:
A new parameter is introduced by the convertor team for optimization. Adding the same under CONVERTOR_DEFAULT_ARGS


quic-rishinr:
The flag is not supported on 1.18 or earlier version of QNN. So we wont be able to merge this change on 1.18 QNN branch. So closing the request.

PR #200, opened 2024-12-13
Migrating HL compile and export to infer APIs

asmigosw:
Migrating HL compile API and export API to infer APIs


QEfficient/cloud/infer.py
```diff
@@ -75,51 +72,27 @@ def main(
...
+        prefill_seq_len=prompt_len,
+        ctx_len=ctx_len,
+        num_cores=num_cores,
+        mxfp6_matmul=mxfp6,
+        aic_enable_depth_first=aic_enable_depth_first,
+        batch_size=batch_size,
+        mos=mos,
+        num_devices=len(device_group),
+        full_batch_size=full_batch_size,
+        allow_mxint8_mdp_io=allow_mxint8_mdp_io,
```
quic-rishinr, line 89: We should extend the support for kwargs in the infer function as well. We can add an option for compiler parameters in argparse that accepts key-value pairs and passes them to the compile function. Since this functionality is supported in our high-level APIs, we cannot limit ourselves to the standard options here.

quic-rishinr (changes requested):

quic-rishinr:
Changes moved to #214

PR #207, opened 2025-01-02
[QEff. Finetune] : To plot correct loss values on tensorboard

quic-swatia:
Plotting the correct (scaled) loss on tensorboard from the training.
Removed this change for now: Plotting the loss for each sample of eval on tensorboard rather the average.

### DIFF
--- a/QEfficient/finetune/utils/train_utils.py
+++ b/QEfficient/finetune/utils/train_utils.py
@@ -115,7 +115,6 @@ def train(
 
         # enable profile for qaic
         qaic_profile.start_profiling(device, 1) if train_config.use_profiler else None
-
         for step, batch in enumerate(train_dataloader):
             total_train_steps += 1
             #  stop when the maximum number of training steps is reached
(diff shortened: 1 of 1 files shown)


QEfficient/finetune/utils/train_utils.py
```diff
@@ -222,12 +221,14 @@ def train(
                 dist.barrier()
                 dist.all_reduce(eval_epoch_loss, op=dist.ReduceOp.SUM)
                 if local_rank == 0:
-                    tensorboard_updates.add_scalars("loss", {"val": eval_epoch_loss}, total_train_steps)
+                    for i in range(len(temp_val_loss)):
+                        tensorboard_updates.add_scalars("loss", {"eval": temp_val_loss[i]}, i)
```
quic-mamta, line 225: This will not work as it will plot the values of eval step in same range (0,len(temp_val_loss)) for each epoch.

quic-mamta, line 225: This will not work as it will plot the values of eval step in same range (0,len(temp_val_loss)) for each epoch.

quic-mamta (changes requested):

quic-mamta (approved):

quic-amitraj:
Please change the timeout for `Non CLI and non QAIC` stage in Jenkins from 10 to 15.

vbaddi (approved):

PR #216, opened 2025-01-09
Fix finite lorax generation in cb mode

quic-jouachen:
The `examples/lora_models.py` script encounters issues in cb mode. This PR addresses the following:

* Resolves the regression in finite lorax generation within cb mode in `QEfficient/generation/text_generation_inference.py` that occurred after the last refactoring.
* Adds an additional unit test in `tests/peft/lora/test_lora_model.py` to verify the compile-generate flow for finite lorax cb mode.


tests/peft/lora/test_lora_model.py
```diff
@@ -232,3 +234,28 @@ def test_auto_lora_model_for_causal_lm_export_compile_generate(base_model_name,
...
+    # test compile
+    qeff_model.compile(prefill_seq_len=32, ctx_len=64, full_batch_size=2)
+    assert Path(qeff_model.qpc_path).is_dir()
+
+    # test generate
+    prompts = ["hello!", "hi", "hello, my name is", "hey"]
+    qeff_model.generate(
+        tokenizer=load_hf_tokenizer(pretrained_model_name_or_path=base_model_name),
+        prompts=prompts,
+        device_id=[0],
```
quic-rishinr, line 259: Please remove the device ID from the generate. Generate method will auto pick the device. You can add 

if not get_available_device_id():
        pytest.skip("No available devices to run model on Cloud AI 100")

quic-jouachen: Added to the two tests that requires device_id.

quic-jouachen: Updated `QEfficient/peft/lora/auto.py` for `device_id` to be optional args. It will be passed with autodevicepicker() when default to `None` in the `text_generation_inference.py` backend

quic-rishinr (approved):

PR #228, opened 2025-01-17
Added quick support for the supported features

abukhoy:
Added documents for supported the features with the sample script hyperlink under the Quick start tab.


abukhoy:
Continuing with #240

PR #233, opened 2025-01-21
[QEff. Finetune] Adding the support to resume the fine tuning using pre computed 

quic-swatia:
Adding the support to resume the fine tuning using checkpoints from a prev run which would have stopped in between. 
There's no necessity to pass tokenizer_name if a model_name is passed. It will take the same name as model_name by default. 
If a different tokenizer_name is required than the model_name, then it can be passed separately as an argument in the command.

### DIFF
--- a/QEfficient/finetune/configs/training.py
+++ b/QEfficient/finetune/configs/training.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass
 @dataclass
 class train_config:
     model_name: str = "meta-llama/Llama-3.2-1B"
-    tokenizer_name: str = "meta-llama/Llama-3.2-1B"
+    tokenizer_name: str = None  # if not passed as an argument, it uses the value of model_name
     run_validation: bool = True
     batch_size_training: int = 1
     context_length: int = None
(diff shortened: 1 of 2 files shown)


QEfficient/finetune/configs/training.py
```diff
@@ -10,7 +10,7 @@
 @dataclass
 class train_config:
     model_name: str = "meta-llama/Llama-3.2-1B"
-    tokenizer_name: str = "meta-llama/Llama-3.2-1B"
+    tokenizer_name: str = None  # if not passed as an argument, it uses the value of model_name
```
quic-mamta, line 13: "if not passed as an argument, it uses the value of model_name" for this, assigning tokenizer_name = model_name under the condition of tokenizer_name being None is also required?

quic-mamta (changes requested):

QEfficient/finetune/configs/training.py
quic-swatia: Wherever tokenizer_name is used in the code, it checks if its value is None, it assigns it the value of model_name

quic-mamta:
If we don't change output_dir then after resuming FT, will new tensorboard data be appended to previous tensorboard log files?

quic-swatia:
Irrespective of the

quic-mamta (approved):

vbaddi (approved):

PR #238, opened 2025-01-23
Transformers version upgrade

quic-amitraj:
Upgrading transformer version to 4.46.0.


quic-amitraj:
Already addressed in #267.

PR #262, opened 2025-02-06
Support for Prefix caching Feature in QNN Compilation Path.

shubhagr-quic:


QEfficient/utils/generate_qnn_network_specialization_config.py
```diff
@@ -46,7 +46,8 @@ def fetch_nodes_info(
             if full_batch_size:
                 input_info["Shape"] = f"(1, 1), ({full_batch_size}, 1)"
             else:
-                input_info["Shape"] = "(1, 1)"
+                print("ERROR: Full batch size is required for populating batch_index in custom_io_config.yaml")
```
ochougul, line 49: better use `raise AttributeError`

ochougul (changes requested):

QEfficient/utils/generate_qnn_network_specialization_config.py
shubhagr-quic: Done

quic-rishinr (approved):

PR #268, opened 2025-02-13
Example script to run InternVL

quic-dhirajku:
A demo script to run InternVL model via Qefficient Transformers.


tests/transformers/models/intern_example_script.py
ochougul, line 1: This is not a test file. take reference of `examples/lora_models.py` file and rewrite this file so that it's intuitive for user.
Also move this under examples directory.

Remove all unused code. Follow good code practices like
1. import at top, 
2. define variables locally unless putting it global is required.
3. No print statements
4. formatting etc.

Will review once the PR is in good state.

quic-amitraj:
Better to raise now PR directly to the Main branch. Please fix DCO.

PR #274, opened 2025-02-21
HOTFIX/compiler arguments fix for VLM

quic-amitraj:
Hotfix - compiler arguments fix for VLM

As we are popping few arguments of the compiler option inside function `get_specializations` and by default `**compiler_options` is passed by value not reference, so to handle this we are returning the updated `compiler_options` from the `get_specializations`.

### DIFF
--- a/QEfficient/transformers/models/internvl/modeling_internvl.py
+++ b/QEfficient/transformers/models/internvl/modeling_internvl.py
@@ -34,7 +34,7 @@ class QEffInternVLModel(nn.Module):
             img_size = 448
             logger.warning("Setting img_size to be 448, as it was neither passed nor found in vision_config")
 
-        return [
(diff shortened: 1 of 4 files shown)


QEfficient/transformers/models/internvl/modeling_internvl.py
```diff
@@ -49,7 +49,7 @@ def get_specializations(
                 "num_patches": num_patches,
```
quic-rishinr, line 49: Suggestion: Although its acceptable to return it this way. But its always better to assign this to a variable and then return it. Code becomes more readable that way. This applies to both  get_specializations methods.

quic-rishinr (approved):

quic-rishinr (approved):

PR #279, opened 2025-02-24
Added support of 2qpcs for internvl and llava

mohiso22:


QEfficient/transformers/models/modeling_auto.py
```diff
@@ -417,9 +423,12 @@ class QEffCausalLMForTextImageToTextModel(QEFFBaseModel):
         VlmKVOffloadTransform,
     ]
     _onnx_transforms = [FP16ClipTransform, SplitTensorsTransform]
+    MODELS_WITH_LANGUAGE_MODEL_AS_SECOND_QPC = ["LlavaForConditionalGeneration", "InternVLChatModel"]
```
quic-amitraj, line 426: Make a single global variable of early fusion models and use that on both the places-`MODELS_WITH_LANGUAGE_MODEL_AS_SECOND_QPC` and `EARLYFUSION_MODELS`.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -417,9 +423,12 @@ class QEffCausalLMForTextImageToTextModel(QEFFBaseModel):
...
 
     def __init__(self, model):
         super().__init__(model)
+        if self.model_name in self.MODELS_WITH_LANGUAGE_MODEL_AS_SECOND_QPC:
+            self.model = self.model.language_model
```
quic-amitraj, line 431: Can you create a same function like `get_qeff_vision_endcoder`-> `get_language_model` inside the modeling file of all the VLM, which will besicall return self.language_model incase of internvl and llava and self.model in case of mllama.

quic-amitraj (changes requested):

QEfficient/transformers/models/modeling_auto.py
quic-amitraj: No need  of this if you will address my comment of creating function`get_language_model`.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -678,16 +682,22 @@ def kv_offload_generate(
             [x for x in lang_session.input_names + lang_session.output_names if x.startswith("past_")]
         )
 
+        if self.model_name in self.EARLYFUSION_MODELS:
+            input_ids = "inputs_embeds"
```
quic-amitraj, line 686: There is no need of these codes, as you can use existing code to fetch the batch size and prefill seq len.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -723,10 +733,14 @@ def kv_offload_generate(
             k: v for k, v in inputs.items() if k in {"pixel_values", "aspect_ratio_ids", "aspect_ratio_mask"}
         }
 
+        if input_ids == "inputs_embeds":
```
quic-amitraj, line 736: As `vision_inputs` also needed `input_ids` then handle this inside the line 732, where we are extracting vision inputs from the `inputs`.

QEfficient/transformers/models/internvl/modeling_internvl.py
```diff
@@ -90,21 +155,28 @@ def get_dummy_inputs(self, kv_offload: bool = False):
...
         )
         inputs_shapes["pixel_values"] = (num_patches, C, img_size, img_size)
 
         # Define inputs
-        inputs = {}
-        inputs["input_ids"] = torch.zeros((inputs_shapes["input_ids"]), dtype=torch.int64)
-        inputs["position_ids"] = (
+        vision_inputs = {}
+        lang_inputs = {}
+        vision_inputs["input_ids"] = torch.zeros((inputs_shapes["input_ids"]), dtype=torch.int64)
```
quic-amitraj, line 172: Better to create vision input first then create lang input. It will be more readable.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -723,10 +733,14 @@ def kv_offload_generate(
...
+            vision_inputs["input_ids"] = inputs["input_ids"]
         vision_inputs["pixel_values"] = vision_inputs["pixel_values"].astype("float16")
         vision_outputs = vision_session.run(vision_inputs)
 
         lang_inputs = {k: v for k, v in inputs.items() if k not in vision_inputs}
+        if input_ids == "inputs_embeds":
+            lang_inputs["inputs_embeds"] = vision_outputs["inputs_embeds"].astype("float32")
```
quic-amitraj, line 743: Same way, handle this.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -739,7 +753,7 @@ def kv_offload_generate(
         # Run prefill
         for i in range(num_chunks):
             chunk_inputs = lang_inputs.copy()
-            chunk_inputs["input_ids"] = lang_inputs["input_ids"][:, i * prefill_seq_len : (i + 1) * prefill_seq_len]
+            chunk_inputs[input_ids] = lang_inputs[input_ids][:, i * prefill_seq_len : (i + 1) * prefill_seq_len]
```
quic-amitraj, line 756: No need.

quic-amitraj (changes requested):

QEfficient/transformers/models/internvl/modeling_internvl.py
```diff
@@ -50,61 +107,88 @@ def get_specializations(
...
+        vision_output_names = ["vit_embeds"]
+        lang_output_names = ["logits", "pixel_values_RetainedState", "vit_embeds_RetainedState"]
         for i in range(self.language_model.config.num_hidden_layers):
             for kv in ["key", "value"]:
-                output_names.append(f"past_{kv}.{i}_RetainedState")
+                lang_output_names.append(f"past_{kv}.{i}_RetainedState")
+
+        output_names = {}
+        if kv_offload:
+            lang_output_names.pop(1)
```
ochougul, line 150: instead of popping. insert the value based on kv_offload flag.
This makes code more intuitive and readable

QEfficient/transformers/models/internvl/modeling_internvl.py
```diff
@@ -50,61 +107,88 @@ def get_specializations(
...
+            output_names["vision"] = vision_output_names
+            output_names["lang"] = lang_output_names
+        else:
+            lang_output_names.pop(2)
```
ochougul, line 154: same as above.

QEfficient/transformers/models/llava/modeling_llava.py
```diff
@@ -103,32 +180,53 @@ def get_specializations(
...
+        vision_output_names = ["image_features"]
+        lang_output_names = ["logits", "pixel_values_RetainedState", "image_features_RetainedState"]
         for i in range(self.language_model.config.num_hidden_layers):
             for kv in ["key", "value"]:
-                output_names.append(f"past_{kv}.{i}_RetainedState")
+                lang_output_names.append(f"past_{kv}.{i}_RetainedState")
+
+        output_names = {}
+        if kv_offload:
+            lang_output_names.pop(1)
```
ochougul, line 224: same comment as for intern

ochougul (changes requested):
Rebase

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -356,7 +356,13 @@ def pytorch_feature_generate(self, model, inputs: Union[torch.Tensor, np.ndarray
...
+    _pytorch_transforms = [
+        AwqToMatmulNbitsTransform,
+        GPTQToMatmulNbitsTransform,
+        CustomOpsTransform,
+        KVCacheTransform,
+        KVCacheModuleMethodMapperTransform,
+    ]
     _onnx_transforms = [FP16ClipTransform, SplitTensorsTransform]
 
     def __init__(self, model: nn.modules):
```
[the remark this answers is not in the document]
quic-amitraj: Agree, but I think here we cannot do this, as here `model` is without transform`,  so `model.get_qeff_vision_encoder()` will through an error, but this can be done at the time of language at line 429.

QEfficient/transformers/models/internvl/modeling_internvl.py
```diff
@@ -14,9 +14,57 @@
...
+
+
+class QEffInternLanguageDecoder2QPC(nn.Module):
+    def __init__(self, model):
+        super().__init__()
+        self.model = model
+        self.config = self.model.language_model.config
+
+    def forward(self, input_ids, vit_embeds, position_ids, past_key_values):
+        # TODO: Check if Hardcoding this is okay, i.e. check if this value is common for all intern models
```
quic-hemagnih, line 34: Why are we hardcoding and using magic numbers? How do we handle it in case of single QPC?

QEfficient/transformers/models/llava/modeling_llava.py
```diff
@@ -50,33 +101,50 @@ def forward(self, input_ids, position_ids, pixel_values, past_key_values):
...
         return outputs.logits, pixel_values, outputs.past_key_values
 
-    def get_dummy_inputs(self, **kwargs):
+    def get_dummy_inputs(self, kv_offload: bool = False, **kwargs):
         num_layers = self.config.text_config.num_hidden_layers
         num_key_value_heads = self.config.text_config.num_key_value_heads
         head_dim = self.config.text_config.hidden_size // self.config.text_config.num_attention_heads
         if vis_cfg := getattr(self.config, "vision_config", None):
             img_size = getattr(vis_cfg, "image_size", 336)
         else:
```
quic-hemagnih, line 110: why are we using magic numbers in the code?

QEfficient/transformers/models/llava/modeling_llava.py
```diff
@@ -50,33 +101,50 @@ def forward(self, input_ids, position_ids, pixel_values, past_key_values):
...
             img_size = 336
-        inputs = {
+        vision_inputs = {
+            "pixel_values": torch.zeros((BS, NUM_CHANNEL, img_size, img_size), dtype=torch.float32),
+        }
+        lang_inputs = {
             "input_ids": torch.ones((BS, SEQ_LEN), dtype=torch.int64),
+            "image_features": torch.ones((BS, 576, self.language_model.config.hidden_size), dtype=torch.float32),
```
quic-hemagnih, line 117: same as above for 576?

QEfficient/transformers/models/internvl/modeling_internvl.py
mohiso22: This is the image token_id, and it is same for all the Internmodels available so we are directly keeping that here

QEfficient/transformers/models/llava/modeling_llava.py
mohiso22: U mean to say img_size as 336 ?

QEfficient/transformers/models/llava/modeling_llava.py
mohiso22: So if we fix the img_size to 336, it will create a shape of (BS,576,hs), and if we don't fix 336 as img_size, we might have to run the vision part inside this function to get the shapes of image_features

quic-hemagnih: So let say user changes the img_size in config.json and then execute it. Then hardcoding shape of 576 won't work correct? Shouldn't we discuss and plan for a generic approach rather having support for a very specific config.

mohiso22: Yes you are correct, we can try for a generic approach, and should discuss about it.

QEfficient/transformers/models/internvl/modeling_internvl.py
```diff
@@ -14,9 +14,57 @@
...
+        input_embeds = self.model.language_model.get_input_embeddings()(input_ids)
+        B, N, C = input_embeds.shape
+        image_input_embeds = input_embeds.reshape(B * N, C)
+        image_input_ids = input_ids.reshape(B * N)
+        selected = image_input_ids == IMG_CONTEXT_TOKEN
+        indices1 = selected.unsqueeze(0).to(torch.int64).cumsum(1) - 1
+        indices0 = torch.arange(selected.unsqueeze(0).shape[0]).view(-1, 1)
+        image_features_expanded = vit_embeds.reshape(-1, C).unsqueeze(0)[indices0, indices1]
+        image_input_embeds = torch.where(selected.unsqueeze(0).unsqueeze(-1), image_features_expanded, input_embeds)
+        inputs_embeds = torch.where(input_ids.shape[1] == torch.tensor(1), input_embeds, image_input_embeds)
```
quic-rishinr, line 46: During decode phase the fusing image and vision embeddings is not required. its unnecessary computing from line 38 to 45 and then selecting input_embeds if the input ids shape is 1. Can we optimize this?

quic-rishinr (changes requested):

QEfficient/transformers/models/internvl/modeling_internvl.py
quic-hemagnih: I think compiler shall optimize this piece of code, so based on line#46 it would execute only the required code. 
Please correct me if my understanding is wrong.

QEfficient/transformers/models/modeling_auto.py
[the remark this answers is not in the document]
ochougul: Okay, let's keep it like this for now. in both vision and causal classes.
Later we will create `Mixin` classes to fix this by applying transform in the `AutoImageText` class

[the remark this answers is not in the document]
mohiso22: For the vision I had kept the same but for the lang, I had done in one line

QEfficient/transformers/models/internvl/modeling_internvl.py
mohiso22: It all depends on how compiler is creating the graph.

quic-rishinr (approved):

PR #281, opened 2025-02-24
Updated mos to optional argument

asmigosw:
Made mos to optional argument and passed None as default.


QEfficient/cloud/infer.py
```diff
@@ -22,7 +22,7 @@ def main(
     prompt: Optional[str] = None,  # type: ignore
     prompts_txt_file_path: Optional[str] = None,
     aic_enable_depth_first: bool = False,
-    mos: int = -1,
+    mos: Optional[int] = None,
```
vbaddi, line 25: should be 1 by default (recommended for LLMs)

vbaddi (approved):
nit: small change, rest LGTM. Thanks :)

QEfficient/cloud/infer.py
quic-rishinr: Do we need to set a default value from Qeff side? the idea was if user pass any value, we can add it to compilation command otherwise we should let the compiler default settings take effect

vbaddi: Yes, better to keep default to 1 then None and override based on user.

asmigosw: Updated mos default value to 1.

ochougul:
Please rebase

quic-rishinr (approved):

PR #287, opened 2025-02-27
Enabled Infer CLI for VLM

asmigosw:
Added support for enabling VLMs via CLI.

Sample command:
```bash
python -m QEfficient.cloud.infer --model_name meta-llama/Llama-3.2-11B-Vision-Instruct 
--batch_size 1 --prompt_len 32 --ctx_len 512 --num_cores 16 --device_group [0]
--prompt "Descrive the image?" --mos 1  --allocator_dealloc_delay 1 
--image_url https://i.etsystatic.com/8155076/r/il/0825c2/1594869823/il_fullxfull.1594869823_5x0w.jpg
```


QEfficient/transformers/models/modeling_auto.py
```diff
@@ -818,14 +820,17 @@ def kv_offload_generate(
...
-        return CloudAI100ExecInfoNew(
+        exec_info = CloudAI100ExecInfoNew(
             batch_size=batch_size,
             generated_ids=generated_ids,
             perf_metrics=PerfMetrics(
                 prefill_time=prefill_time, decode_perf=decode_perf, total_perf=total_perf, total_time=total_time
             ),
         )
 
+        print(exec_info)
```
ochougul, line 831: remove

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1113,14 +1118,17 @@ def cloud_ai_100_generate(
...
```
ochougul, line 1129: remove

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1498,6 +1509,9 @@ def compile(
             decode_specialization.update({"num_logits_to_keep": num_speculative_tokens + 1}) if self.is_tlm else ...
             specializations.append(decode_specialization)
 
+        if compiler_options.pop("img_size", None):
+            logger.warning("img_size is not a valid argument for Text-to-Text Model.")
+
```
ochougul, line 1514: Why? out compiler will fail if user passes img_size .
Let it fail

ochougul (changes requested):

QEfficient/transformers/models/modeling_auto.py
asmigosw: This is printing the performance numbers, don't we need it ?

QEfficient/transformers/models/modeling_auto.py
asmigosw: Since this img_size is passed as an args to compiler which is not an acceptable argument by Causal Models, it throws error as Invalid option, so this needs to be popped out. And while popping if its value is not None then throwing a warning.

quic-amitraj: We can throw an error with the proper error message.

QEfficient/cloud/infer.py
```diff
@@ -85,6 +89,9 @@ def main(
         if args.mxint8:
             logger.warning("mxint8 is going to be deprecated in a future release, use -mxint8_kv_cache instead.")
 
+    image_path = kwargs.pop("image_path", None)
+    image_url = kwargs.pop("image_url", None)
```
quic-amitraj, line 93: Is there any reason, we are making not making this as optional arguments? We can make this as optional argument so-
1. There will be no need of popping
2. We will have uniform argument format `image_path` and `image-path` same like other arguments.
3. Then we can properly document our arguments.

QEfficient/cloud/infer.py
```diff
@@ -110,20 +117,70 @@ def main(
         allow_mxint8_mdp_io=allow_mxint8_mdp_io,
         enable_qnn=enable_qnn,
         qnn_config=qnn_config,
+        img_size=img_size,
         **kwargs,
     )
 
+    tokenizer = load_hf_tokenizer(
```
quic-amitraj, line 124: Move these line after 132, as its getting used inside `Execute`.

QEfficient/cloud/infer.py
```diff
@@ -110,20 +117,70 @@ def main(
...
     #########
-    _ = qeff_model.generate(
-        tokenizer,
-        prompts=prompt,
-        device_id=device_group,
-        prompt=prompt,
-        prompts_txt_file_path=prompts_txt_file_path,
-        generation_len=generation_len,
-    )
+    config = AutoConfig.from_pretrained(model_name)
```
quic-amitraj, line 133: There is no need of the AutoConfig, as you can access the model config using qeff_model.model.config.

quic-amitraj (changes requested):

QEfficient/base/common.py
```diff
@@ -42,9 +50,13 @@ def from_pretrained(cls, pretrained_model_name_or_path: str, *args, **kwargs) ->
         config = AutoConfig.from_pretrained(pretrained_model_name_or_path)
         architecture = config.architectures[0] if config.architectures else None
 
-        if architecture in MODEL_FOR_CAUSAL_LM_MAPPING_NAMES.values():
-            model_class = QEFFAutoModelForCausalLM
-        else:
+        model_class = None
+        for key_tuple, class_name in MODEL_CLASS_MAPPING.items():
+            if architecture in key_tuple:
```
quic-hemagnih, line 55: Atleast have a TODO or FIXME tag for optimizations, I think we can come up with a way where we can avoid the for loops and fetch the architecture in O(1) time complexity.

QEfficient/transformers/models/modeling_auto.py
quic-amitraj: Instead of this to print the perf, you can print perf after the calling the generate inside the infer.

QEfficient/base/common.py
```diff
@@ -42,9 +50,13 @@ def from_pretrained(cls, pretrained_model_name_or_path: str, *args, **kwargs) ->
         config = AutoConfig.from_pretrained(pretrained_model_name_or_path)
         architecture = config.architectures[0] if config.architectures else None
 
-        if architecture in MODEL_FOR_CAUSAL_LM_MAPPING_NAMES.values():
-            model_class = QEFFAutoModelForCausalLM
-        else:
+        model_class = None
```
quic-amitraj, line 53: Here, Please add a condition that if `image_size` is passed and `model_class` is not belonging to the `MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES` or ` MISCLASSIFIED_CAUSAL_LM_TO_QEFF_AUTO_CLASS_MAP`, then throw an error with proper warning. This will handle both the condition of throwing an error for `AutoModelForCausalLM` and `AutoModelForSpeechSeq2Seq`.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -601,6 +601,8 @@ def compile(
             )
 
         output_names = self.model.get_output_names(kv_offload=True)
+        vision_onnx_path = compiler_options.get("vision_onnx_path", None)
```
quic-amitraj, line 604: Why, already we have arguments in the compile?

QEfficient/cloud/infer.py
```diff
@@ -110,20 +117,70 @@ def main(
...
+        if image_url is not None:
+            raw_image = Image.open(requests.get(image_url, stream=True).raw)
+        elif image_path is not None:
+            raw_image = Image.open(image_path)
+        else:
+            raise FileNotFoundError(
+                'Neither Image URL nor Image Path is found, either provide "image_url" or "image_path"'
+            )
+
+        conversation = [
```
quic-amitraj, line 143: Is this work for all the different types of VLM models?

QEfficient/cloud/infer.py
```diff
@@ -110,20 +117,70 @@ def main(
...
+            add_special_tokens=False,
+        )
+        streamer = TextStreamer(processor.tokenizer)
+        output = qeff_model.generate(
+            inputs=split_inputs,
+            streamer=streamer,
+            device_ids=device_group,
+            generation_len=generation_len,
+        )
+        print(output)
```
quic-amitraj, line 169: I think, as you are already passing the `streamer`, no need of printing whole output. Just print the perf mat as we have discussed?

QEfficient/cloud/infer.py
```diff
@@ -110,20 +117,70 @@ def main(
...
-    )
+    config = qeff_model.model.config
+    architecture = config.architectures[0] if config.architectures else None
+
+    if architecture in MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES.values():
+        processor = AutoProcessor.from_pretrained(model_name, use_fast=False)
+
+        raw_image = None
+        if image_url is not None:
+            raw_image = Image.open(requests.get(image_url, stream=True).raw)
```
quic-amitraj, line 135: Handle both image_url and image-url

quic-amitraj (changes requested):

QEfficient/cloud/infer.py
asmigosw: It is not printing the whole output; this only prints the perf metric. If you want to print whole output, you need to use `tokenizer.batch_decode` for generated ids from output.

QEfficient/cloud/infer.py
asmigosw: As we have enabled 3 models out of which it works for mllama and llava, for InternVL extra support of processor needs to be added so haven't tested for it.

QEfficient/transformers/models/modeling_auto.py
asmigosw: As we have passed vision_onnx_path to the extra compiler, so here it needs to be taken from the compiler_options and overwrite the values.

quic-amitraj: Better to handle this inside the infer file only, initialize the vision and lang_path inside the infer only despite of user is passing or not.

quic-amitraj (approved):
LGTM

QEfficient/cloud/infer.py
```diff
@@ -116,14 +129,47 @@ def main(
...
-        prompts=prompt,
-        device_id=device_group,
-        prompt=prompt,
-        prompts_txt_file_path=prompts_txt_file_path,
-        generation_len=generation_len,
-    )
+    if architecture in MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES.values():
+        processor = AutoProcessor.from_pretrained(model_name, use_fast=False)
+
+        if not (image_url or image_path):
```
quic-hemagnih, line 135: Can we move this check at line 133, before calling AutoProcessor.from_pretrained()

QEfficient/cloud/infer.py
```diff
@@ -116,14 +129,47 @@ def main(
...
-    _ = qeff_model.generate(
-        tokenizer,
-        prompts=prompt,
-        device_id=device_group,
-        prompt=prompt,
-        prompts_txt_file_path=prompts_txt_file_path,
-        generation_len=generation_len,
-    )
+    if architecture in MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES.values():
+        processor = AutoProcessor.from_pretrained(model_name, use_fast=False)
```
quic-hemagnih, line 133: Nitpick - I would suggest to break the code in multiple functions, rather than increasing the size of single function. It shall increase the complexity during debug and code browsing. 
Here probably we can make two functions - one we can call at line 133 and other for the else case of line 160

QEfficient/cloud/infer.py
```diff
@@ -10,11 +10,53 @@
...
+    device_group: Optional[List[int]] = None,
+    generation_len: Optional[int] = None,
+):
+    if not (image_url or image_path):
+        raise ValueError('Neither Image URL nor Image Path is found, either provide "image_url" or "image_path"')
+    raw_image = Image.open(requests.get(image_url, stream=True).raw) if image_url else Image.open(image_path)
+
+    processor = AutoProcessor.from_pretrained(model_name, use_fast=False)
+
+    conversation = constants.Constants.conversation
```
ochougul, line 38: import directly at the top of the file and make caps

ochougul: also deepcopy this to avoid race condition

ochougul: better move conversation string here. makes things simpler

QEfficient/cloud/infer.py
```diff
@@ -93,6 +133,16 @@ def main(
         local_model_dir=local_model_dir,
     )
 
+    image_path = kwargs.pop("image_path", None)
+    image_url = kwargs.pop("image_url", None)
+
+    config = qeff_model.model.config
+    architecture = config.architectures[0] if config.architectures else None
+    if architecture not in MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES.values():
```
ochougul, line 141:
```
if architecture not in MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES.values() and "img_size" in kwargs and image and image_path, and image_url:
    kwargs.pop()
    warning
```

QEfficient/cloud/infer.py
```diff
@@ -116,14 +175,31 @@ def main(
...
+            hf_token=hf_token,
+        )
+        _ = qeff_model.generate(
+            tokenizer,
+            prompts=prompt,
+            device_id=device_group,
+            prompt=prompt,
+            prompts_txt_file_path=prompts_txt_file_path,
+            generation_len=generation_len,
+        )
```
ochougul, line 202: this was supposed to be simplified right>?
Combine with if on line 150?
something like set execution function above and then use it here?

ochougul: see we handle all params and set execution function in single if condition and only call it here. 
It's not much actually we only need to handle image_path or url. params other params are common.

This is minor change, but should be done.
In interest of time, I will approve for now and merge once CI passes, Do this change if you can before that.

ochougul (changes requested):
NO tests are added here to test the new code.
Please add tests. Or need to test manually.

Post the results of the manual run before merging.
And this will be merged at the end so we don't have to test this again and again unless tests are added in CLI.

QEfficient/cloud/infer.py
asmigosw: We took `image_path` or `url` through `extra_compiler_args`, so they have to be removed/handled before sending kwargs to compile in case of non-vlm (causal) models otherwise we will get compilation error of "not known args passed". Therefore, that first `if` statement was mandatory.

Second `if' statement is necessary because the arguments we are sending to vlm's generate and causal's generate are different, which could not be made generic.

QEfficient/base/common.py
```diff
@@ -42,8 +41,10 @@ def from_pretrained(cls, pretrained_model_name_or_path: str, *args, **kwargs) ->
         config = AutoConfig.from_pretrained(pretrained_model_name_or_path)
         architecture = config.architectures[0] if config.architectures else None
 
-        if architecture in MODEL_FOR_CAUSAL_LM_MAPPING_NAMES.values():
-            model_class = QEFFAutoModelForCausalLM
+        class_name = MODEL_CLASS_MAPPING.get(architecture)
```
quic-hemagnih, line 44: I would suggest to create a utility function for extracting the class name from maps and define it in modelling_utils.py so that this peice of code becomes reusable for future also. You can take arguments as architecture, Model_Class_Mapping and QEfficient.transformers.models.modeling_auto and return the model_class

QEfficient/cloud/infer.py
```diff
@@ -10,11 +10,61 @@
...
+from transformers import AutoProcessor, PreTrainedModel, TextStreamer
+from transformers.models.auto.modeling_auto import MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES
+
 from QEfficient.base.common import QEFFCommonLoader
 from QEfficient.utils import check_and_assign_cache_dir, load_hf_tokenizer
 from QEfficient.utils.logging_utils import logger
 
 
+def execute_vlm_model(
+    qeff_model: PreTrainedModel,
```
quic-hemagnih, line 24: Pase add doc string and hint type for function arguments to increase the code readability.

QEfficient/cloud/infer.py
```diff
@@ -219,23 +295,25 @@ def main(
...
-        "qnn_config",
-        nargs="?",
-        type=str,
-    )
 
     args, compiler_options = parser.parse_known_args()
+
+    if isinstance(args.enable_qnn, str):
+        args.qnn_config = args.enable_qnn
+        args.enable_qnn = True
```
quic-hemagnih, line 311: Why we are doing enable_qnn = TRUE here?

asmigosw: This code checking for if enable_qnn is passed as string implies qnn_config is passed along with it, then directly make enable_qnn arg as True and fill qnn_config with the config string. 
Since whenever qnn_config is passed, enable_qnn has to be true for it work.

So, making both true and passing config was a redundant code earlier.

abukhoy: Because, if CLI argument `enable_qnn ` passed as string with the value of qnn config file path. And this `if` block creates two arguments `qnn_config` would take file path and `enable_qnn` would take bool value to indicate whether qnn is enabled or not.
@asmigosw correct me if I am wrong.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1118,13 +1119,14 @@ def cloud_ai_100_generate(
         total_time = decode_end - prefill_start
         total_perf = num_token / total_time
 
-        return CloudAI100ExecInfoNew(
+        exec_info = CloudAI100ExecInfoNew(
```
abukhoy, line 1122: Is this `exec_info` variable needed?  Is directly return not possible?

abukhoy (commented):
I think, we should add one small multimodal model under CLI api testing.
https://github.com/quic/efficient-transformers/blob/a706a0198ff67eec2abdb72994aa7da903ca86b4/tests/cloud/high_level_testing.json#L3

quic-hemagnih (approved):

quic-hemagnih:
TODO:
Refactoring Infer for compile and generate
Making CLI tests generic for all modelling classes by refactoring "conftest" and "json" file.

quic-hemagnih (approved):

QEfficient/cloud/infer.py
```diff
@@ -93,6 +156,17 @@ def main(
         local_model_dir=local_model_dir,
     )
 
+    image_path = kwargs.pop("image_path", None)
+    image_url = kwargs.pop("image_url", None)
+
+    config = qeff_model.model.config
+    architecture = config.architectures[0] if config.architectures else None
+
+    if architecture not in MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES.values() and (
```
quic-rishinr, line 165: architecture can be also None. its better to make code this way
`image_path = kwargs.pop("image_path", None)
image_url = kwargs.pop("image_url", None)
img_size = kwargs.pop("img_size", None)

config = qeff_model.model.config
architecture = config.architectures[0] if config.architectures else None
if architecture is None:
    logger.warning("Architecture is not defined in the model configuration.")
if architecture not in MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES.values() and (img_size or image_path or image_url):
        logger.warning(f"Skipping image arguments as they are not valid for {architecture}")`

QEfficient/cloud/infer.py
```diff
@@ -10,11 +10,76 @@
...
+    prompt: Optional[str] = None,  # type: ignore
+    device_group: Optional[List[int]] = None,
+    generation_len: Optional[int] = None,
+):
+    """
+    This method generates output by executing the compiled ``qpc`` on ``Cloud AI 100`` Hardware cards.
+    ``Mandatory`` Args:
+        :qeff_model (PreTrainedModel): QEfficient model object.
+        :model_name (str): Hugging Face Model Card name, Example: ``llava-hf/llava-1.5-7b-hf``
+        :num_cores (int): Number of cores to compile model on.
```
quic-amitraj, line 37: I don't see num_cores in parameters. Please remove that.

QEfficient/cloud/infer.py
```diff
@@ -10,11 +10,75 @@
...
 
 
+def execute_vlm_model(
+    qeff_model: PreTrainedModel,
+    model_name: str,
+    image_url: str,
+    image_path: str,
+    prompt: Optional[str] = None,  # type: ignore
+    device_group: Optional[List[int]] = None,
+    generation_len: Optional[int] = None,
```
quic-amitraj, line 30: Please add TODO here that, we have to remove this function asa we will start supporting execute of the VLM through CI, which will eventually replace this.

QEfficient/cloud/infer.py
asmigosw: Done

QEfficient/cloud/infer.py
```diff
@@ -10,11 +10,75 @@
...
+        :generation_len (int): Number of tokens to be generated. ``Defaults to None.``
+    Returns:
+        :dict: Output from the ``AI_100`` runtime.
+    """
+    if not (image_url or image_path):
+        raise ValueError('Neither Image URL nor Image Path is found, either provide "image_url" or "image_path"')
+    raw_image = Image.open(requests.get(image_url, stream=True).raw) if image_url else Image.open(image_path)
+
+    processor = AutoProcessor.from_pretrained(model_name, use_fast=False)
+
```
quic-amitraj, line 51: No need to import Autoprocessor here, use load_hf_processor.

quic-amitraj (changes requested):

QEfficient/cloud/infer.py
```diff
@@ -116,14 +189,31 @@ def main(
...
-        prompts=prompt,
-        device_id=device_group,
-        prompt=prompt,
-        prompts_txt_file_path=prompts_txt_file_path,
-        generation_len=generation_len,
-    )
+    if architecture in MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES.values():
+        exec_info = execute_vlm_model(
+            qeff_model=qeff_model,
+            model_name=model_name,
```
quic-amitraj, line 195: TODO: Here use `load_hf_processor` and `load_steamer` to load processor and streamer. Create a list of conversation in _utill that will be mapped with the model architecture. At the end use `qeff_model.generate` as else condition. This way this code will be more scalable and well formatted.Then there will be no need of function `execute_vlm_model`.

quic-amitraj (approved):
LGTM

PR #291, opened 2025-02-27
Adding support for registaration of non transformer models like swiftkv in QEfficient

quic-hemagnih:
Adding support for registaration of non transformer models like swiftkv in QEfficient


QEfficient/__init__.py
```diff
@@ -1,19 +1,37 @@
...
 
+from QEfficient.transformers.modeling_utils import (
+    MODEL_TYPE_TO_CONFIG_CLS_AND_ARCH_CLS,
+    get_auto_model_class,
+    get_model_class_type_from_model_type,
+)
 from QEfficient.utils.logging_utils import logger
 
+# loop over all the models which are not present in transformers and register them
+for key, value in MODEL_TYPE_TO_CONFIG_CLS_AND_ARCH_CLS.items():
```
quic-hemagnih, line 25: rename the key and value[0] and value[1] - so that readability increases

ochougul:
copy of #314

PR #297, opened 2025-03-03
Enabled VLMs via CLI on v1.19.3

asmigosw:
Added support for enabling VLMs via CLI.

Sample command:

`python -m QEfficient.cloud.infer --model_name meta-llama/Llama-3.2-11B-Vision-Instruct 
--batch_size 1 --prompt_len 32 --ctx_len 512 --num_cores 16 --device_group [0]
--prompt "Descrive the image?" --mos 1  --allocator_dealloc_delay 1 
--image_url https://i.etsystatic.com/8155076/r/il/0825c2/1594869823/il_fullxfull.1594869823_5x0w.jpg`


QEfficient/base/common.py
```diff
@@ -42,8 +49,10 @@ def from_pretrained(cls, pretrained_model_name_or_path: str, *args, **kwargs) ->
         config = AutoConfig.from_pretrained(pretrained_model_name_or_path)
         architecture = config.architectures[0] if config.architectures else None
 
-        if architecture in MODEL_FOR_CAUSAL_LM_MAPPING_NAMES.values():
-            model_class = QEFFAutoModelForCausalLM
+        class_name = MODEL_CLASS_MAPPING.get(architecture)
+        if class_name:
+            module = importlib.import_module("QEfficient.transformers.models.modeling_auto")
```
ochougul, line 54: do we need to use `importlib` here? 
Is it causing circular import without it?

QEfficient/base/common.py
```diff
@@ -12,13 +12,20 @@
...
 
 from QEfficient.base.modeling_qeff import QEFFBaseModel
-from QEfficient.transformers.models.modeling_auto import QEFFAutoModelForCausalLM
+
+MODEL_CLASS_MAPPING = {}
+for architecture in mapping.MODEL_FOR_CAUSAL_LM_MAPPING_NAMES.values():
+    MODEL_CLASS_MAPPING[architecture] = "QEFFAutoModelForCausalLM"
+
+for architecture in mapping.MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES.values():
+    MODEL_CLASS_MAPPING[architecture] = "QEFFAutoModelForImageTextToText"
```
ochougul, line 28: can we resort to simple technique like if the architecture name has suffix `lm_head` or `causalLM` it's text only.
instead if it has `ConditionalGeneration` it's image-text.
We can take a look at all suffixes in the above maps and decide this logic?

QEfficient/cloud/infer.py
```diff
@@ -110,20 +117,70 @@ def main(
         allow_mxint8_mdp_io=allow_mxint8_mdp_io,
         enable_qnn=enable_qnn,
         qnn_config=qnn_config,
+        img_size=img_size,
```
ochougul, line 120: Doesn't this fail when we pass image_size and the model is causalLM?

ochougul: We can choose to drop such extra params in case of causalLM like `image_url` , `img_size` etc.

QEfficient/cloud/infer.py
```diff
@@ -110,20 +117,70 @@ def main(
...
+
+        raw_image = None
+        if image_url is not None:
+            raw_image = Image.open(requests.get(image_url, stream=True).raw)
+        elif image_path is not None:
+            raw_image = Image.open(image_path)
+        else:
+            raise FileNotFoundError(
+                'Neither Image URL nor Image Path is found, either provide "image_url" or "image_path"'
+            )
```
ochougul, line 141: You can add a `nor` condition and fail there. also if both are passed, we should issue warning saying one of them will be ignored.

QEfficient/cloud/infer.py
```diff
@@ -110,20 +117,70 @@ def main(
...
+
+        conversation = [
+            {
+                "role": "user",
+                "content": [
+                    {"type": "image"},
+                    {"type": "text", "text": prompt[0]},  # Currently accepting only 1 prompt
+                ],
+            },
+        ]
```
ochougul, line 151: Is this common for any image-text model. at least for the ones that we support?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -603,6 +603,8 @@ def compile(
             )
 
         output_names = self.model.get_output_names(kv_offload=True)
+        vision_onnx_path = compiler_options.get("vision_onnx_path", None)
+        lang_onnx_path = compiler_options.get("lang_onnx_path", None)
```
ochougul, line 607: why? aren't those already parameters?

QEfficient/utils/constants.py
```diff
@@ -136,7 +136,6 @@ class QnnConstants:
         "--float_bitwidth ",
         "--preserve_io_datatype",
         "--onnx_skip_simplification",
-        "--onnx_defer_loading",
```
ochougul, line 139: why is this being changed in this PR?

ochougul (changes requested):

QEfficient/cloud/infer.py
```diff
@@ -226,10 +283,11 @@ def main(
              Sample Config: QEfficient/compile/qnn_config.json",
     )
     parser.add_argument(
-        "qnn_config",
+        "--qnn_config",
```
quic-rishinr, line 286: In the current workflow qnn_config is taken as an optional argument. Changing it to a positional argument would disrupt the existing flow. Instead we should pass a constant value of True for --enable_qnn and remove the qnn_config argument. @shubhagr-quic any thoughts on it?

QEfficient/cloud/infer.py
```diff
@@ -110,20 +117,70 @@ def main(
...
+            add_special_tokens=False,
+        )
+        streamer = TextStreamer(processor.tokenizer)
+        output = qeff_model.generate(
+            inputs=split_inputs,
+            streamer=streamer,
+            device_ids=device_group,
+            generation_len=generation_len,
+        )
+        print(output)
```
quic-rishinr, line 169: We should not be printing it this way in infer. Can we set a verbose level and print it accordingly from Auto classes itself @ochougul @quic-amitraj ?

QEfficient/cloud/infer.py
quic-rishinr: Also should we keep it here? can we keep it in the constants file?

QEfficient/cloud/infer.py
```diff
@@ -226,10 +283,11 @@ def main(
              Sample Config: QEfficient/compile/qnn_config.json",
     )
     parser.add_argument(
-        "qnn_config",
+        "--qnn_config",
         nargs="?",
         type=str,
     )
+    parser.add_argument("--img-size", "--img_size", default=None, type=int, required=False, help="Size of Image")
```
quic-rishinr, line 290: if we are taking image path and image url as kwarg its better to keep it consistent. There is no point in only providing img_size as arg.

quic-rishinr (changes requested):

QEfficient/cloud/infer.py
asmigosw: Yes, this condition was discussed on original PR #287, we have popped out the image_size before passing it to compile of CausalLM models.

QEfficient/cloud/infer.py
asmigosw: Yes, this works for both Llava and MLlama.

QEfficient/transformers/models/modeling_auto.py
asmigosw: These are params, but if user passed them then it will overwrite them with user's input

QEfficient/base/common.py
asmigosw: This will generate ambiguity as even in `MODEL_FOR_CAUSAL_LM_MAPPING_NAMES` we have architecture which have `ConditionalGeneration` in their names, eg : [https://github.com/huggingface/transformers/blob/6966fa190172b48b2fb46fe4552a13b943e692cf/src/transformers/models/auto/modeling_auto.py#L523](url)

Also, there are many different architecture names in `MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES`, therefore mapping to all these names would not be possible as in future more architectures might add with different names.

quic-rishinr:
Closing this PR as this feature will be merged into mainline.

PR #303, opened 2025-03-04
Docs string added for the Image class and granite models are added in validation page

abukhoy:


QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1150,6 +1150,59 @@ class QEFFAutoModelForImageTextToText:
     A factory class for creating QEFFAutoModelForImageTextToText instances with for single and Dual QPC approach
```
quic-rishinr, line 1150: Can you also update the doc string and add the args supported in pretrained?

quic-rishinr (changes requested):

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1147,9 +1147,69 @@ def get_model_config(self) -> dict:
...
+    The QEFFAutoModelForImageTextToText class is used to work with multimodal language models from the HuggingFace hub.
+    While you can initialize the class directly, it's best to use the ``from_pretrained`` method for this purpose. This class supports both single and dual QPC approaches.
     Attributes:
         _hf_auto_class (class): The Hugging Face AutoModel class for ImageTextToText models.
+
+    ``Mandatory`` Args:
+        :pretrained_model_name_or_path (str): Model card name from HuggingFace or local path to model directory.
+
+    ``Optional`` Args:
+        :kv_offload (bool): Should the KV of vision encoder be offloaded to CPU and use Two QPC. Defaults to None.
```
quic-rishinr, line 1159: Update the description to 
 :kv_offload (bool): Flag to toggle between single and dual QPC approaches. If set to `False`, the Single QPC approach will be used; otherwise, the dual QPC approach will be applied. Defaults to `True`.

quic-rishinr (changes requested):

QEfficient/transformers/models/modeling_auto.py
quic-hemagnih, line 1: Please correct the below documentation, 
https://quic.github.io/efficient-transformers/source/quick_start.html

For example,
Section : 2. Export and Compile with one API
generated_qpc_path = qeff_model.compile(num_cores=14,mxfp6=True,)

Similarly there are options like generation_len missing in the generate API.

abukhoy: `generation_len` is the kwargs param, Do we need to add it here? just a doubt.

quic-rishinr: @quic-hemagnih its a sample script we are providing there. If you can navigate to https://quic.github.io/efficient-transformers/source/python_api.html#QEfficient.transformers.models.modeling_auto.QEFFAutoModelForCausalLM.compile we have all the params updated there.

quic-rishinr (approved):

quic-hemagnih (approved):
Looks good to me

PR #304, opened 2025-03-04
VLM Test pipeline added

quic-akuruvil:
- VLM Test pipeline for `ImageTextToText` Models
- Tests the token match on diff platforms 
- Current models in test config - Llava, Mllama, Intern
- classes for input generation in ONNXRT, Pytorch and AIC platforms
- classes for inference in different platforms


quic-amitraj:
This requirement will be addressed in PR #337

PR #306, opened 2025-03-04
SpD, multiprojection heads

eplatero97:
### Objective
This PR seeks to implement post-attention hidden size projections used to speculate tokens ahead of the base model. This PR contains three primary components:

1. extending base model with multi-projection in `modeling_auto.py`
2. implementing multi-projections forward pass.
3. app demo of multi-projection model.

### Initial Implementation
Initial implementation gives the user the flexibility to define its own projection architecture and pass it to `QEffAutoModelForCausalLM`. 

Then, QEfficient simply attaches these projections to the model to be used during the forward pass. The "attaching" of these projections is done by using the `accelerate` library. I used this library because it has a robust implementation to attach weights to an already existing model. We can implement our own abstraction if needed, but first we must agree on what the external API will be to the user.

> NOTE: Please keep in mind that to integrate medusa, similar changes will be needed (instead of doing multiple hidden size projections, medusa uses multiple `lm_heads` to speculate ahead of the base model).


eplatero97:
using sdk `1.20.0.63`, I manually validated that all qwen vanilla spd tests passed as well as llama multiprojection pytorch unit test

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -85,6 +93,19 @@ def from_pretrained(cls, pretrained_model_name_or_path: str, is_tlm: bool = Fals
...
 
         model = cls._hf_auto_class.from_pretrained(pretrained_model_name_or_path, *args, **kwargs)
+        if hidden_size_projections is not None:
+            if isinstance(hidden_size_projections, tuple):
+                projs, checkpoint = hidden_size_projections
+                assert isinstance(projs, nn.ModuleList)
+                assert isinstance(checkpoint, str)
+                model.hidden_size_projections = load_checkpoint_and_dispatch(
+                    projs, checkpoint=checkpoint, device_map="auto"
+                )
```
ochougul, line 103: Do we need to use accelerate here?
device_map is mainly for multi-device eager execution.
We re doing AOT here

QEfficient/transformers/spd/causal_lm_forward.py
```diff
@@ -113,7 +135,10 @@ def tlm_forward(
     )
 
     hidden_states = filter_hidden_states(outputs[0], position_ids, num_logits_to_keep)
-    if self.config.pretraining_tp > 1:
+    hidden_size_projections = getattr(self, "projections", None)
```
quic-rishinr, line 138: are these still applicable?

quic-rishinr (changes requested):

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1375,6 +1387,10 @@ def from_pretrained(
 
         kwargs.update({"attn_implementation": "eager", "low_cpu_mem_usage": False})
         model = cls._hf_auto_class.from_pretrained(pretrained_model_name_or_path, *args, **kwargs)
+        speculative_config = None
```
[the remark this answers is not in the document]
eplatero97: if I remove this and `speculative_model` is None, this will result in error because `speculative_config` will not be defined. 

we could switch `speculative_model` to `speculative_config` to resolve it however (this is similar to how vllm does it)

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1375,6 +1387,10 @@ def from_pretrained(
...
+        if speculative_model is not None:
```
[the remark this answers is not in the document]
eplatero97: replaced the `assert` statements with proper exceptions. 

As for using init instead of `from_pretrained`, can you provide an explicit example of how this would look like? 

I'm open to the idea but need to see the benefit of this approach while also ensuring that this approach has an easy way to look for a configuration file that does not have the standard `config.json` naming convention

QEfficient/transformers/spd/__init__.py
```diff
@@ -5,3 +5,18 @@
 #
 # -----------------------------------------------------------------------------
 
+from QEfficient.transformers.spd.causal_lm_forward import tlm_forward
+from QEfficient.transformers.spd.turbo import build_and_attach_turbo
+
+
+def build_and_attach_mlp(model, config):
```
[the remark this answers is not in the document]
eplatero97: definitely not necessary to keep this logic here. I agree that we can move this somewhere else. However, I don't think placing this in `utils` is the best approach. 

The idea is that we have some directory here that will have all the builds for all future attached spd modules. similar to something like below:
```bash
some_spd_directory
├───turbo.py
├───medusa.py
├───eagle.py
├───...
```

I'm thinking maybe creating a directory like:
`QEfficient/modules`
?

QEfficient/transformers/spd/causal_lm_forward.py
eplatero97: yes, this is actually how the model is "aware" that it has projections to apply to the last hidden_size. is there something in particular that you want to change here?

QEfficient/transformers/models/modeling_auto.py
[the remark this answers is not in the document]
quic-rishinr: instead of passing speculative_config we can pass the speculative_model itself to init and from there we can fetch the config.

QEfficient/transformers/spd/__init__.py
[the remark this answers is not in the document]
quic-rishinr: lets create a post_processing.py file under transforms and lets maintain it there.

QEfficient/utils/helper_utils.py
```diff
@@ -0,0 +1,23 @@
...
+#
+# Copyright (c) 2024 Qualcomm Innovation Center, Inc. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# -----------------------------------------------------------------------------
+
+import inspect
+
+
+def filter_kwargs(func, kwargs):
```
quic-rishinr, line 11: keep it under QEfficient/utils/_utils.py

QEfficient/transformers/spd/causal_lm_forward.py
quic-rishinr, line 1: Please rename the file name

quic-rishinr (changes requested):

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1504,9 +1528,17 @@ def compile(
             :str: Path of the compiled ``qpc`` package.
         """
         if self.is_tlm:
-            # assert num_speculative_tokens cfg is acceptable if defined
             if num_speculative_tokens is None:
-                raise TypeError("missing required argument `num_speculative_tokens` as `is_tlm` is True.")
+                if self.num_speculative_tokens is not None:
```
ochougul, line 1532: create a new function as explained below to keep it simple
```Python
    def check_and_get_num_logits_to_keep(self, num_speculative_tokens, prefill_seq_len):
        if hasattr(self.model.config, "speculative_config"):
            if num_speculative_tokens is not None:
                logger.warning(f"arg `num_speculative_tokens` is a fixed value of {self.num_speculative_tokens} for this model."
                " Passed value will be ignored.")
            num_speculative_tokens = self.config.speculative_config.num_speculative_tokens
        elif num_speculative_tokens is None:
            raise TypeError("missing required argument `num_speculative_tokens` as `is_tlm` is True.")

        if not isinstance(num_speculative_tokens, int) and num_speculative_tokens < 2:
            ValueError(
                f"`num_speculative_tokens` arg should be an integer greater than 1, got {num_speculative_tokens}"
            )
        num_logits_to_keep = num_speculative_tokens + 1
        if prefill_seq_len < num_logits_to_keep:
            raise ValueError(
                f"sequence length ({prefill_seq_len}) must be at least `num_speculative_tokens+1` ({num_logits_to_keep})"
            )
        return num_logits_to_keep
```

and use it as
```Python
        if self.is_tlm:
            num_logits_to_keep = self.check_and_get_num_logits_to_keep(num_speculative_tokens=num_speculative_tokens,
                                                                       prefill_seq_len=prefill_seq_len)
```
use the same value for filling specialization

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -390,12 +391,12 @@ class SpDTransform:
...
     # supported architectures
     _module_mapping = {
-        # Llama
         QEffLlamaForCausalLM,
+        QEffQwen2ForCausalLM,
     }
 
     @classmethod
-    def apply(cls, model: nn.Module) -> Tuple[nn.Module, bool]:
+    def apply(cls, model: nn.Module, speculative_config: Optional[dict] = None) -> Tuple[nn.Module, bool]:
```
ochougul, line 399: take speculative_model instead of speculative_config and attach config to the model inside this method

ochougul (changes requested):

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1393,6 +1412,11 @@ def model_hash(self) -> str:
         mhash.update(to_hashable({"continuous_batching": self.continuous_batching}))
         mhash.update(to_hashable({"is_tlm": self.is_tlm}))
         mhash.update(to_hashable(self._transform_names()))
+        if self.speculative_config is not None:
+            speculative_config = self.speculative_config.copy()
+            if "speculative_weights" in speculative_config:
+                del speculative_config["speculative_weights"]
+            mhash.update(to_hashable(speculative_config))
```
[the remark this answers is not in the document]
eplatero97: done. since `speculative_config` is now at `model.config.speculative_config`, we do not need an extra hash, we get it by default

QEfficient/transformers/models/pytorch_transforms.py
eplatero97: passing `speculative_config` thorugh `kwargs` to add more flexibility

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -1,13 +1,14 @@
 # -----------------------------------------------------------------------------
 #
-# Copyright (c) 2025 Qualcomm Innovation Center, Inc. All rights reserved.
+# Copyright (c) 2024 Qualcomm Innovation Center, Inc. All rights reserved.
```
quic-hemagnih, line 3: why 2024?

QEfficient/utils/checkpoint_utils.py
```diff
@@ -0,0 +1,27 @@
...
+    Args:
+        model: model to load wights into
+        checkpoint (str): checkpoint path
+        strict (bool, optional): strictness of loading weights. Defaults to False.
+        post_process_func (optional): Optional post-processing of loaded state dict. Defaults to None.
+
+    Returns:
+        model: model with applied weights
+    """
+    state_dict: dict = load_file(checkpoint)
```
quic-hemagnih, line 23: Should we add error handling here, in case if file load fails

eplatero97: Hey Hem, what kind of error handling are you thinking? Currently, if a file does not exist, it throws below error:
```bash
>>> from QEfficient.utils.checkpoint_utils import load_checkpoint
>>> load_checkpoint(1, './file.txt')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/prj/crd/austin/validation/scratch/users/eplatero/qefficient_spd/efficient-transformers/QEfficient/utils/checkpoint_utils.py", line 23, in load_checkpoint
    state_dict: dict = load_file(checkpoint)
  File "/prj/crd/austin/validation/scratch/users/eplatero/qefficient_spd/efficient-transformers/qeff_env/lib/python3.10/site-packages/safetensors/torch.py", line 313, in load_file
    with safe_open(filename, framework="pt", device=device) as f:
FileNotFoundError: No such file or directory: "./file.txt"
```

and if the file can't open, it will throw below:
```bash
>>> load_checkpoint(1, './README.md')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/prj/crd/austin/validation/scratch/users/eplatero/qefficient_spd/efficient-transformers/QEfficient/utils/checkpoint_utils.py", line 23, in load_checkpoint
    state_dict: dict = load_file(checkpoint)
  File "/prj/crd/austin/validation/scratch/users/eplatero/qefficient_spd/efficient-transformers/qeff_env/lib/python3.10/site-packages/safetensors/torch.py", line 313, in load_file
    with safe_open(filename, framework="pt", device=device) as f:
safetensors_rust.SafetensorError: Error while deserializing header: HeaderTooLarge
```

Maybe to have a cleaner handle on the latter case, we can add a check that before loading the file, we assert the extension is of `safetensors` type?

ochougul:
Please rebase

quic-hemagnih:
Is the CI passing now?

quic-rishinr:
CI is passing for this change, but the VLLM stage is expected to fail. The changes need to be added into VLLM as well to make it functional, which will be done once our changes are merged.

ochougul (approved):

quic-hemagnih (approved):

quic-hemagnih:
Please merge the changes both in Release 1.20 branch and mainline

PR #322, opened 2025-03-19
GraniteCausalLM Fix with customNorm addition

qcdipankar:
Absent of customrmsnorm was causing GraniteCausalLM to fail in aic with full model in 4.46.3

Addition of CustomRMSNormAIC fixes the issue

### DIFF
--- a/QEfficient/transformers/models/pytorch_transforms.py
+++ b/QEfficient/transformers/models/pytorch_transforms.py
@@ -48,6 +48,7 @@ from transformers.models.granite.modeling_granite import (
     GraniteAttention,
     GraniteForCausalLM,
     GraniteModel,
+    GraniteRMSNorm,
 )
 from transformers.models.llama.modeling_llama import (
     LlamaAttention,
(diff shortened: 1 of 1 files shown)


QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -301,6 +303,7 @@ class KVCacheTransform(ModuleMappingTransform):
         Gemma2Model: QEffGemma2Model,
         Gemma2ForCausalLM: QEffGemma2ForCausalLM,
         # Granite
+        GraniteRMSNorm: CustomRMSNormAIC,
```
ochougul, line 306: not required to do here.

ochougul (changes requested):

QEfficient/transformers/models/pytorch_transforms.py
qcdipankar: okay will remove from here and commit back on this

qcdipankar: @ochougul why is it there for mllama?

ochougul (approved):

quic-hemagnih (approved):

PR #324, opened 2025-03-21
Revert "GraniteCausalLM Fix with customNorm addition"

quic-rishinr:
Reverts quic/efficient-transformers#322


ochougul:
Did the change in main directly.
This way we still reserve the commit history.
Doing directly in main does not save history in main branch.

PR #326, opened 2025-03-25
Context length specialization

quic-vjanfaza:
Context-Length-Specialization technique tries to optimize the throughput of large language models (LLMs) on Qualcomm devices when handling very large context lengths. The current Ahead Of Time (AOT) compilation on Qualcomm devices doesn't predict the number of tokens needed, leading to significant throughput drops during the prefilling and the decoding phases. This happens because the system performs attention calculations based on large context length. To address this issue, we introduce Compute Context Length (CCL), an additional ONNX variable that allows for dynamic context-length specialization. By generating tokens using smaller, more manageable context lengths (CCL), we optimize memory reads and attention calculations, thereby improving throughput.

### DIFF
--- a/QEfficient/transformers/models/modeling_auto.py
+++ b/QEfficient/transformers/models/modeling_auto.py
@@ -1298,7 +1298,7 @@ class QEFFAutoModelForCausalLM(QEFFBaseModel):
         self,
         model: nn.Module,
         continuous_batching: bool = False,
-        is_tlm: bool = False,
+        qaic_config: Optional[dict] = None,
         **kwargs,
     ):
         model_class_name = model.__class__.__name__
@@ -1324,11 +1324,8 @@ class QEFFAutoModelForCausalLM(QEFFBaseModel):
         self.model.config.use_cache = True
         self.num_layers = model.config.num_hidden_layers
         self.continuous_batching = continuous_batching
-
-        if is_tlm:
-            # TODO: It is possible to always apply this transform and make value of indices as last indices by default in PyTorch
-            self.model, transformed = SpDTransform.apply(self.model)
-        self.is_tlm = is_tlm
+        self.model, transformed = SpDTransform.apply(self.model, qaic_config, **kwargs)
+        self.is_tlm = transformed
 
     @property
     def model_name(self) -> str:
@@ -1343,7 +1340,12 @@ class QEFFAutoModelForCausalLM(QEFFBaseModel):
     @classmethod
     @with_replaced_quantizers
     def from_pretrained(
-        cls, pretrained_model_name_or_path, continuous_batching: bool = False, is_tlm: bool = False, *args, **kwargs
+        cls,
+        pretrained_model_name_or_path,
+        continuous_batching: bool = False,
+        qaic_config: Optional[dict] = None,
+        *args,
+        **kwargs,
     ):
         """
         This method serves as the easiest entry point into using QEfficient. The interface is designed to be similar to transformers.AutoModelForCausalLM.
@@ -1388,6 +1390,8 @@ class QEFFAutoModelForCausalLM(QEFFBaseModel):
 
         kwargs.update({"attn_implementation": "eager", "low_cpu_mem_usage": False})
         model = cls._hf_auto_class.from_pretrained(pretrained_model_name_or_path, *args, **kwargs)
+        if qaic_config is not None:
+            qaic_config["pretrained_model_name_or_path"] = pretrained_model_name_or_path
 
         # This is support models that should be classified to in a different auto class but transformers load them via this class
 
@@ -1396,7 +1400,12 @@ class QEFFAutoModelForCausalLM(QEFFBaseModel):
                 model, kv_offload=kv_offload
             )
 
-        return cls(model, is_tlm=is_tlm, continuous_batching=continuous_batching)
+        return cls(
+            model,
+            continuous_batching=continuous_batching,
+            qaic_config=qaic_config,
+            **kwargs,
+        )
 
     @property
     def model_hash(self) -> str:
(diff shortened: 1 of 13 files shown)


examples/example_ccl.py
```diff
@@ -0,0 +1,48 @@
...
+
+# Create tokenizer and run model.generate and passs the input prompts to it. It also recieves comp_ctx_lengths list which will be used during the decoding process to apply the best and most efficient compute context length.
+tokenizer = AutoTokenizer.from_pretrained(model_name)
+model.generate(
+    prompts=[
+        "What are some healthy foods to include in a balanced diet?",
+    ],
+    tokenizer=tokenizer,
+    comp_ctx_lengths=comp_ctx_lengths,
+    generation_len=128,
```
quic-rishinr, line 46: if you are giving GL=128 ccl would not kick in right? Do we need to have GL in the example? also please remove the device IDs.

tests/cloud/test_ccl.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# -----------------------------------------------------------------------------
+
+## In this example, you can run a model for static and continuous batching with different Compute-Context-Length (CCL) inputs. ##
+
+from transformers import AutoTokenizer
+
+from QEfficient import QEFFAutoModelForCausalLM
+
+model_name = "meta-llama/Llama-3.1-8B-Instruct"
```
quic-rishinr, line 14: can we use smaller model? like llama 3.2 1B?

tests/cloud/test_ccl.py
```diff
@@ -0,0 +1,48 @@
...
+#     comp_ctx_lengths=comp_ctx_lengths,
+#     num_cores=16,
+#     num_devices=4,
+#     full_batch_size=4,
+# )
+model.compile(
+    prefill_seq_len=128, ctx_len=8192, comp_ctx_lengths=comp_ctx_lengths, num_cores=16, num_devices=4, batch_size=4
+)
+
+# Create tokenizer and run model.generate and passs the input prompts to it. It also recieves comp_ctx_lengths list which will be used during the decoding process to apply the best and most efficient compute context length.
```
quic-rishinr, line 38: typo on pass and receive.

tests/cloud/test_ccl.py
```diff
@@ -0,0 +1,48 @@
...
+tokenizer = AutoTokenizer.from_pretrained(model_name)
+model.generate(
+    prompts=[
+        "What are some healthy foods to include in a balanced diet?",
+    ],
+    tokenizer=tokenizer,
+    comp_ctx_lengths=comp_ctx_lengths,
+    generation_len=128,
+    device_ids=[16, 17, 18, 19],
```
quic-rishinr, line 47: refrain from providing device ID in test. auto device picker will pick the devices during the run. Also can we run the test on single device?

tests/cloud/test_ccl.py
```diff
@@ -0,0 +1,48 @@
...
+)
```
quic-rishinr, line 48: is the test incomplete? 
I believe we should also check if the tokens which are getting generated are right? can you refer to tests/transformers/models/test_causal_lm_models.py and update the test?
also please move the test to tests/transformers

quic-rishinr (changes requested):

quic-rishinr:
please update the CCL in the documentation as well under [[efficient-transformers](https://github.com/quic/efficient-transformers/tree/main)/[docs](https://github.com/quic/efficient-transformers/tree/main/docs)/[source](https://github.com/quic/efficient-transformers/tree/main/docs/source)
/quick_start.md](https://github.com/quic/efficient-transformers/blob/main/docs/source/quick_start.md#supported-features)

QEfficient/generation/text_generation_inference.py
```diff
@@ -808,18 +835,38 @@ def run_continuous_batching_decode(self, prompt_queue, generation_len):
...
                             ]
 
+                        if self.comp_ctx_lengths is not None:
+                            ###Recalculate ccl_id based on position ids###
+                            # Determine the maximum value of position_ids across all batch elements
+                            max_position_id = np.max(decode_inputs["position_ids"])
+
+                            # Update ccl_id and comp_ctx_lengths based on the maximum position id
+                            ccl_id = 1
+                            for i in range(1, len(self.comp_ctx_lengths)):
```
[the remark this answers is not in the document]
quic-vjanfaza: The length of ccl list can't be more than a certain range because of the compiler limit and more specializations have negative effect on the performance. I don't think implementing a heap structure for a limited list is beneficial here.

examples/example_ccl.py
quic-vjanfaza: If we set the GL, ccl would work up to that generation length. For example, when GL=128, the model will work with the smallest ccl in the list. However, I will remove the generation length and device IDs from example.

QEfficient/generation/text_generation_inference.py
```diff
@@ -808,18 +835,38 @@ def run_continuous_batching_decode(self, prompt_queue, generation_len):
                                 batch_id_map[decode_batch_id]
                             ]
 
+                        if self.comp_ctx_lengths is not None:
+                            ###Recalculate ccl_id based on position ids###
+                            # Determine the maximum value of position_ids across all batch elements
+                            max_position_id = np.max(decode_inputs["position_ids"])
```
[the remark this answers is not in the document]
quic-vjanfaza: Here, we are looking for the maximum position id between all requests in the batch. During continuous batching, I don't think the value of decode_inputs["position_ids"][decode_batch_id, -1] is the maximum position ID between all request in the current batch.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1402,7 +1402,7 @@ def model_hash(self) -> str:
     def get_model_config(self) -> dict:
         return self.model.config.__dict__
 
-    def export(self, export_dir: Optional[str] = None) -> str:
+    def export(self, comp_ctx_lengths: Optional[List[int]] = None, export_dir: Optional[str] = None) -> str:
```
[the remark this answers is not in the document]
quic-vjanfaza: For second command, because of the compiler limitations the number of specializations can't be more than a certain range and more specializations have negative effect on the performance. So, passing a fixed list of the most efficient context lengths would be the best use case here.

PR #331, opened 2025-03-27
Upgrading transformers version to latest 4.50

quic-amitraj:
1. Upgrading transformers version to latest 4.50
2. Code cleaned
3. Fixed issue of `DynamicCache`

### DIFF
--- a/QEfficient/transformers/models/codegen/modeling_codegen.py
+++ b/QEfficient/transformers/models/codegen/modeling_codegen.py
@@ -12,7 +12,6 @@ from typing import Optional, Tuple, Union
 import torch
 import torch.utils.checkpoint
 from torch import nn
-from torch.nn import CrossEntropyLoss
 from transformers.cache_utils import Cache, DynamicCache
 from transformers.modeling_outputs import BaseModelOutputWithPast, CausalLMOutputWithPast
 from transformers.models.codegen.modeling_codegen import (
@@ -21,7 +20,6 @@ from transformers.models.codegen.modeling_codegen import (
     CodeGenForCausalLM,
     CodeGenModel,
     apply_rotary_pos_emb,
-    logger,
 )
 
 from QEfficient.transformers.modeling_attn_mask_utils import _create_causal_mask
@@ -261,14 +259,6 @@ class QEffCodeGenModel(CodeGenModel):
 
         output_shape = input_shape + (hidden_states.size(-1),)
 
-        if self.gradient_checkpointing and self.training:
-            if use_cache:
-                logger.warning_once(
-                    "`use_cache=True` is incompatible with `config.gradient_checkpointing=True`. Setting "
-                    "`use_cache=False`..."
-                )
-                use_cache = False
-
         if position_ids is None:
             position_ids = cache_position.unsqueeze(0)
 
@@ -279,19 +269,7 @@ class QEffCodeGenModel(CodeGenModel):
             if output_hidden_states:
                 all_hidden_states = all_hidden_states + (hidden_states,)
 
-            if self.gradient_checkpointing and self.training:
-                outputs = self._gradient_checkpointing_func(
-                    block.__call__,
-                    hidden_states,
-                    None,
-                    attention_mask,
-                    position_ids,
-                    head_mask[i],
-                    use_cache,
-                    output_attentions,
-                    cache_position,
-                )
-            elif batch_index is not None:
+            if batch_index is not None:
                 outputs = block(
                     hidden_states=hidden_states,
                     layer_past=layer_past,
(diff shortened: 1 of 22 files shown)


vbaddi (approved):

QEfficient/transformers/models/whisper/modeling_whisper.py
```diff
@@ -630,26 +576,13 @@ def _update_causal_mask(
         past_key_values: Cache,
         output_attentions: bool,
     ):
-        if self.config._attn_implementation == "flash_attention_2":
```
quic-amitraj, line 633: Is there any specific reason that this function is still here?

ochougul:
- Remove unused imports from all modelling files.
- Remove unused params from all forward methods or raise warnings
- There are un-necessary if conditions for `attention_mask` and `past_key_values`
    But we know that we don't support `attention_mask` so it will always be `None` and we always need `past_key_value` so it will always be passed. so these if condition can be removed.

QEfficient/transformers/models/codegen/modeling_codegen.py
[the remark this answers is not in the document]
ochougul: Also device related code

QEfficient/transformers/models/gemma/modeling_gemma.py
```diff
@@ -39,17 +36,12 @@ class QEffGemmaRotaryEmbedding(GemmaRotaryEmbedding):
...
 
-    def __init__(self, dim, max_position_embeddings=2048, base=10000, device=None):
-        super(GemmaRotaryEmbedding, self).__init__()
-        self.dim = dim
-        self.max_position_embeddings = max_position_embeddings
-        self.base = base
-        inv_freq = 1.0 / (self.base ** (torch.arange(0, self.dim, 2, dtype=torch.int64).float().to(device) / self.dim))
-        self.register_buffer("inv_freq", inv_freq, persistent=False)
+    def __init__(self, config: GemmaConfig, device=None):
+        GemmaRotaryEmbedding.__init__(self, config=config)
```
ochougul, line 40: use super

QEfficient/transformers/models/gemma/modeling_gemma.py
```diff
@@ -115,47 +130,32 @@ def __init__(self, config: GemmaConfig, layer_idx: Optional[int] = None):
         super().__init__(config, layer_idx)
         # Define the general __qeff_init__() for any changes in the init calls
         # Set the init in the module mapping pytorch transforms
+        self.config = config
         self.__qeff_init__()
```
ochougul, line 134: should be removed

QEfficient/transformers/models/gemma/modeling_gemma.py
```diff
@@ -432,50 +315,24 @@ def forward(
         # decoder layers
         all_hidden_states = () if output_hidden_states else None
         all_self_attns = () if output_attentions else None
-        next_decoder_cache = None
 
-        for decoder_layer in self.layers:
+        for decoder_layer in self.layers[: self.config.num_hidden_layers]:
```
ochougul, line 319: revert

ochougul: We don't want use to just set this attribute and change the behavior of the model

ochougul (changes requested):
most of the comments apply to multiple files. please change the code in all files that follow same structure

QEfficient/transformers/quantizers/quantizer_compressed_tensors.py
```diff
@@ -130,12 +130,20 @@ def forward(self, x):
 
 class QEffFP8Config(QuantizationConfigMixin):
     def __init__(
-        self, quant_method: str, activation_scheme: str, ignored_layers: List[str] = None, kv_cache_scheme: str = None
+        self,
+        quant_method: str,
+        activation_scheme: str,
+        ignored_layers: List[str] = None,
+        kv_cache_scheme: str = None,
+        run_compressed: bool = True,
```
ochougul, line 138: why?

QEfficient/transformers/quantizers/quantizer_compressed_tensors.py
```diff
@@ -130,12 +130,20 @@ def forward(self, x):
...
     ):
         self.quant_method = quant_method
         self.activation_scheme = activation_scheme
         self.ignored_layers = ignored_layers
         self.kv_cache_scheme = kv_cache_scheme
+        self.run_compressed = run_compressed
```
ochougul, line 144: why?

QEfficient/transformers/quantizers/quantizer_compressed_tensors.py
```diff
@@ -156,7 +164,7 @@ def __init__(self, quantization_config, **kwargs):
             raise TypeError(f"Only {QEffFP8Config} is supported for initialization got {type(quantization_config)}")
 
         self.quantization_config = quantization_config
-
+        self.run_compressed = quantization_config.run_compressed
```
ochougul, line 167: why?

QEfficient/transformers/quantizers/quantizer_compressed_tensors.py
ochougul, line 1: why is `run_compresed, quantization_config, sparsity_config` being added here?

ochougul (changes requested):

QEfficient/transformers/quantizers/quantizer_compressed_tensors.py
quic-amitraj: So there are few function in super class ex- `update_missing_keys_after_loading` which is using these variables and we have not overwrite that in the child class. So to overcome that we are just following the same approach of tf to assign self.run_compressed from quantization_config and same follow to another variables as well.

quic-amitraj:
Key Note of this PR-

1. Upgrade: Updated TensorFlow from version 4.46.3 to 4.50.0.
2. Code Cleaning and Optimization: Performed code cleaning and optimizations across all modeling files to some extent. Further changes will be addressed in upcoming PRs to enhance the code quality.

QEfficient/transformers/quantizers/quantizer_compressed_tensors.py
```diff
@@ -226,6 +238,10 @@ def __init__(
         self.global_compression_ratio = global_compression_ratio
         self.ignore = ignore
 
+        self.quantization_config = None
+        self.sparsity_config = None
+
+        self.run_compressed = run_compressed
```
ochougul, line 244: In our case this should be `False`, according to transformers this is set to True when we want to emulate compressed execution, i.e. actual matmul in FP8 , but we never do that, we always run decompressed module

quic-amitraj:
The further code cleaning of attention class, unused parameters will be done in upcoming PR.

ochougul (approved):
Approving now. But few things need to be addressed.
1. don't use if condition on `attention_mask` and `past_key_value` because we know their default values
2. remove unused params from forward or raise warning
3. change year to 2025 on top

quic-amitraj:
Thanks @ochougul, all the suggested and pending changes will be addressed in upcoming PR.

PR #337, opened 2025-04-02
VLM e2e test pipeline 

quic-akuruvil:


quic-rishinr:
@quic-dhirajku as discussed offline please create a new tag for VLM test in jenkins.

tests/transformers/models/test_image_text_to_text_models.py
```diff
@@ -0,0 +1,324 @@
...
+from QEfficient.utils._utils import get_num_layers_vlm
+from QEfficient.utils.device_utils import get_available_device_id
+from QEfficient.utils.run_utils import ApiRunnerInternVL, ApiRunnerVlm
+from QEfficient.utils.test_utils import InternProcessor
+
+HF_TOKEN = ""
+NEW_GENERATION_TOKENS = 10
+test_models_config = [
+    (
+        "llava-hf/llava-1.5-7b-hf",
```
quic-amitraj, line 34: Add comment here like this-
 "llava-hf/llava-1.5-7b-hf", # Model_name
 1, # batch_size

tests/transformers/models/test_image_text_to_text_models.py
```diff
@@ -0,0 +1,324 @@
...
+    n_layer: int = 1,
+    kv_offload: bool = False,
+    num_devices: int = 1,
+):
+    model_config = {"model_name": model_name}
+    model_config["img_size"] = img_size
+    config = AutoConfig.from_pretrained(
+        model_config["model_name"], token=HF_TOKEN, trust_remote_code=True, padding=True
+    )
+    config._attn_implementation = "eager"
```
quic-amitraj, line 130: No need of this as we are always update the attention to eager

tests/transformers/models/test_image_text_to_text_models.py
```diff
@@ -0,0 +1,324 @@
...
+    streamer = TextStreamer(processor.tokenizer)
+    pytorch_hf_tokens = api_runner.run_vlm_hf_model_on_pytorch(model_hf, inputs)
+    qeff_model = QEFFAutoModelForImageTextToText.from_pretrained(
+        model_config["model_name"],
+        kv_offload=kv_offload,
+        config=config,
+        token=HF_TOKEN,
+    )
+
+    # pytorch_kv_tokens = api_runner.run_vlm_kv_model_on_pytorch(qeff_model.model)
```
quic-amitraj, line 170: Remove these commented codes, if not needed.

QEfficient/utils/test_utils.py
```diff
@@ -0,0 +1,146 @@
...
+
+import torch
+import torch.nn as nn
+import torchvision.transforms as T
+from torchvision.transforms.functional import InterpolationMode
+
+
+# Processor class for InternVL models
+class InternProcessor:
+    def __init__(self, model: nn.Module, tokenizer):
```
quic-amitraj, line 16: Write  docstring regarding this class.

quic-amitraj (changes requested):

tests/transformers/models/test_image_text_to_text_models.py
quic-dhirajku: Will add an extra commented item with the required fields.

tests/transformers/models/test_image_text_to_text_models.py
quic-dhirajku: Won't run as part of CI, but useful for individual model testing during onboarding so keeping it commented here.

QEfficient/utils/test_utils.py
quic-dhirajku: Will add one for the class.

abukhoy (commented):
I think that `run_utils.py` and `test_utils.py` are used in testing only, So wouldn't be good if we move these two modules inside tests directory or another some directory under parent dir itself just like `transformers` repo?
What do you think @quic-rishinr @quic-amitraj @ochougul ?

tests/transformers/models/test_image_text_to_text_models.py
```diff
@@ -0,0 +1,361 @@
...
+    num_devices: int = 1,
+):
+    model_config = {"model_name": model_name}
+    model_config["img_size"] = img_size
+    config = AutoConfig.from_pretrained(
+        model_config["model_name"], token=HF_TOKEN, trust_remote_code=True, padding=True
+    )
+    config = set_num_layers(config, n_layer=n_layer)
+    model_hf, _ = load_image_text_to_text_model(config)
+    processor = AutoProcessor.from_pretrained(model_name, token=HF_TOKEN, trust_remote_code=True, padding=True)
```
abukhoy, line 168: Please remove `token` param from here, otherwise by default it is trying to login in HF but the `HF_TOKEN` is empty. Then Only CI would pass, I have already downloaded the processor config in the cache.

PR #339, opened 2025-04-02
Vlm test

quic-dhirajku:
Adding vlm test script changes to the original PR


QEfficient/transformers/models/mllama/modeling_mllama.py
```diff
@@ -429,6 +432,7 @@ def forward(
             hidden_states=hidden_states,
             attention_mask=cross_attention_mask,
             cross_attention_states=cross_attention_states,
+            # position_ids=position_ids,
```
quic-akuruvil, line 435: Please remove all commented lines of code

tests/transformers/models/test_image_text_to_text_models.py
```diff
@@ -55,37 +60,52 @@
...
+            config=model_config,
         )
     params = sum(p.numel() for p in model_hf.parameters())
     model_hf.eval()
     return model_hf, params
 
 
-def set_single_num_layers(config, n_layer=1):
-    if hasattr(config, "architectures") and "LlavaForConditionalGeneration" in config.architectures:
+def set_num_layers(config, n_layer=1):
```
quic-akuruvil, line 94: Is this tested and verified, if not please do, whether single layer runs are matching for AIC and pytorch.

tests/transformers/models/test_image_text_to_text_models.py
```diff
@@ -125,100 +151,48 @@ def check_image_text_to_text_pytorch_vs_kv_vs_ort_vs_ai100(
...
-
-    generated_texts = processor.tokenizer.batch_decode(generated_ids, skip_special_tokens=True)
-    print(generated_texts)
+    inputs = processor(images=image, text=prompt, return_tensors="pt")
+    print("QPC Outputs (QAIC):")
+    output = qeff_model.generate(inputs=inputs, generation_len=NEW_GENERATION_TOKENS, streamer=streamer)
+    qpc_tokens = output.generated_ids[:, :-1]
+    # assert (pytorch_hf_tokens == qpc_tokens).all(), "Tokens don't match for pytorch HF output and QPC output"
+    checks.append(pytorch_hf_tokens == qpc_tokens)
+    print(checks)
```
quic-akuruvil, line 195: Are all three outputs in 3 different platforms verified now? Is it tested for single or full model?

QEfficient/utils/_utils.py
```diff
@@ -373,7 +373,7 @@ def get_num_layers_vlm(config):
     return (n_layers_text, n_layers_vision)
 
 
-def get_padding_shape_vlm(config, ctx_len, batch_size=1):
+def get_padding_shape_vlm(config, prompt_len, batch_size=1):
```
quic-akuruvil, line 376: For KV shape/padding shape keep `ctx_len` instead of `prompt_len`. It is correct logical naming.

quic-akuruvil (commented):
Please paste the screenshot showing all tests passed too in PR. For others easy review too.

tests/transformers/models/test_image_text_to_text_models.py
quic-dhirajku: Yes, tested and verified for the 3 models. If we take only single layer for Mllama, cross_attention won't come into picture, so we'll need to provide what's the minimum number of layers we want for each model to be tested properly. Also, simply pass -1 to not take any subset model.

QEfficient/utils/_utils.py
quic-dhirajku: Sure, will do

tests/transformers/models/test_image_text_to_text_models.py
quic-dhirajku: All tests done on single layer model; they match in all the cases except mllama qeff kv model. InternWrapper model isn't working as of now, will need changes in it to get those values.

Full model testing won't require changes in the modeling files and only the input parameter of `n_layer` to be changed so not explicitly testing it yet.

quic-akuruvil: Okay if we have single layer full match, I think we are good now.

PR #352, opened 2025-04-09
Bug Fix: update qeff wheel package name

quic-shagun:
The wheel file name is qefficient and Not QEfficient
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ pip install git+https://github.com/quic/efficient-transformers
 # Or build wheel package using the below command.
 pip install build wheel
 python -m build --wheel --outdir dist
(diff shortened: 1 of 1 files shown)


ochougul:
We should change the name of the wheel file to match ReadMe actually.
I guess `pyproject.toml` has it.

quic-shagun:
> We should change the name of the wheel file to match ReadMe actually. I guess `pyproject.toml` has it.

The name in pyproject.toml is QEfficient but the python package name are case-insensitive and are normalized to lowercase during creation to avoid any discrepancy 
Ref: https://packaging.python.org/en/latest/specifications/binary-distribution-format/

quic-rishinr (approved):

ochougul (approved):

PR #363, opened 2025-04-15
Add compile option is_prefill_only

quic-xiyushi:
Adding support for prefill-only compilation in QEfficient to enable pipeline prefill.


ochougul:
Addressed in #365

PR #379, opened 2025-04-23
[Bug Fix]: qpc sdk config dump issue fixing

abukhoy:
This PR addresses an issue causing the qconfig dump to fail when QAIC is not installed but QNN is present. With this update, the qconfig will be successfully dumped in both QAIC and QNN environments.


QEfficient/utils/_utils.py
```diff
@@ -538,25 +525,49 @@ def make_serializable(obj):
...
     if enable_qnn:
+        # Extract QNN SDK details from YAML file if the environment variable is set
+        qnn_sdk_path = os.getenv(QnnConstants.QNN_SDK_PATH_ENV_VAR_NAME)
+        if not qnn_sdk_path:
+            raise EnvironmentError(
+                f"QNN_SDK_PATH {qnn_sdk_path} is not set. Please set {QnnConstants.QNN_SDK_PATH_ENV_VAR_NAME}"
+            )
+        qnn_sdk_yaml_path = os.path.join(qnn_sdk_path, QnnConstants.QNN_SDK_YAML)
+        with open(qnn_sdk_yaml_path, "r") as file:
+            qnn_sdk_details = yaml.safe_load(file)
```
asmigosw, line 537: `qnn_sdk_details` is not declared earlier, what if `qnn_sdk_yaml_path` is None, then this variable will not be initialized and will give error at line 544. Please declare `qnn_sdk_details = None` before file open.

QEfficient/utils/_utils.py
```diff
@@ -538,25 +525,49 @@ def make_serializable(obj):
...
+        # Extract QAIC SDK Apps or Platform Version from SDK XML file
+        sdk_xml_file = (
+            Constants.SDK_APPS_XML
+            if os.path.exists(Constants.SDK_APPS_XML)
+            else Constants.SDK_PLATFORM_XML
+            if os.path.exists(Constants.SDK_PLATFORM_XML)
+            else None
+        )
+        qaic_version = None
+        if sdk_xml_file is not None:
```
asmigosw, line 557: What if `sdk_xml_file` is None?

asmigosw, line 557: What if `sdk_xml_file` is None?

QEfficient/utils/_utils.py
```diff
@@ -492,7 +492,7 @@ def create_and_dump_qconfigs(
     many other compilation options.
     """
     qnn_config = compiler_options["qnn_config"] if "qnn_config" in compiler_options else None
```
quic-rishinr, line 494: you can replace it as compiler_options.get("qnn_config", None)

QEfficient/utils/_utils.py
```diff
@@ -538,25 +525,49 @@ def make_serializable(obj):
...
+            "enable_qnn": enable_qnn,
+            "qnn_config_path": qnn_config_path,
+        }
         qconfigs["qpc_config"]["qnn_config"] = qnn_config
         if qnn_sdk_details:
             qconfigs["qpc_config"]["qnn_config"].update(qnn_sdk_details)
+
     else:
+        # Extract QAIC SDK Apps or Platform Version from SDK XML file
+        sdk_xml_file = (
```
quic-rishinr, line 549: 1. create a method for fetching the apps and platform sdk. 
2. Its better to capture platform sdk as well if its installed. it will be useful while debugging

QEfficient/utils/_utils.py
```diff
@@ -538,25 +525,49 @@ def make_serializable(obj):
...
+            if os.path.exists(Constants.SDK_PLATFORM_XML)
+            else None
+        )
+        qaic_version = None
+        if sdk_xml_file is not None:
+            tree = ET.parse(sdk_xml_file)
+            root = tree.getroot()
+            qaic_version = root.find(".//base_version").text
+
+        aic_compiler_config = {
```
quic-rishinr, line 562: These details will not be captured while qnn is enabled. Which is not correct. its better to keep sdk check and others outside as we had done it earlier. if SDK are installed, we will capture that

quic-rishinr (changes requested):

QEfficient/utils/_utils.py
```diff
@@ -492,7 +492,7 @@ def create_and_dump_qconfigs(
     many other compilation options.
     """
     qnn_config = compiler_options["qnn_config"] if "qnn_config" in compiler_options else None
-    enable_qnn = True if "qnn_config" in compiler_options else None
+    enable_qnn = True if "qnn_config" in compiler_options else False
```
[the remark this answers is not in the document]
abukhoy: But this` qconfig_dump` function is being called from `QEFFBaseModel`'s `_compile` method in which we have `qnn_config` and  `enable_qnn` params. Here, we need for dumping the specific configs whether it is aic or qnn.

QEfficient/utils/_utils.py
abukhoy: Then `aic_version` would be None.

QEfficient/utils/_utils.py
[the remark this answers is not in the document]
quic-rishinr: Agree. But QNN is enabled by passing enable_qnn flag. Qnn_config is an optional param. if its passed we need to add it to the configs. But in this case if the qnn_config is None and enable_qnn is true we would not be creating the qnn config.

QEfficient/utils/_utils.py
abukhoy: `aic_compiler_config` only for qaic not for qnn, earlier also it was same. If I need to change something, please let me know.

asmigosw:
LGTM

QEfficient/utils/_utils.py
```diff
@@ -495,6 +495,23 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
+def get_qaic_sdk_version():
+    # Extract QAIC SDK Apps or Platform Version from SDK XML file
+    sdk_xml_file = (
```
quic-rishinr, line 500: platform and apps sdk can be different. This method should return both

QEfficient/utils/_utils.py
```diff
@@ -495,6 +495,23 @@ def wrapper(self, *args, **kwargs):
...
+        if os.path.exists(Constants.SDK_APPS_XML)
+        else Constants.SDK_PLATFORM_XML
+        if os.path.exists(Constants.SDK_PLATFORM_XML)
+        else None
+    )
+    qaic_version = None
+    if sdk_xml_file is not None:
+        tree = ET.parse(sdk_xml_file)
+        root = tree.getroot()
+        qaic_version = root.find(".//base_version").text
```
quic-rishinr, line 511: if its not able to find ".//base_version" it would cause an attribute error while trying to access .text. check if root.find(".//base_version") is not none and then call .text

QEfficient/utils/_utils.py
```diff
@@ -495,6 +495,23 @@ def wrapper(self, *args, **kwargs):
...
+    sdk_xml_file = (
+        Constants.SDK_APPS_XML
+        if os.path.exists(Constants.SDK_APPS_XML)
+        else Constants.SDK_PLATFORM_XML
+        if os.path.exists(Constants.SDK_PLATFORM_XML)
+        else None
+    )
+    qaic_version = None
+    if sdk_xml_file is not None:
+        tree = ET.parse(sdk_xml_file)
```
quic-rishinr, line 509: better to have an exception hander here as well.

QEfficient/utils/_utils.py
```diff
@@ -511,29 +528,13 @@ def create_and_dump_qconfigs(
...
-    qaic_version = root.find(".//base_version").text
-
-    # Extract QNN SDK details from YAML file if the environment variable is set
-    qnn_sdk_details = None
-    qnn_sdk_path = os.getenv(QnnConstants.QNN_SDK_PATH_ENV_VAR_NAME)
-    if enable_qnn and qnn_sdk_path:
-        qnn_sdk_yaml_path = os.path.join(qnn_sdk_path, QnnConstants.QNN_SDK_YAML)
-        with open(qnn_sdk_yaml_path, "r") as file:
-            qnn_sdk_details = yaml.safe_load(file)
+    qnn_config_path = qnn_config if qnn_config is not None else "QEfficient/compile/qnn_config.json"
```
quic-rishinr, line 537: better to fetch the path from constant file than hardcoding here

quic-rishinr (changes requested):

QEfficient/utils/_utils.py
abukhoy: But I am seeing that both are same! 
Okay, if they are different then let's dump both versions into the qconfig. What do you think?

asmigosw: We keep them same generally, but they can be different as well so it's better to track both.

QEfficient/utils/_utils.py
```diff
@@ -558,29 +585,13 @@ def create_and_dump_qconfigs(
...
-    qaic_version = root.find(".//base_version").text
-
-    # Extract QNN SDK details from YAML file if the environment variable is set
-    qnn_sdk_details = None
-    qnn_sdk_path = os.getenv(QnnConstants.QNN_SDK_PATH_ENV_VAR_NAME)
-    if enable_qnn and qnn_sdk_path:
-        qnn_sdk_yaml_path = os.path.join(qnn_sdk_path, QnnConstants.QNN_SDK_YAML)
-        with open(qnn_sdk_yaml_path, "r") as file:
-            qnn_sdk_details = yaml.safe_load(file)
+    qnn_config_path = qnn_config if qnn_config is not None else QnnConstants.QNN_DEFAULT_CONFIG_PATH
```
shubhagr-quic, line 594: If qnn_config is passed as None, we should be mentioning qnn_config_path as None only, since QNN_DEFAULT_CONFIG_PATH will not be used here. Also, in case where we are using qefficient as a package and dont have access to the qefficient repository, qnn_config will not be present in the default path. 

@quic-rishinr Please validate the observation.

shubhagr-quic:
Thanks @abukhoy for incorporating the comment. LGTM

quic-hemagnih (approved):

quic-rishinr (approved):

asmigosw (approved):

PR #388, opened 2025-05-05

quic-vjanfaza:


QEfficient/generation/text_generation_inference.py
```diff
@@ -741,6 +751,19 @@ def run_prefill(self, prompt, generation_len, prefill_logit_bs=1, decode_batch_i
             generation_len,
         )
 
+    def initialize_ccl(self, decode_inputs):
+        max_ccl_id = len(self.comp_ctx_lengths) - 1
+        max_position_id = np.max(decode_inputs["position_ids"])
+        ccl_id = 1
+        for i in range(1, len(self.comp_ctx_lengths)):
```
quic-rishinr, line 758: can we have a reverse list and pop out the last value if max_position_id < self.comp_ctx_lengths[-1]?  this way we can avoid the loop

QEfficient/generation/text_generation_inference.py
```diff
@@ -808,6 +841,19 @@ def run_continuous_batching_decode(self, prompt_queue, generation_len):
...
+                        if self.comp_ctx_lengths is not None:
+                            ###Recalculate ccl_id based on position ids###
+                            # Determine the maximum value of position_ids across all batch elements
+                            max_position_id = np.max(decode_inputs["position_ids"])
+
+                            # Update ccl_id and comp_ctx_lengths based on the maximum position id
+                            ccl_id = 1
+                            for i in range(1, len(self.comp_ctx_lengths)):
+                                if max_position_id < self.comp_ctx_lengths[i]:
+                                    ccl_id = i
```
quic-rishinr, line 853: same as above

quic-rishinr (changes requested):

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1388,6 +1389,9 @@ def from_pretrained(
 
         kv_offload = kwargs.pop("kv_offload", None)
 
+        comp_ctx_lengths = kwargs.pop("comp_ctx_lengths", None)
+        cls.comp_ctx_lengths = comp_ctx_lengths
```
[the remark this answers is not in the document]
quic-vjanfaza: We can't remove these lines since we call cls._hf_auto_class.from_pretrained(pretrained_model_name_or_path, *args, **kwargs) in line 1396 which is looking for comp_ctx_lengths in the input arguments and causes error.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1442,10 +1446,12 @@ def export(self, export_dir: Optional[str] = None) -> str:
             "input_ids": torch.zeros((bs, seq_len), dtype=torch.int64),
             "position_ids": torch.arange(seq_len, dtype=torch.int64).view(1, seq_len).repeat(bs, 1),
             "past_key_values": [[] for _ in range(self.num_layers)],
+            "comp_ctx_lengths": torch.randint(0, 100, (40,), dtype=torch.long),
```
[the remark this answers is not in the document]
quic-vjanfaza: No, we don't need to do that. Since, I generalized the code to use the same onnx file for both with CCL and the default without CCL experiments. This will reduce the number of changes in model file as well and will prevent having multiple onnx generation for with and without CCL experiments.

QEfficient/generation/text_generation_inference.py
quic-vjanfaza: Why should we check with the last element? Each request can be finished in different position_id and we need to check to find the most suitable CCL window to get the best performance. This for loop only happens at the end of a request and it's an order of length(CCL) that can't be more than a few values because of compiler limitation in the number of specializations.

quic-rishinr:
The updated change on this feature is added on #401

quic-rishinr:
Closing this PR as the changes are added in #401

PR #393, opened 2025-05-08
Adding compilation support for io_encrypt flag

asmigosw:
As part of Model-IP I/O encryption feature qaic-exec has added a new command line option: `-io-encrypt` to enable support for decryption/encryption of network I/O.
Added support in Qeff to pass this param to qaic-exec when a model needs to be compiled with I/O encryption support.

CLI command: `python -m QEfficient.cloud.infer --model_name gpt2 --batch_size 1 --prompt_len 32 --ctx_len 128 --mxfp6 --num_cores 16 --device_group [0] --prompt "My name is" --mos 1 --aic_enable_depth_first --io-encrypt "chacha20"`

Note: If the flag is passed via infer CLI, CLI will exit after QPC generation.

### DIFF
--- a/QEfficient/cloud/compile.py
+++ b/QEfficient/cloud/compile.py
@@ -85,17 +85,29 @@ if __name__ == "__main__":
     parser.add_argument(
         "--enable_qnn",
         "--enable-qnn",
-        action="store_true",
+        nargs="?",
+        const=True,
+        type=str,
         default=False,
         help="Enables QNN. Optionally, a configuration file can be provided with [--enable_qnn CONFIG_FILE].\
              If not provided, the default configuration will be used.\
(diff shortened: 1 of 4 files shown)


abukhoy (commented):
I think that we should add these compilation params at somewhere either in the help section or in the docs for future reference. Otherwise, we may not know how many params we have incorporated till now. Just a thought, what do you think!?

asmigosw:
> I think that we should add these compilation params at somewhere either in the help section or in the docs for future reference. Otherwise, we may not know how many params we have incorporated till now. Just a thought, what do you think!?

You can get info on all these extra params using: `/opt/qti-aic/exec/qaic-exec -help`

PR #412, opened 2025-05-20
Fix for Replicate_kv_heads script

quic-hemagnih:
As we have moved to the transformer version 4.50.0, we have changed the attention implementation and hence number of attention heads and hidden size are no more attributes of this class.

We have added these parameters are optional arguments for the replicate_kv_heads script. User can now pass them as arguments, if not passed explicitly then it shall be picked from config.json file of the model.

### DIFF
--- a/scripts/replicate_kv_head/replicate_kv_heads.py
+++ b/scripts/replicate_kv_head/replicate_kv_heads.py
@@ -103,13 +103,24 @@ def main(args):
     print("Original KV heads:", orig_kv_heads)
     print("Modified KV heads:", new_kv_heads)
 
+    # Check if hidden size and number of attention heads are explicitly passed as arguments or not
+    if args.num_attention_heads is None:
+        num_attn_heads = model.config.num_attention_heads
+    else:
+        num_attn_heads = args.num_attention_heads
+
+    if args.hidden_size is None:
+        hidden_size = model.config.hidden_size
+    else:
+        hidden_size = args.hidden_size
+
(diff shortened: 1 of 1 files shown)


scripts/replicate_kv_head/replicate_kv_heads.py
```diff
@@ -103,13 +103,24 @@ def main(args):
...
+    # Check if hidden size and number of attention heads are explicitly passed as arguments or not
+    if args.num_attention_heads is None:
+        num_attn_heads = model.config.num_attention_heads
+    else:
+        num_attn_heads = args.num_attention_heads
+
+    if args.hidden_size is None:
+        hidden_size = model.config.hidden_size
+    else:
+        hidden_size = args.hidden_size
```
ochougul, line 115: Should we raise warning if there is a mismatch between the value that user has sent and the value that exists in config if it exists in config as well as user send it ?

ochougul (approved):
LGTM

asmigosw (approved):
LGTM

scripts/replicate_kv_head/replicate_kv_heads.py
quic-hemagnih: By this change we are adding flexibility for user to pass these two arguments, so don't want to put another check on it. So that we use this functionality for positive as well as negative testing. Along with this we can use this in future for any experimental work too.

quic-rishinr (approved):

PR #413, opened 2025-05-20
Added Model mapping for swiftKV model to __init__.py

asmigosw:
`__init__.py` was calling modelling_utils leading to creation of `MODEL_CLASS_MAPPING` before init. As a result, registered swiftKV model's config class and QEff Class was not present in the `MODEL_CLASS_MAPPING` and hence giving error of unknown architecture.

Updated this `__init__` to be initialized first and then all other classes by removing the modelling_utils call and adding model_class_mapping in __init__.

### DIFF
--- a/QEfficient/__init__.py
+++ b/QEfficient/__init__.py
@@ -12,10 +12,19 @@ import os
 # hf_transfer is imported (will happen on line 15 via leading imports)
 os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
 
-from transformers import AutoConfig
+from transformers import AutoConfig, AutoModelForCausalLM
 
-from QEfficient.transformers.modeling_utils import MODEL_TYPE_TO_CONFIG_CLS_AND_ARCH_CLS
+# Placeholder for all non-transformer models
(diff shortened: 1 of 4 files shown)


QEfficient/__init__.py
```diff
@@ -12,11 +12,20 @@
 # hf_transfer is imported (will happen on line 15 via leading imports)
 os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
 
-from transformers import AutoConfig
+from transformers import AutoConfig, AutoModelForCausalLM
```
quic-hemagnih, line 16: Should we move below code from __init__.py to some other file, may be we can create a file (like non-transformer based config) and then move the logic there. 

The new file may have one module exposed which we can call from this file.

Why I am suggesting this is because generally __init__.py is kept very simple and small from modularity perspective. I think in future if any asks similar to swiftkv comes then it will keep on increasing the size of __init__.py.

Thoughts??

asmigosw: Yes, I believe we can create a new register file and move our complete register code there and initialize that file in `__init__.py` file. It will be easier for user as well to read it and for code clarity.

quic-amitraj (approved):
LGTM

PR #414, opened 2025-05-20
[release/v1.20.0] Added Model mapping for swiftKV model to __init__.py 

asmigosw:
`__init__.py` was calling modelling_utils leading to creation of `MODEL_CLASS_MAPPING` before init. As a result, registered swiftKV model's config class and QEff Class was not present in the `MODEL_CLASS_MAPPING` and hence giving error of unknown architecture.

Updated this `__init__` to be initialized first and then all other classes by removing the `modelling_utils` call and adding model_class_mapping in init.


asmigosw:
Updated New PR #415

PR #418, opened 2025-05-21
Fix for replicate kv heads script

quic-hemagnih:
As we have moved to the transformer version 4.50.0, we have changed the attention implementation and hence number of attention heads and hidden size are no more attributes of this class.

We have added these parameters are optional arguments for the replicate_kv_heads script. User can now pass them as arguments, if not passed explicitly then it shall be picked from config.json file of the model.


quic-hemagnih:
Duplicate PR, we need to raise for release branch 1.20.0

PR #426, opened 2025-05-29
Updated Python APIs Compile doc string to clearly reflect QNN Compilation path

shubhagr-quic:


abukhoy:
> @abukhoy please update the GH pages with this change

Will update it once #423 is merged.

PR #314, opened 2025-03-13
Added Infra in QEfficient for execution of swiftkv models

quic-hemagnih:
Added Infra in QEfficient for execution of models whose modelling file is not present at hugging face and checkpoint like swiftkv

PR #427, opened 2025-05-29
[Tests]: Adding dummy causal models for testing in regular CI run

abukhoy:
### Purpose of this PR:
This update aims to reduce test execution time for causal language model inference. Previously, tests were run using full-scale models with one or two layers, which was inefficient and time-consuming. Refactoring CLI api testing for independent testing and redundant conftest code.

### What’s Changed:
Introduced dummy models with significantly smaller configurations by adjusting parameters such as `max_position_embeddings, num_hidden_layers, num_attention_heads, hidden_size, intermediate_size, vocab_size and additional_params`.
These lightweight models are used exclusively for testing purposes to ensure faster execution without compromising test coverage.

And CLI testing has two test scripts one is for export, compile, and execute, another is for infer cli api.

**Note:** This optimization is applied only to causal language models.


tests/transformers/models/test_causal_lm_models.py
```diff
@@ -252,6 +414,37 @@ def test_causal_lm_export_with_deprecated_api(model_name):
...
+@pytest.mark.parametrize(
+    "test_dummy_model_config, test_dummy_model_name", test_dummy_model_configs, ids=test_dummy_model_names
+)
+def test_dummy_causal_lm_pytorch_vs_kv_vs_ort_vs_ai100(test_dummy_model_config, test_dummy_model_name):
+    """
+    Test function to validate the PyTorch model, the PyTorch model after KV changes, the ONNX model, and the Cloud AI 100 model, both with and without continuous batching.
+    ``Mandatory`` Args:
+        :model_name (str): Hugging Face Model Card name, Example: ``gpt2``
+    """
+    if test_dummy_model_name in {
```
quic-rishinr, line 428: We should avoid putting such constants. May be have a separate test for quantized models

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -292,6 +515,35 @@ def test_causal_lm_pytorch_vs_kv_vs_ort_vs_ai100_qnn(model_name):
 
 
 @pytest.mark.skip()  # remove when the SDK 1.20.0 issue solved for compiling this model
```
quic-rishinr, line 517: Can we remove it now? same on line 545 as well

tests/cloud/test_infer.py
```diff
@@ -39,9 +67,84 @@ def test_infer(setup, mocker):
...
         prompt_len=ms.prompt_len,
         ctx_len=ms.ctx_len,
-        generation_len=ms.generation_len,
+        generation_len=generation_len,
         mxfp6=ms.mxfp6,
         mxint8=ms.mxint8,
         full_batch_size=ms.full_batch_size,
         enable_qnn=ms.enable_qnn,
+        image_url=kwargs["image_url"],
+    )
```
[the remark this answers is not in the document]
abukhoy: Added checks and assertions which were possible.

tests/cloud/test_infer.py
```diff
@@ -39,9 +67,84 @@ def test_infer(setup, mocker):
...
+    local_config.update(
+        {
+            "model_name": "llava-hf/llava-1.5-7b-hf",
+            "prompt": "Describe the image.",
+            "full_batch_size": None,
+            "enable_qnn": False,
+            "qnn_config": None,
+        }
     )
+    check_infer(mocker=mocker, generation_len=20, **local_config)
```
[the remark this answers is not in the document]
abukhoy: I tried adding this test but got error, so removed from the test.

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -45,21 +47,182 @@
...
     "neuralmagic/Qwen2-0.5B-Instruct-FP8",  # fp8 quant method, static, with lm head ignored
     "ibm-granite/granite-3.1-2b-instruct",
-    "ibm-granite/granite-guardian-3.1-2b",
+    # "ibm-granite/granite-guardian-3.1-2b",
     "hpcai-tech/grok-1",
 ]
 
+test_dummy_model_configs = [
+    # model_name, model_type, max_position_embeddings, num_hidden_layers, num_attention_heads, hidden_size, intermediate_size, vocab_size, additional_params
+    ("TinyLlama/TinyLlama-1.1B-Chat-v1.0", "llama", 128, 1, 2, 64, 256, 32000, {"num_key_value_heads": 1}),
```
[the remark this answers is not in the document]
abukhoy: No criteria, but some of the models has constraint for some of the config params with specific value.

tests/cloud/test_export_compile_execute.py
```diff
@@ -0,0 +1,115 @@
...
+
+    assert os.path.isfile(onnx_model_path)
+
+    base_key = "past_key."
+    base_value = "past_value."
+    precision = "float16"
+
+    data = []
+
+    for i in range(12):
```
quic-hemagnih, line 44: Please avoid using magic numbers, we should define constants for 12

tests/cloud/test_export_compile_execute.py
```diff
@@ -0,0 +1,167 @@
...
+        aic_enable_depth_first=model_setup.aic_enable_depth_first,
+        mos=model_setup.mos,
+        batch_size=model_setup.batch_size,
+        prompt_len=model_setup.prompt_len,
+        ctx_len=model_setup.ctx_len,
+        mxfp6=model_setup.mxfp6,
+        mxint8=model_setup.mxint8,
+        full_batch_size=model_setup.full_batch_size,
+        enable_qnn=model_setup.enable_qnn,
+    )
```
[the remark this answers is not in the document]
quic-hemagnih: Yes we should add a check for custom IO file too

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -5,13 +5,15 @@
 #
 # -----------------------------------------------------------------------------
 
+import copy
+import json
 import os
```
quic-hemagnih, line 10: Can you add changes of the file https://github.com/quic/efficient-transformers/pull/314/files#diff-fcc05ce2e83110153682cf6f488ad15825018072c3efd5e48266119cebe4e77a in this PR. I think some how we missed the swiftkv Unit test from our repo. I think since you are making changes in this file, we can add these changes too,

tests/cloud/test_export_compile_execute.py
abukhoy: This is model specific, so we cannot generalize now. This `12` is number of layers of the model like `gpt2` model in this case. Actually, we are creating a custom IO file here and later it will be no longer needed when the compile CLI Api would be backed by HL Api.

tests/cloud/test_export_compile_execute.py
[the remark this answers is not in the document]
abukhoy: I don't think it is needed as we are creating custom IO file here itself before compile called.

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -47,42 +48,120 @@
...
+        if config["model_type"] is None:
+            config_dict[config["model_name"]] = AutoConfig.from_pretrained(
+                config["model_name"],
+                trust_remote_code=config["model_name"] in extrenal_models,
+                **config["additional_params"],
+            )
+        else:
+            config_dict[config["model_name"]] = AutoConfig.for_model(
+                config["model_type"],
+                max_position_embeddings=config["max_position_embeddings"],
```
quic-rishinr, line 100: Can we use a get function instead of config[<param>]? in case some params are not available it will throw a key error

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -22,7 +24,6 @@
 from QEfficient.utils.device_utils import get_available_device_id
 from QEfficient.utils.run_utils import ApiRunner
 
-extrenal_models = {"hpcai-tech/grok-1"}
 test_models_qaic = [
```
quic-rishinr, line 27: Can we make this as a set so the in operations will be O(1)

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -47,42 +48,120 @@
...
+
+    Returns:
+        Tuple[List[Tuple[AutoConfig, str]], List[str]]: A tuple containing:
+            - A list of (AutoConfig, model_name) tuples.
+            - A list of model names.
+    """
+    config_objects = []
+    model_names = []
+    for name, config in model_config_dict.items():
+        if name in selected_model_names:
```
quic-rishinr, line 128: please make sure the selected_model_names is a set and not list or tuple

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -47,42 +48,120 @@
...
+        model_hf = AutoModelForCausalLM.from_pretrained(
+            model_path,
+            use_cache=True,
+            num_hidden_layers=n_layer,
+            attn_implementation="eager",
+            low_cpu_mem_usage=False,
+            trust_remote_code=model_name in extrenal_models,
+        )
+    else:
+        torch.manual_seed(42)
```
quic-rishinr, line 159: Can you define this at the top?

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -113,9 +277,9 @@ def check_causal_lm_pytorch_vs_kv_vs_ort_vs_ai100(
     replace_transformers_quantizers()
     model_config = {"model_name": model_name}
     model_config["n_layer"] = n_layer
-
-    model_hf, _ = load_causal_lm_model(model_config)
-
+    if model_hf is None:
+        model_hf, _ = load_causal_lm_model(model_config)
+    model_hf_cb = copy.deepcopy(model_hf)
```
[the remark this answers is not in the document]
quic-rishinr: why we need to have a copy of model_hf for CB?

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -213,21 +295,46 @@ def check_causal_lm_pytorch_vs_kv_vs_ort_vs_ai100(
...
+                all(pt_token[:24] == cloud_token[:24])
+                for pt_token, cloud_token in zip(pytorch_hf_tokens, exec_info_fbs.generated_ids)
+            ]
+        ), "Tokens don't match for  HF PyTorch model output and Cloud AI 100 output."
+
     assert os.path.isfile(os.path.join(os.path.dirname(qpc_path), "qconfig.json"))
 
 
+# Load the custom models configuration data from the JSON file
+with open("tests/transformers/models/custom_tiny_model_configs.json", "r") as f:
```
quic-rishinr, line 317: is there a better way to handle this? We can use a class or using a global variable to load the configs and keep it. Read the configs and model names from each test methods

quic-rishinr (changes requested):

tests/transformers/models/test_causal_lm_models.py
abukhoy: why we need to have a copy of model_hf for CB?
=> If we don't copy then we will have to load it again for CB testing as we cannot use the same model for CB and Non-CB testing.

tests/cloud/test_export_compile_execute.py
abukhoy: No, we cannot fetch the model configs from test function as test function can pass the args to the main code for execution. As I already told that it looks not good just because it is not supported by HL API as export does.
Later it won't be needed for creating custom IO file when this CLI compile is backed by HL compile method.

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -253,61 +277,109 @@ def test_causal_lm_export_with_deprecated_api(model_name):
...
+@pytest.mark.nightly
+@pytest.mark.on_qaic
+@pytest.mark.parametrize("model_name", test_models_qaic)
+def test_causal_lm_pytorch_vs_kv_vs_ort_vs_ai100(model_name):
+    """
+    Test function to validate the PyTorch model, the PyTorch model after KV changes, the ONNX model, and the Cloud AI 100 model, both with and without continuous batching.
+    ``Mandatory`` Args:
+        :model_name (str): Hugging Face Model Card name, Example: ``gpt2``
+    """
+    n_layer = (
```
quic-rishinr, line 304: nit: It would better if we can have a method for it right? its not very readable

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -21,8 +22,8 @@
 from QEfficient.utils.constants import Constants, QnnConstants
 from QEfficient.utils.device_utils import get_available_device_id
 from QEfficient.utils.run_utils import ApiRunner
+from QEfficient.utils.test_utils import ModelConfig
 
-extrenal_models = {"hpcai-tech/grok-1"}
 test_models_qaic = [
     "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
```
quic-rishinr, line 28: nit Better to maintain a common naming standard for the model list. like causal_models as you have defined below.

quic-rishinr (changes requested):

tests/transformers/models/test_causal_lm_models.py
```diff
@@ -64,6 +64,22 @@
...
+    Function to set number layers of the variuos types of models such as swiftkv models and others
+    --------
+
+    :model_name: str
+
+    :return n_layer
+    """
+    if model_name in {"microsoft/Phi-3-mini-4k-instruct", "neuralmagic/Qwen2-0.5B-Instruct-FP8"}:
+        return 2
+    elif model_name in ModelConfig.SWIFTKV_MODELS:
```
quic-rishinr, line 78: Can we remove the hardcoded values? in case of swift KV we should not be passing the num_hidden_layers

quic-rishinr (changes requested):

tests/transformers/models/test_causal_lm_models.py
abukhoy: I have removed it and now it will return `None` for swiftkv models. That means it will load the model via `load_causal_model` method.

PR #436, opened 2025-06-10
Name change fix in QNN Config Data format file

shubhagr-quic:
Updated generate_data_format_config API to disable original (framework) name sanitization on AIC backend for QNN Compilation path.


shubhagr-quic:
Merged in mainline, will be picked up to 1.20 after branch cut.

PR #463, opened 2025-06-18
Unit Tests for On Device Sampling

quic-sanising:
- `test_sampler_transform`
    Currently, QEfficient is responsible for export and compiling a model with and without Sampler nodes attached. This unit test confirms the presence of the requisite sampler nodes in the compiled QPC.

### DIFF
--- a/tests/transformers/sampler/test_sampler.py
+++ b/tests/transformers/sampler/test_sampler.py
@@ -1,0 +1,119 @@
+# -----------------------------------------------------------------------------
+#
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# -----------------------------------------------------------------------------
+
+from typing import List
+
+import pytest
+
+from QEfficient import QEFFAutoModelForCausalLM as AutoModelForCausalLM
+from QEfficient.generation.cloud_infer import QAICInferenceSession
+from QEfficient.utils.constants import Constants
+
+configs = [
+    pytest.param(
+        "meta-llama/Llama-3.1-8B",  # model
+        Constants.INPUT_STR,  # prompts
+        32,  # prefill_seq_len
+        256,  # ctx_len
+        4,  # full_batch_size
+        1,  # num_devices
+        16,  # num_cores
+        1,  # spec_length
+        id="Llama-3.1-8B_32_256_4_1_16_1",
+    ),
+    pytest.param(
+        "meta-llama/Llama-3.1-8B",  # model
+        Constants.INPUT_STR,  # prompts
+        32,  # prefill_seq_len
+        256,  # ctx_len
+        4,  # full_batch_size
+        4,  # num_devices
+        16,  # num_cores
+        1,  # spec_length
+        id="Llama-3.1-8B_32_256_4_4_16_1",
+    ),
+]
+
+
+@pytest.mark.on_qaic
+@pytest.mark.parametrize(
+    "model, prompts, prefill_seq_len, ctx_len, full_batch_size, num_devices, num_cores, spec_length",
+    configs,
+)
+def test_sampler_transform(
+    model: str,
+    prompts: List[str],
+    prefill_seq_len: int,
+    ctx_len: int,
+    full_batch_size: int,
+    num_devices: int,
+    num_cores: int,
+    spec_length: int,
+):
+    # Export and compile QEfficient models
+    qaic_config = {
+        "include_sampler": True,
+        "return_pdfs": False,
+        "max_top_k_ids": 512,
+    }
+    model_w_sampler = AutoModelForCausalLM.from_pretrained(model, continuous_batching=True, qaic_config=qaic_config)
+    model_wo_sampler = AutoModelForCausalLM.from_pretrained(model, continuous_batching=True, qaic_config=None)
+    model_w_sampler_qpc_path: str = model_w_sampler.compile(
+        prefill_seq_len=prefill_seq_len,
+        ctx_len=ctx_len,
+        full_batch_size=full_batch_size,
+        num_devices=num_devices,
+        num_cores=num_cores,
+        num_speculative_tokens=spec_length - 1,
+        mxint8_kv_cache=True,
+        mxfp6_matmul=True,
+    )
+    model_wo_sampler_qpc_path: str = model_wo_sampler.compile(
+        prefill_seq_len=prefill_seq_len,
+        ctx_len=ctx_len,
+        full_batch_size=full_batch_size,
+        num_devices=num_devices,
+        num_cores=num_cores,
+        num_speculative_tokens=spec_length - 1,
+        mxint8_kv_cache=True,
+        mxfp6_matmul=True,
+    )
+
+    # Init qaic session
+    model_w_sampler_session = QAICInferenceSession(model_w_sampler_qpc_path)
+    model_wo_sampler_session = QAICInferenceSession(model_wo_sampler_qpc_path)
+
+    # Skip inputs/outputs buffers
+    model_w_sampler_session.skip_buffers(set([x for x in model_w_sampler_session.input_names if x.startswith("past_")]))
+    model_w_sampler_session.skip_buffers(
+        set([x for x in model_w_sampler_session.output_names if x.endswith("_RetainedState")])
+    )
+    model_wo_sampler_session.skip_buffers(
+        set([x for x in model_wo_sampler_session.input_names if x.startswith("past_")])
+    )
+    model_wo_sampler_session.skip_buffers(
+        set([x for x in model_wo_sampler_session.output_names if x.endswith("_RetainedState")])
+    )
+
+    # Validate sampler inputs
+    sampler_inputs = [
+        "last_accepted_output_tokens",
+        "repetition_penalties",
+        "presence_penalties",
+        "temperatures",
+        "top_ks",
+        "top_ps",
+        "min_ps",
+        "random_numbers",
+    ]
+    for input_name in sampler_inputs:
+        assert (
+            input_name in model_w_sampler_session.input_names
+        ), f"Sampler input {input_name} not found in QPC compiled with Sampler"
+        assert (
+            input_name not in model_wo_sampler_session.input_names
+        ), f"Sampler input {input_name} found in QPC compiled without Sampler"


tests/transformers/sampler/test_sampler.py
ochougul, line 1: can we also run the two session with the fixed prompt and make sure the outputs don't match with each other. But match with a golden output that we know matches with pytorch execution as well?

ochougul (changes requested):

quic-rishinr:
@quic-sanising can you add a small feature description under /docs/source/quick_start.md supported feature section? also provide the example script link in the description

quic-sanising:
> @quic-sanising can you add a small feature description under /docs/source/quick_start.md supported feature section? also provide the example script link in the description

Done

quic-amitraj (changes requested):
Please fix lint error.

QEfficient/generation/text_generation_inference.py
```diff
@@ -342,6 +345,11 @@ def cloud_ai_100_exec_kv(
         :Write_io_dir (str): Path to write the input and output files. ``Defaults to None``.
         :automation (bool): If true, it prints input, output, and performance stats. ``Defaults to False``.
         :prompt_to_lora_id_mapping (List[int]): Mapping to associate prompts with their respective LoRA adapter.
+        :include_sampler (bool): Enable/Disable sampling of next tokens.
+        :return_pdfs (bool): Return probability distributions along with sampled
```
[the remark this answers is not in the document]
quic-sanising: Done

QEfficient/generation/text_generation_inference.py
```diff
@@ -411,14 +426,59 @@ def __init__(
...
+            if session_input_name in sampler_inputs:
+                count += 1
+                if count == len(sampler_inputs):
+                    break
+        if count == 0:
+            self.include_sampler = False
+        elif count < len(sampler_inputs):
+            raise ValueError(
+                "The provided QPC does not have the required number of inputs to run sampling "
+                f"on the QAIC device (only {count}/{len(sampler_inputs)} inputs provided). Partial "
```
quic-hemagnih, line 465: I think we should do count % sampler_inputs here. If we divide count by len(sampler_inputs) then it would return 0.

QEfficient/generation/text_generation_inference.py
```diff
@@ -411,14 +426,59 @@ def __init__(
...
+            "temperatures",
+            "top_ks",
+            "top_ps",
+            "min_ps",
+            "random_numbers",
+        ]
+        count = 0
+        for session_input_name in self._session.input_names:
+            if session_input_name in sampler_inputs:
+                count += 1
```
quic-hemagnih, line 457: Can there be a case where user provides the same session_input_names multiple times. In that case how we will catch it in this code. 
count variable will keep on incrementing and may satisfy the condition

QEfficient/generation/text_generation_inference.py
```diff
@@ -411,14 +426,59 @@ def __init__(
...
+                if count == len(sampler_inputs):
+                    break
+        if count == 0:
```
quic-hemagnih, line 460: I think we can avoid this if.. else block.
at line 455 by default set self.include_sampler = False.
Then at line 458 before break set it to True.
At line 462 just check for error condition.

QEfficient/generation/text_generation_inference.py
quic-sanising: `self._session.input_names` come from the exported ONNX file. If there are duplicate names, say `abc`, the ONNX will convert them to something like `abc_0`, `abc_1`, so on. So, we would never get the same name multiple times. 

However, if accuracy is the only priority here and performance is not, I could see `set()` but it would add a slight overhead of `O(n)`.

QEfficient/generation/text_generation_inference.py
quic-sanising: This is only a print statement. We are not actually dividing here. So, if `count = 5` and `len(sampler_inputs) = 10`, it would print `(only 5/10 inputs provided)`.

QEfficient/generation/text_generation_inference.py
quic-sanising: In case the user provides `include_sampler` as input, `self.include_sampler` is not set to `False`. That is why, we need the check in line 460. 

We can only avoid the else block in line 468.

quic-sanising:
@quic-amitraj The lint failures were happening because the linter is installing `ruff v0.12.2` whereas the `.pre-commit-config.yaml` file has an older version of `v0.5.2`. 

To fix the errors, we need to either install `ruff v0.5.2` in the linter or update the `.pre-commit-config.yaml` file to version `v0.12.2`.

tests/transformers/sampler/test_sampler.py
quic-sanising: Please refer to the test `test_random_sampling` below.

tests/transformers/sampler/test_sampler.py
```diff
@@ -0,0 +1,344 @@
...
+    pytest.param(
+        "meta-llama/Llama-3.1-8B",  # model
+        Constants.INPUT_STR * 4,  # prompts
+        32,  # prefill_seq_len
+        256,  # ctx_len
+        20,  # generation_len
+        4,  # full_batch_size
+        1,  # spec_length
+    ),
+]
```
ochougul, line 29: can we use single layer model for this?

ochougul (changes requested):
Everything else LGTM, just see if we can use single layer model in our tests

QEfficient/generation/text_generation_inference.py
```diff
@@ -411,14 +426,59 @@ def __init__(
...
+            self.include_sampler = False
+        elif count < len(sampler_inputs):
+            raise ValueError(
+                "The provided QPC does not have the required number of inputs to run sampling "
+                f"on the QAIC device (only {count}/{len(sampler_inputs)} inputs provided). Partial "
+                "sampling support is not available. Please check the QPC and try again."
+            )
+
+        if include_sampler and not self.include_sampler:
```
quic-rishinr, line 470: Can we include this check in line 489 itself?

QEfficient/generation/text_generation_inference.py
```diff
@@ -574,6 +641,21 @@ def prepare_decode_inputs(self):
             decode_inputs["position_ids"] = self.decode_pos_ids
         if self.batch_index is not None:
             decode_inputs["batch_index"] = self.batch_index
+        if self.include_sampler:
+            decode_inputs["last_accepted_output_tokens"] = decode_inputs["input_ids"]
+            for op in [
```
quic-rishinr, line 646: if we can maintain sampler_inputs in constants as mentioned above we should be able to reuse here as well right?

QEfficient/generation/text_generation_inference.py
```diff
@@ -775,10 +894,7 @@ def run_continuous_batching_decode(self, prompt_queue, generation_len):
             outputs = self._session.run(decode_inputs)
 
             # Prepare inputs for next iteration
-            logits = outputs["logits"]
-            if len(logits.shape) == 2:
-                logits = np.expand_dims(logits, 1)
-            next_token_id = logits.argmax(2)
+            next_token_id = self._fetch_next_token_id(outputs)
```
quic-rishinr, line 897: Since this is inside the decode loop would it create any performance drop? Have you done any tests for checking any performance deviation?

tests/transformers/sampler/test_sampler.py
```diff
@@ -0,0 +1,344 @@
...
+        mxint8_kv_cache=True,
+        mxfp6_matmul=True,
+    )
+
+    # Generate texts from prompts
+    tokenizer = load_hf_tokenizer(pretrained_model_name_or_path=model)
+    model_w_sampler_exec_info = model_w_sampler.generate(
+        tokenizer=tokenizer,
+        prompts=prompts,
+        device_id=get_available_device_id(),
```
quic-rishinr, line 167: you need not pass the device ID. you can keep it as None so auto device picker will pick up the available device ID during execution. Please remove device_id=get_available_device_id() from all the methods below.

tests/transformers/sampler/test_sampler.py
```diff
@@ -0,0 +1,344 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# -----------------------------------------------------------------------------
+
+from typing import List
+
+import numpy as np
+import pytest
+
+from QEfficient import QEFFAutoModelForCausalLM as AutoModelForCausalLM
```
quic-rishinr, line 13: Do we need to import QEFFAutoModelForCausalLM as AutoModelForCausalLM? it would create some unnecessary confusions

tests/transformers/sampler/test_sampler.py
```diff
@@ -0,0 +1,344 @@
...
+        prompts=prompts,
+        device_id=get_available_device_id(),
+        generation_len=generation_len,
+        include_sampler=False,
+        return_pdfs=False,
+        sampling_params=None,
+    )
+
+    # Compare generated texts
+    golden_texts = {
```
quic-rishinr, line 278: Can we make this test more robust? The goal is to verify whether the random sampler is working as expected right? One idea is to check if the sampler produces the same output when given the same random seed or random values, and different outputs when  the seed or values change. Secondly if we want to compare outputs with and without the sampler cam we use cosine similarity with a threshold instead of exact matching? im open for suggestions

tests/transformers/sampler/test_sampler.py
quic-sanising: Changed model to `TinyLlama/TinyLlama-1.1B-Chat-v1.0`

QEfficient/generation/text_generation_inference.py
```diff
@@ -342,6 +345,15 @@ def cloud_ai_100_exec_kv(
         :Write_io_dir (str): Path to write the input and output files. ``Defaults to None``.
         :automation (bool): If true, it prints input, output, and performance stats. ``Defaults to False``.
         :prompt_to_lora_id_mapping (List[int]): Mapping to associate prompts with their respective LoRA adapter.
+        :include_sampler (bool): Enable/Disable sampling of next tokens.
```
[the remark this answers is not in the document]
quic-sanising: Done

QEfficient/generation/text_generation_inference.py
```diff
@@ -411,14 +426,59 @@ def __init__(
...
         self.is_tlm = is_tlm
+        self.include_sampler = include_sampler
+        self.return_pdfs = return_pdfs
+        self.sampling_params = sampling_params
 
         # Load QPC
         self._session = QAICInferenceSession(qpc_path, device_id, enable_debug_logs=enable_debug_logs)
 
+        # Validate sampler inputs for On-Device Sampling
+        sampler_inputs = [
```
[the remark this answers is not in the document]
quic-sanising: Done

QEfficient/generation/text_generation_inference.py
quic-sanising: Yes, done

tests/transformers/sampler/test_sampler.py
quic-sanising: Done

QEfficient/generation/text_generation_inference.py
quic-sanising: Could you clarify this?

tests/transformers/sampler/test_sampler.py
quic-sanising: Thanks for the suggestion! The current test checks for exact output matches, which ensures reproducibility with a fixed seed. As for using cosine similarity or other metrics like perplexity or rogue score, they are useful for semantic comparison but can be vague as there are no clear-defined or universally accepted thresholds. For this test, exact matching keeps things deterministic and easier to maintain. Happy to explore alternatives...

QEfficient/generation/text_generation_inference.py
```diff
@@ -411,14 +428,25 @@ def __init__(
...
         write_io_dir: Optional[str] = None,
         is_tlm: Optional[int] = None,
+        include_sampler: bool = False,
+        return_pdfs: bool = False,
+        sampling_params: Optional[Dict[str, Any]] = None,
     ) -> None:
         self._ctx_len = ctx_len
         self._write_io_dir = write_io_dir
         self.is_tlm = is_tlm
+        self.include_sampler = include_sampler
```
quic-rishinr, line 438: do we need this? as we are assigning self.include_sampler =  validate_sampler_inputs() on line number 446

QEfficient/generation/text_generation_inference.py
```diff
@@ -411,14 +428,25 @@ def __init__(
...
+        self.return_pdfs = return_pdfs
+        self.sampling_params = sampling_params
 
         # Load QPC
         self._session = QAICInferenceSession(qpc_path, device_id, enable_debug_logs=enable_debug_logs)
 
+        # Validate sampler inputs for On-Device Sampling
+        self.include_sampler = validate_sampler_inputs(
```
quic-rishinr, line 446: can we have something like self.include_sampler = validate_sampler_inputs(set(self._session.input_names), include_sampler if include_sampler is not False else False

tests/transformers/sampler/test_sampler.py
```diff
@@ -0,0 +1,334 @@
...
+        mxint8_kv_cache=True,
+        mxfp6_matmul=True,
+    )
+
+    # Generate texts from prompts
+    tokenizer = load_hf_tokenizer(pretrained_model_name_or_path=model)
+    model_w_sampler_exec_info = model_w_sampler.generate(
+        tokenizer=tokenizer,
+        prompts=prompts,
+        device_id=None,
```
quic-rishinr, line 157: you can remove the device ID completely from test. Anyway default value is None

quic-rishinr (changes requested):

QEfficient/generation/text_generation_inference.py
quic-sanising: Right, removed.

QEfficient/generation/text_generation_inference.py
quic-sanising: I don't think we need to do this. `include_sampler` is a mandatory boolean variable with a default value of `False`. The only other possible value is `True`. Both of these scenarios are handled in `validate_sampler_inputs()` function.

QEfficient/generation/text_generation_inference.py
quic-sanising: I still do not follow. Are you talking about this reshape?
```py
return outputs["next_tokens"].reshape(outputs["next_tokens"].shape[0], outputs["next_tokens"].shape[1])
```

If yes, this converts 3D tensor of shape `(batch_size, 1, 1)` to a 2D tensor of shape `(batch_size, 1)`. There is no operation that can cause a drop in performance. 

If you are still curious about the performance gains, please reach out to me for a complete performance report.

.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.5.2
+    rev: v0.12.7
```
ochougul, line 4: why?

examples/on_device_sampling.py
ochougul, line 1: are the print statements intentional here?

quic-sanising: What print statements?

quic-hemagnih:
Anything pending for this PR? @ochougul  @quic-sanising

quic-sanising:
> Anything pending for this PR? @ochougul @quic-sanising

Nothing from my end. Please go ahead with the merge if the CI has passed.

QEfficient/cloud/infer.py
```diff
@@ -231,7 +231,6 @@ def main(
             tokenizer,
             prompts=prompt,
             device_id=device_group,
-            prompt=prompt,
```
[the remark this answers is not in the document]
quic-rishinr: As rightly shown update the infer to follow the same.

ochougul (approved):

quic-hemagnih (approved):

PR #466, opened 2025-06-19

quic-vjanfaza:
Context-Length-Specialization technique optimizes the throughput of large language models (LLMs) on Qualcomm devices when handling very large context lengths. The current Ahead Of Time (AOT) compilation on Qualcomm devices doesn't predict the number of tokens needed, leading to significant throughput drops during the prefilling and the decoding phases. This happens because the system performs attention calculations based on large context length. To address this issue, we introduce Compute Context Length (CCL), an additional ONNX variable that allows for dynamic context-length specialization. By generating tokens using smaller, more manageable context lengths (CCL), we optimize memory reads and attention calculations, thereby improving throughput.


QEfficient/cloud/infer.py
```diff
@@ -102,6 +102,7 @@ def main(
     full_batch_size: Optional[int] = None,
     prompt_len: int = 32,
     ctx_len: int = 128,
+    comp_ctx_lengths: Optional[List[int]] = None,
```
quic-amitraj, line 105: Add this in the docstring as well of the function.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1489,6 +1491,8 @@ def from_pretrained(
 
         kv_offload = kwargs.pop("kv_offload", None)
```
quic-amitraj, line 1493: I think, `comp_ctx_length` should be handled as a explicit parameter in `from_pretrained` rather than handling inside the kwargs. After this there will be no need of pooping that var from kwargs and we can add proper docstring as well here.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1419,6 +1419,8 @@ def __init__(
         if self.is_tlm:
             self.model.qaic_config["return_pdfs"] = True
 
+        self.comp_ctx_lengths = kwargs.pop("comp_ctx_lengths", None)
+
```
quic-amitraj, line 1423: Why, I don't think there is any need of this.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1788,26 +1804,45 @@ def compile(
...
-                num_speculative_tokens=num_speculative_tokens,
-            )
-            if decode_spec:
-                specializations.append(decode_spec)
+            if self.comp_ctx_lengths is not None:
+                # Adding elements from self.comp_ctx_lengths to decode_specialization
+                for i in range(1, len(self.comp_ctx_lengths)):
+                    decode_spec = self.build_decode_specialization(
+                        prefill_seq_len=prefill_seq_len,
+                        ctx_len=ctx_len,
```
quic-amitraj, line 1825: There is no need of if else condition, please handle this for loop inside the `build_decode_specialization`.

quic-amitraj (changes requested):

quic-rishinr:
Closing this PR and changes will be tracked as part of #576

PR #467, opened 2025-06-20
Bug fix for spdTransform

qcdipankar:
Added fix for spdtransform due to change in hash

### DIFF
--- a/QEfficient/transformers/models/pytorch_transforms.py
+++ b/QEfficient/transformers/models/pytorch_transforms.py
@@ -504,6 +504,7 @@ class SpDTransform:
     def apply(cls, model: nn.Module, qaic_config: Optional[dict] = None, **kwargs) -> Tuple[nn.Module, bool]:
         transformed = False
         kwargs.pop("pretrained_model_name_or_path")
+
         if qaic_config is None or (speculative_model_type := qaic_config.get("speculative_model_type")) is None:
             return model, transformed
         elif speculative_model_type not in (
--- a/QEfficient/transformers/models/pytorch_transforms.py
+++ b/QEfficient/transformers/models/pytorch_transforms.py
@@ -503,7 +503,7 @@ class SpDTransform:
     @classmethod
     def apply(cls, model: nn.Module, qaic_config: Optional[dict] = None, **kwargs) -> Tuple[nn.Module, bool]:
         transformed = False
-        kwargs.pop("pretrained_model_name_or_path")
+        kwargs.pop("pretrained_model_name_or_path", None)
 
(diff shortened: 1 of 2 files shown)


QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -503,6 +503,8 @@ class SpDTransform:
     @classmethod
     def apply(cls, model: nn.Module, qaic_config: Optional[dict] = None, **kwargs) -> Tuple[nn.Module, bool]:
         transformed = False
+        kwargs.pop("pretrained_model_name_or_path", None)
+
```
ochougul, line 507: This is passed by reference. We are losing this info here. we would want to use it later for hash creation.

qcdipankar: We have two such values one needs to be popped out. That is the error they have reported in the Jira. This has been added after discussion with @quic-rishinr. Thanks

quic-rishinr: We shouldnt pop pretrained_model_name_or_path inside SpDTransform as this value is initialized only after the transform is applied for caching purposes. Instead it can be popped and assigned within the AutoModelForCausalLM initialization method prior to applying the SpDTransform.

ochougul: assigning and popping will also not help because this is pass by reference

> We have two such values one needs to be popped out. That is the error they have reported in the Jira. This has been added after discussion with @quic-rishinr. Thanks

You could pop and add back or make a copy of `kwargs` and pass the copy to transform. Please consider future improvements, such as hashing, while writing the code.

qcdipankar: This change has been suggested by @quic-rishinr today in the morning. I would request you to have a word with him. Since this is a P1 Jira would need to be closed asap. The error they are getting is 

pretrained_model_name_or_path is passed twice:

Some weights of the model checkpoint at /local/mnt/p_drive/qranium/hf_home/hub/models-PredibaseShareQualcomm-private-llama-3-1-8b-instruct-merged-qualcomm-conllpp/snapshots/b6f4dd92bed981126d9787522381615d664201c5/ were not used when initializing LlamaForCausalLM: ['projections.0.0.linear.bias', 'projections.0.0.linear.weight', 'projections.1.0.linear.bias', 'projections.1.0.linear.weight', 'projections.2.0.linear.bias', 'projections.2.0.linear.weight']

ochougul (approved):
Approving now. But this needs to be handled in a cleaner way later.
by `deepcopying kwargs`, and hashing the `kwargs` instead of `pretrained_model_name_or_path`

quic-rishinr (approved):

PR #477, opened 2025-06-23
[QEff. Finetune]: Added support to sync gradients across devices during backward step only.

quic-meetkuma:


QEfficient/finetune/utils/train_utils.py
```diff
@@ -174,38 +178,29 @@ def train(
                 break
             batch = {k: v.to(device) for k, v in batch.items()}  # move the batch elements to qaic device
 
-            with (
-                torch.autocast(device_type=device_type, dtype=torch.float16)
-                if train_config.use_autocast
-                else nullcontext()
-            ):
+            is_backward_step = (step + 1) % train_config.gradient_accumulation_steps == 0 or step == len(
```
quic-swatia, line 181: Since backward() will happen at every step and scaler.step(optimizer) will happen only when the condition is True,, better to name this flag as is_optimizer_step.

quic-meetkuma: Good suggestion. Thanks.

quic-mamta (approved):

PR #478, opened 2025-06-24
[QEff Finetune] Adding dataset padding changes

quic-swatia:
Padding the dataset with dummy samples (they won't contribute in total_loss) to make the #samples a multiple of degree of ddp*batch_size) in case of
1) Fine tuning through DDP 
2) train_batch_size > 1 or val_batch_size > 0


QEfficient/finetune/utils/dataset_utils.py
```diff
@@ -64,19 +64,35 @@ def get_dataloader_kwargs(train_config, dataset, dataset_processer, split):
 
 def get_dataloader(tokenizer, dataset_config, train_config, split: str = "train"):
     dataset = get_preprocessed_dataset(tokenizer, dataset_config, split, context_length=train_config.context_length)
-    dl_kwargs = get_dataloader_kwargs(train_config, dataset, tokenizer, split)
+    dataset = dataset.select(range(0, 10))
```
quic-meetkuma, line 67: Why slicing dataset to pick first 10 samples?

QEfficient/finetune/utils/dataset_utils.py
```diff
@@ -64,19 +64,35 @@ def get_dataloader_kwargs(train_config, dataset, dataset_processer, split):
...
+    dataset = dataset.map(lambda x: {"input_length": len(x["input_ids"])})
+    dataset = dataset.sort("input_length")
```
quic-meetkuma, line 69: Why sorting?

QEfficient/finetune/utils/dataset_utils.py
```diff
@@ -64,19 +64,35 @@ def get_dataloader_kwargs(train_config, dataset, dataset_processer, split):
...
+    dataset = dataset.remove_columns("input_length")
+    dummy_row = next(iter(dataset))
+    dummy_row["labels"] = [-100] * len(dummy_row["labels"])
+    padding_size = 0
+    num_replicas = dist.get_world_size()
+    if len(dataset) % num_replicas > 0:
```
quic-meetkuma, line 75: bs>1 is not considered here.

QEfficient/finetune/utils/dataset_utils.py
```diff
@@ -64,19 +64,35 @@ def get_dataloader_kwargs(train_config, dataset, dataset_processer, split):
...
+        padding_size = num_replicas - len(dataset) % num_replicas
+
+    dummy_data = [dummy_row.copy() for _ in range(padding_size)]
```
quic-meetkuma, line 78: L78 to L80 can be refactored.

QEfficient/finetune/utils/train_utils.py
```diff
@@ -192,6 +192,9 @@ def train(
                     ) as verifier:
                         model_outputs = model(**batch)
                         loss = model_outputs.loss  # Forward call
+                        if (batch["labels"] != -100).sum() == 0:
+                            loss = loss.nan_to_num(nan=0.0)
```
quic-meetkuma, line 196: This loss is zeroed for dummy samples. But the total loss is averaged across all samples including dummy samples. Correct it.

QEfficient/finetune/utils/train_utils.py
```diff
@@ -237,6 +242,9 @@ def train(
                     step_metric_val = float(torch.exp(loss.detach().float()))
                 train_step_metric.append(step_metric_val)
 
+            # Accumalate gradients
+            loss = loss / train_config.gradient_accumulation_steps
```
quic-meetkuma, line 246: This should change.
E.g. 100 samples, 30 global bs.

For first 30 samples, loss = loss / 30
For next 30 samples, loss = loss / 30
For next 30 samples, loss = loss / 30
For last 10 samples, loss = loss / 10 not loss = loss / 30

QEfficient/finetune/utils/train_utils.py
```diff
@@ -439,6 +447,9 @@ def evaluation_helper(model, train_config, eval_dataloader, device):
                 outputs = model(**batch)
             loss = outputs.loss
 
+            if (batch["labels"] != -100).sum() == 0:
+                loss = loss.nan_to_num(nan=0.0)
```
quic-meetkuma, line 451: Same comment as above

QEfficient/finetune/utils/dataset_utils.py
```diff
@@ -64,19 +64,35 @@ def get_dataloader_kwargs(train_config, dataset, dataset_processer, split):
...
+    dummy_dataset = datasets.Dataset.from_list(dummy_data)
+    combined_dataset = datasets.concatenate_datasets([dataset, dummy_dataset])
```
quic-meetkuma, line 80: try to enclose this padding logic in separate function.

quic-meetkuma (commented):
Please generate the ppl numbers across different ddp devices, grad accum step to make this change concrete.

QEfficient/finetune/utils/dataset_utils.py
quic-swatia: Done the sorting here for the non padded dataset in place of being done in sampler.py for the padded dataset to keep the dummy samples in the end.

QEfficient/finetune/utils/dataset_utils.py
quic-swatia: Had skipped it since we are not supporting bs > 1 as of now. Made this change for sake of completion.

QEfficient/finetune/utils/train_utils.py
quic-swatia: Done.

QEfficient/finetune/utils/train_utils.py
quic-swatia: Done.

QEfficient/finetune/utils/dataset_utils.py
quic-swatia: Done.

QEfficient/finetune/utils/dataset_utils.py
quic-swatia: Found this way cleaner.

quic-swatia: As discussed, proceeding with the current approach.

QEfficient/finetune/data/sampler.py
```diff
@@ -45,11 +43,18 @@ def __len__(self):
...
 class DistributedLengthBasedBatchSampler(torch.utils.data.BatchSampler):
     def __init__(
-        self, data_source, batch_size: int, num_replicas: int, rank: int, shuffle: bool = True, seed: int = 0
+        self,
+        data_source,
+        batch_size: int,
+        num_replicas: int,
+        rank: int,
+        shuffle: bool = True,
+        bool=True,
```
quic-meetkuma, line 52: please remove this. This is added by mistake.

QEfficient/finetune/dataset/samsum_dataset.py
```diff
@@ -9,7 +9,7 @@
 
 
 def get_preprocessed_samsum(dataset_config, tokenizer, split, context_length=None):
-    dataset = datasets.load_dataset("Samsung/samsum", split=split, trust_remote_code=True)
+    dataset = datasets.load_dataset("knkarthick/samsum", split=split, trust_remote_code=True)
```
quic-meetkuma, line 12: do not touch this. There is another PR which is completely removing samsum dataset.

QEfficient/finetune/utils/dataset_utils.py
```diff
@@ -54,27 +54,56 @@ def get_dataloader_kwargs(train_config, dataset, dataset_processer, split):
                 dataset, num_replicas=dist.get_world_size(), rank=dist.get_rank(), shuffle=False
             )
             kwargs["batch_size"] = batch_size
-            kwargs["drop_last"] = True
+            kwargs["drop_last"] = False
```
quic-meetkuma, line 57: Why we do not need to set drop_last argument for L45? (For DDP + Sorting case).

QEfficient/finetune/utils/dataset_utils.py
```diff
@@ -54,27 +54,56 @@ def get_dataloader_kwargs(train_config, dataset, dataset_processer, split):
...
 
+def padding_dataset(train_config, dataset):
+    dataset = dataset.map(lambda x: {"input_length": len(x["input_ids"])})
+    if train_config.enable_sorting_for_ddp:
+        dataset = dataset.sort("input_length")
+        dataset = dataset.remove_columns("input_length")
+    dummy_row = next(iter(dataset))
+    dummy_row["labels"] = torch.tensor([-100] * len(dummy_row["labels"]))
+    padding_size = 0
+    num_replicas = 1
```
quic-meetkuma, line 73: L73 to L75 can be simplified by defining below helper function and use it.

def get_num_ddp_devices():
    return int(os.getenv("WORLD_SIZE", 1))

QEfficient/finetune/utils/dataset_utils.py
```diff
@@ -54,27 +54,56 @@ def get_dataloader_kwargs(train_config, dataset, dataset_processer, split):
...
+    dummy_data = [dummy_row.copy() for _ in range(padding_size)]
+    dummy_dataset = datasets.Dataset.from_list(dummy_data)
+    combined_dataset = datasets.concatenate_datasets([dataset, dummy_dataset])
+    return combined_dataset
+
+
 def get_dataloader(tokenizer, dataset_config, train_config, split: str = "train"):
     dataset = get_preprocessed_dataset(tokenizer, dataset_config, split, context_length=train_config.context_length)
+
+    if (
```
quic-meetkuma, line 88: Why should we only use this logic in case of ddp ?

Non ddp case with 100 samples and bs=3 would drop last 1 sample. Don't you think this is required for that case as well?

QEfficient/finetune/utils/train_utils.py
```diff
@@ -201,15 +212,25 @@ def train(
                 else:
                     model_outputs = model(**batch)
                     loss = model_outputs.loss  # Forward call
+                    if (batch["labels"] != -100).sum() == 0:
+                        loss = loss.nan_to_num(nan=0.0)
+                        num_dummy_samples += train_config.val_batch_size
```
quic-meetkuma, line 217: Why train_config.val_batch_size?

QEfficient/finetune/utils/train_utils.py
```diff
@@ -201,15 +212,25 @@ def train(
...
+                    else:
+                        num_dummy_samples_per_batch = (
+                            (torch.sum(batch["labels"] == -100, dim=1) == batch["labels"].shape[1]).sum().item()
+                        )
+                        if num_dummy_samples_per_batch > 0:
+                            num_dummy_samples += num_dummy_samples_per_batch
+                            loss = loss * train_config.val_batch_size / num_dummy_samples_per_batch
```
quic-meetkuma, line 224: Why train_config.val_batch_size?

QEfficient/finetune/utils/train_utils.py
```diff
@@ -296,14 +328,31 @@ def train(
 
         if loss_0_counter.item() == train_config.convergence_counter:
             if train_config.use_peft and train_config.from_peft_checkpoint and epoch == intermediate_epoch:
-                train_epoch_loss = total_loss / (step - intermediate_step)
+                train_epoch_loss = (
+                    0.0
+                    if total_loss == 0.0
+                    else total_loss / (step - intermediate_step - num_dummy_samples / train_config.train_batch_size)
```
quic-meetkuma, line 334: num_dummy_samples / train_config.train_batch_size this will perform float division. I think the intension is to remove the padded batches from avg loss computation. We should use // to perform int division. Do the same in below cases and eval case.

quic-meetkuma (changes requested):
Please try to fix the comments at the earliest so that this can be merged ASAP.

QEfficient/finetune/utils/train_utils.py
```diff
@@ -201,15 +212,25 @@ def train(
...
+
                     if train_config.task_type == "seq_classification":
                         logits = model_outputs.logits
                         labels = batch["labels"][:, 0]
                         preds = torch.nn.functional.softmax(logits, dim=-1)
                         acc_helper.forward(preds, labels)
 
             total_loss += loss.detach().float()
-            # Accumalate gradients
-            loss = loss / train_config.gradient_accumulation_steps
```
quic-mamta, line 212: Please don't make this division after tensorboard plotting and json dump, this division makes the loss to be equivalent for single data sample, during validation also we plot the loss value of single data sample, making this division after plotting will show incorrect values on plot.(loss (train) of grad_accum=4 samples  vs loss (val) of 1 sample

QEfficient/finetune/utils/dataset_utils.py
quic-swatia: As discussed, it's already taken care of in the sampler.py, hence it is not required.

QEfficient/finetune/utils/dataset_utils.py
quic-swatia: This condition would have prevented calling padding_data function in the remaining cases as it is not required in them. After discussion, removed the condition to reduce lines of code.

QEfficient/finetune/utils/train_utils.py
quic-swatia: As experimented, it will scale down the loss value by the factor of gradient_accumulation_steps and in two different runs with different value of gradient_accumulation_steps, loss values per step will vary by the factor of gradient_accumulation_steps in both metric_json file and the tensorboard. Hence. this change is required.

quic-mamta (approved):

quic-meetkuma (approved):
LGTM, good work in identifying all the corner cases and fixing those, Swati! :)

PR #481, opened 2025-06-24
Changing the hashing methodology for cache folder creation of models.  

quic-dhirajku:
Detaching hash function for model cache path calculation. changes for QNN compilation not included yet.

Cache folder mechanism has been modified to have a parent directory for a model based on the architecture that we retrieve from the model config. The hash calculation for the ONNX export now incorporates all model kwargs as well as export kwargs and parameters. the parameters that were used to create the hash also gets dumped as a serialized JSON file in the ONNX folder, the same happens for the compile parameters inside the respective qpc folder.

### DIFF
--- a/QEfficient/base/modeling_qeff.py
+++ b/QEfficient/base/modeling_qeff.py
@@ -5,7 +5,7 @@
 #
 # ----------------------------------------------------------------------------
 
-import hashlib
+# import hashlib
 import inspect
 import json
 import logging
@@ -23,8 +23,8 @@ from QEfficient.base.onnx_transforms import OnnxTransform
 from QEfficient.base.pytorch_transforms import PytorchTransform
 from QEfficient.compile.qnn_compiler import compile as qnn_compile
 from QEfficient.generation.cloud_infer import QAICInferenceSession
-from QEfficient.utils import constants, dump_qconfig
-from QEfficient.utils.cache import QEFF_HOME, to_hashable
+from QEfficient.utils import constants, dump_qconfig, make_serializable
+from QEfficient.utils.cache import QEFF_HOME, hash_dict_params
 
 logger = logging.getLogger(__name__)
 
@@ -46,12 +46,23 @@ class QEFFBaseModel(ABC):
     def _transform_names(cls) -> List[str]:
         return [x.__name__ for x in cls._pytorch_transforms + cls._onnx_transforms]
 
-    def __init__(self, model: torch.nn.Module) -> None:
+    def __init__(self, model: torch.nn.Module, **kwargs) -> None:
         super().__init__()
         self.model = model
+
+        # Store Model parameters to Calculate Hash for caching
+        self.model_params = {}
+        self.model_params.update(kwargs)
+        self.model_params["config"] = self.model.config.to_diff_dict()
+        self.model_params["_transform_names"] = self._transform_names()
+        self.compile_params = {}
+
+        if hasattr(self.model.config, "architectures"):
+            self.model_architecture = self.model.config.architectures[0]
         self.onnx_path: Optional[str] = None
         self.qpc_path: Optional[str] = None
         self.qpc_session: Optional[QAICInferenceSession] = None
+        self.pretrained_model_name_or_path = kwargs.get("pretrained_model_name_or_path", None)
 
         # Apply the transformations
         any_transformed = False
@@ -67,10 +78,6 @@ class QEFFBaseModel(ABC):
     @property
     @abstractmethod
     def model_name(self) -> str: ...
-
-    @property
-    @abstractmethod
-    def model_hash(self) -> str: ...
 
     @abstractmethod
     def export(self, export_dir: Optional[str] = None) -> Path:
@@ -135,8 +142,18 @@ class QEFFBaseModel(ABC):
             :onnx_transform_kwargs (dict): Additional arguments to be passed to `Transform.apply` for this class.
             :export_dir (str): Specify the export directory. The export_dir will be suffixed with a hash corresponding to current model.
         """
-        export_dir = Path(export_dir or (QEFF_HOME / self.model_name))
-        export_dir = export_dir.with_name(export_dir.name + "-" + self.model_hash)
+        self.model_params["output_names"] = output_names
+        self.model_params["dynamic_axes"] = dynamic_axes
+
+        if export_kwargs is not None:
+            self.model_params.update(export_kwargs)
+        if onnx_transform_kwargs is not None:
+            self.model_params.update(onnx_transform_kwargs)
+        export_dir = Path(export_dir or (QEFF_HOME / self.model_architecture / self.model_name))
+
+        export_hash = hash_dict_params(self.model_params)
+        export_hash = export_hash.hexdigest()[:16]
+        export_dir = export_dir.with_name(export_dir.name + "-" + export_hash)
         onnx_path = export_dir / f"{self.model_name}.onnx"
         if onnx_path.is_file():
             self.onnx_path = onnx_path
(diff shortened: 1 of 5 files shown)


quic-amitraj (changes requested):
To control the user should not initialize the model directly using the __init__ of modeling class. Please use `Metaclass Control` to control the flow and print the warning or error. 

```
class NoInitMeta(type):
    def __call__(cls, *args, **kwargs):
        raise RuntimeError("Use `from_pretrained` to create an instance.")

class MyModel(metaclass=NoInitMeta):
    def __init__(self, data):
        self.data = data

    @classmethod
    def from_pretrained(cls, path):
        instance = object.__new__(cls)
        instance.__init__(f"Loaded from {path}")
        return instance
```
More about this you can read from here- https://stackoverflow.com/questions/100003/what-are-metaclasses-in-python.
Put that meta class in the utils and use this for all the modelling class.

QEfficient/base/modeling_qeff.py
```diff
@@ -357,6 +388,19 @@ def _compile(
         logger.info(f"Running compiler: {' '.join(command)}")
         try:
             subprocess.run(command, capture_output=True, check=True)
+
+            # Dumping compile paramters in a JSON file after successful QPC compilation
```
quic-amitraj, line 392: Remove all the code related to `compile_param_json` from here including `dumping` and handle these inside the decorator `dump_qconfig`. Lets keep the base methods clean.

QEfficient/base/modeling_qeff.py
```diff
@@ -203,6 +221,20 @@ def _export(
             onnx.save(model, onnx_path)
             logger.info("Transformed onnx saved")
 
+            # Dumping model paramters in a JSON file after successful ONNX export
+            model_params_json = export_dir / "model_params.json"
+            with open(model_params_json, "w") as fp:
+                json.dump(
+                    {
```
quic-amitraj, line 228: Same like compile create a decorator and handle all these param updates and dumping inside that.

QEfficient/base/modeling_qeff.py
```diff
@@ -135,8 +141,20 @@ def _export(
             :onnx_transform_kwargs (dict): Additional arguments to be passed to `Transform.apply` for this class.
             :export_dir (str): Specify the export directory. The export_dir will be suffixed with a hash corresponding to current model.
         """
-        export_dir = Path(export_dir or (QEFF_HOME / self.model_name))
-        export_dir = export_dir.with_name(export_dir.name + "-" + self.model_hash)
+        export_params = {}
+        export_params["output_names"] = output_names
+        export_params["dynamic_axes"] = dynamic_axes
+
+        self.model_params["export_params"] = export_params
```
quic-amitraj, line 148: Handle in decorator. Lets keep our base methods clean.

quic-amitraj (changes requested):

QEfficient/base/modeling_qeff.py
quic-hemagnih: Agree, lets write decorator implementation to handle this.

QEfficient/base/modeling_qeff.py
quic-rishinr: 1. There is a typo in "paramters" line 391
2. Define a method for creating compile/export params_json in utils, use the same method for creating the export and compiler json and dump using dump_json method.

quic-rishinr (changes requested):

QEfficient/base/modeling_qeff.py
```diff
@@ -45,12 +53,24 @@ class QEFFBaseModel(ABC):
...
+        model_params["_transform_names"] = self._transform_names()
+        return model_params
+
+    def __init__(self, model: torch.nn.Module, **kwargs) -> None:
         super().__init__()
         self.model = model
+        self.hash_params = self.create_model_params(**kwargs)
+
+        if hasattr(self.model.config, "architectures"):
+            self.model_architecture = self.model.config.architectures[0]
```
quic-rishinr, line 69: This would raise an attribute error if config doesnt have architectures. instead you can use 
self.model_architecture = getattr(self.model.config, "architectures", [None])[0]

QEfficient/base/modeling_qeff.py
```diff
@@ -134,8 +150,18 @@ def _export(
...
-        export_dir = export_dir.with_name(export_dir.name + "-" + self.model_hash)
+
+        export_dir = Path(export_dir or (QEFF_HOME / self.model_architecture / self.model_name))
+        export_hash, filtered_hash_params = filter_and_create_export_hash(
+            model_params=self.hash_params,
+            output_names=output_names,
+            dynamic_axes=dynamic_axes,
+            export_kwargs=export_kwargs,
+            onnx_transform_kwargs=onnx_transform_kwargs,
+            export_dir=export_dir,
```
quic-rishinr, line 161: Why do we need export dir in hash params?

QEfficient/base/modeling_qeff.py
```diff
@@ -210,6 +236,11 @@ def _export(
         finally:
             shutil.rmtree(tmp_onnx_dir, ignore_errors=True)
 
+        # Dump JSON file with hashed parameters
+        hashed_params_export_path = export_dir / "hashed_model_params.json"
```
quic-rishinr, line 240: (Suggestion) would it be better if we name it as export_model_params or something like this? Since we have compile_model_params as well

QEfficient/base/modeling_qeff.py
```diff
@@ -240,12 +271,10 @@ def _compile(
             :mdp_ts_num_devices (int): Number of devices to partition to use Multi-Device Partitioning with tensor-slicing.
             :num_speculative_tokens (int, optional): Number of speculative tokens to take as input for Speculative Decoding Target Language Model.
             :enable_qnn (bool): Enables QNN Compilation. ``Defaults to False.``
-            :qnn_config (str): Path of QNN Config parameters file. Any extra parameters for QNN compilation can be passed via this file. ``Defaults to None.``
-            :compiler_options: Pass any compiler option as input.
-                Any flag that is supported by `qaic-exec` can be passed. Params are converted to flags as below:
+            :qnn_config (str): Path of QNN Config parameters file. ``Defaults to None.``
```
quic-rishinr, line 274: Why its removed?

QEfficient/base/modeling_qeff.py
```diff
@@ -257,11 +286,6 @@ def _compile(
             raise FileNotFoundError(f"ONNX file not found at: {onnx_path}")
 
         if enable_qnn:
```
quic-rishinr, line 288: it should not be removed, is it a rebase issue?

QEfficient/base/modeling_qeff.py
```diff
@@ -299,24 +323,17 @@ def _compile(
...
-        if num_speculative_tokens:
-            compile_hash.update(to_hashable({"num_speculative_tokens": num_speculative_tokens}))
-
-        # Hash the MDP partition config and the number of devices.
-        compile_hash.update(to_hashable(mdp_ts_json))
-        compile_hash.update(to_hashable({"mdp_ts_num_devices": mdp_ts_num_devices}))
-
         # Check if already compiled
-        compile_hash = compile_hash.hexdigest()[:16]
+        compile_hash, hashed_params = hash_compile_params(
```
quic-rishinr, line 327: remove the above comment as its not needed here " # Check if already compiled"

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -164,15 +162,14 @@ class QEFFAutoModel(QEFFTransformersBase):
...
+        super().__init__(model, **kwargs)
 
         # Make Embedding specific transforms like appending pooling
         if pooling:
             self.model, _ = PoolingTransform.apply(self.model, pooling)
 
         self.model.base_model.config.use_cache = True
-
-        self.pretrained_model_name_or_path = kwargs.get("pretrained_model_name_or_path", None)
+        self.hash_params["qeff_class"] = self.__class__.__name__
```
quic-rishinr, line 172: nit: which would be better to better readability "qeff_class" or "qeff_auto_class"?

QEfficient/utils/cache.py
```diff
@@ -39,3 +43,11 @@ def to_hashable(obj) -> bytes:
         default=json_serializable,
         sort_keys=True,
     ).encode()
+
+
+def hash_dict_params(dict_items: Dict, hash_string_size: int = HASH_HEXDIGEST_STR_LEN):
```
quic-rishinr, line 48: it would be better to create a new python file in utils and maintain all hash related modules there.

quic-rishinr (changes requested):

PR #482, opened 2025-06-27
[QEff. Finetune]: Removed samsum dataset references from FT code.

quic-meetkuma:
- Removed all the references of samsum dataset from finetuning code.
- Samsum dataset can be used via custom dataset path.


QEfficient/finetune/dataset/dataset_config.py
```diff
@@ -21,14 +21,10 @@
...
     get_preprocessed_imdb as get_imdb_dataset,
 )
-from QEfficient.finetune.dataset.samsum_dataset import (
-    get_preprocessed_samsum as get_samsum_dataset,
-)
 
 DATASET_PREPROC = {
     "alpaca_dataset": partial(get_alpaca_dataset),
     "grammar_dataset": get_grammar_dataset,
-    "samsum_dataset": get_samsum_dataset,
```
quic-swatia, line 31: Just removing this line of code is enough for the finetune.py to not throw DatasetNotFoundError error in case of "--dataset samsum_dataset". It will raise an error as follows: 'finetune.py: error: argument --dataset: invalid choice: 'samsum_dataset' (choose from 'alpaca_dataset', 'grammar_dataset', 'gsm8k_dataset', 'custom_dataset', 'imdb_dataset')'

This way we can still keep the code for samsum_dataset for internal testing purpose and also if huggingface puts back the Samsum dataset, we would just need a single code of line to support it through QEfficient.

quic-swatia (approved):

quic-mamta (approved):

PR #499, opened 2025-07-07
Reading mxfp6_matmul for QNN Compilation path from compile API arguments

ghost:
Added mxfp6_matmul as a recognized argument in QEFFBaseModel::_compile. Its required because for VLM models, qefficient is over-writing mxfp6 value as false for vision case and to use the same value for QNN we are reading it from compile API arguments and made mxfp6 matmul parameter as immutable from qnn_config.json.


quic-rishinr:
Closing this PR. If needed we can reopen it when needed

PR #517, opened 2025-07-11
Logger module in Efficient Transformers

quic-hemagnih:
This PR adds a logger module to the Efficient Transformer, enabling users to log messages at various levels such as INFO, DEBUG, WARNING, and ERROR.
It also provides the flexibility to configure the logging level based on the user's specific execution needs.

### DIFF
--- a/QEfficient/__init__.py
+++ b/QEfficient/__init__.py
@@ -5,6 +5,7 @@
 #
 # -----------------------------------------------------------------------------
 
+import atexit
 import os
 import warnings
 
@@ -16,7 +17,12 @@ from QEfficient.utils import custom_format_warning
 os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
 # Placeholder for all non-transformer models registered in QEfficient
 import QEfficient.utils.model_registery  # noqa: F401
-from QEfficient.utils.logging_utils import logger
+from QEfficient.utils.logging_utils import QEFFLogger
+
+logger = QEFFLogger.get_logger()
+
+# Register the cleanup function
+atexit.register(QEFFLogger.close_logger)
 
 # custom warning for the better logging experience
 warnings.formatwarning = custom_format_warning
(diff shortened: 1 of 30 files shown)


QEfficient/__init__.py
```diff
@@ -16,7 +17,12 @@
 os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
 # Placeholder for all non-transformer models registered in QEfficient
 import QEfficient.utils.model_registery  # noqa: F401
-from QEfficient.utils.logging_utils import logger
+from QEfficient.utils.logging_utils import QEFFLogger
+
+logger = QEFFLogger.get_logger()
```
vbaddi, line 22: nit: where will this logs be dumped? in the default `QEFF_HOME `path?

QEfficient/__init__.py
```diff
@@ -16,7 +17,12 @@
...
+
+# Register the cleanup function
+atexit.register(QEFFLogger.close_logger)
```
vbaddi, line 25: Can we make the naming as `QEffLogger`, consistent with other library naming styles.

QEfficient/utils/logging_utils.py
```diff
@@ -6,53 +6,72 @@
...
+            QEFFLogger._instance = self.logger
+
+    def _get_formatter(self) -> logging.Formatter:
+        return logging.Formatter(
+            "[%(asctime)s:%(levelname)s:%(threadName)s:%(filename)s:%(lineno)d:%(funcName)s()]:%(message)s"
+        )
+
+    def _initialize_logger(self) -> logging.Logger:
+        # Define the hidden log directory path
+        log_dir = os.path.expanduser("~/.cache/.log")
```
vbaddi, line 32: should be `QEFF_HOME` set by the user instead of default.
Default would be` ~/.cache/qeff_models`, but can be overridden by the user.

vbaddi (changes requested):

PR #38, opened 2024-06-07
add fix for missing xgen pad_token_id

quic-amitraj:
1. Add padding_check_and_fix to handle the case where padding token is out of range of vocab size and not left padded.
2. Removed save_fp32_onnx and save_fp16_onnx from the export.

PR #519, opened 2025-07-17
[QEff. Finetune] Fixed optional torch_qaic imports by importing when it is actually used.

quic-meetkuma:
- Moved import of torch_qaic.profile and torch.qaic.amp.GradScaler to a function where it is actually to be used.
- Added doc string for all the functions defined in helper.py
- Minor refactoring of the code.

### DIFF
--- a/QEfficient/finetune/utils/helper.py
+++ b/QEfficient/finetune/utils/helper.py
@@ -111,7 +111,7 @@ def get_op_verifier_ctx(
         use_op_by_op_verifier (bool): Boolean flag to enable op-by-op verifier.
         device_type (str): Device on which the model is being executed.
         dump_dir (str): Directory to dump the op-by-op verification results.
-        step (int): Number of steps to run the op-by-op verification.
+        step (int): Step number for which the op-by-op verification is to be performed.
         ref_device (str, optional): Device to use as reference for verification.
             Defaults to "cpu".
         ref_dtype (torch.dtype, optional): Data type to use as reference
@@ -163,8 +163,8 @@ def get_grad_scaler(use_grad_scaler: bool, device_type: str) -> Any:
         from torch.qaic.amp import GradScaler as QAicGradScaler
 
         return QAicGradScaler()
-    elif device_type == Device.CUDA:
-        # Lazily imported cuda's GradScaler when it is actually needed.
+    elif (device_type == Device.CUDA) or (device_type == Device.CPU):
+        # Lazily imported torch's GradScaler when it is actually needed.
         from torch.amp import GradScaler
 
         return GradScaler(device_type)
--- a/QEfficient/finetune/utils/helper.py
+++ b/QEfficient/finetune/utils/helper.py
@@ -9,13 +9,9 @@ import json
 import os
 from contextlib import nullcontext
 from enum import Enum
+from typing import Any, ContextManager, List
 
 import torch
-
-try:
-    import torch_qaic.debug as qaic_debug  # noqa: F401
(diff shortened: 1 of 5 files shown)


QEfficient/finetune/utils/helper.py
```diff
@@ -38,41 +34,104 @@ class Task_Mode(str, Enum):
...
 
-def enum_names(enum_cls):
-    return [x.value for x in enum_cls]
+def enum_names(enum_cls: Enum) -> List[str]:
+    """Returns List of Enum members in string format.
+
+    Args:
+        enum_cls (Enum): Enum class reference.
+
+    Returns:
```
quic-swatia, line 43: This seems redundant. Please merge or rewrite line #43-44 and #38 into 1.

QEfficient/finetune/utils/helper.py
```diff
@@ -38,41 +34,104 @@ class Task_Mode(str, Enum):
...
+        int: Rank of the process in which it is being called from.
+    """
     return int(os.getenv("LOCAL_RANK", 0))
 
 
-def is_rank_zero():
+def is_rank_zero() -> bool:
+    """Checks whether the current process is in rank-0 in case of ddp or main
```
quic-swatia, line 60: In the non DDP, it will aways return True so the second half of the statement seems a little mis leading. Better comment would be: Checks whether the current process is in rank-0 in case of DDP. It always return True in the non DDP case.

QEfficient/finetune/utils/helper.py
```diff
@@ -38,41 +34,104 @@ class Task_Mode(str, Enum):
...
+
+    Returns:
+        List[str]: List of Enum members in string format.
+    """
+    return [member.value for member in enum_cls]
 
 
-def get_rank():
+def get_rank() -> int:
+    """Get the current rank of the process. In case of ddp use case it returns
```
quic-swatia, line 50: Good to write DDP in caps everywhere since it's an acronym.

quic-meetkuma (commented):
Addressed comments.

QEfficient/finetune/utils/helper.py
```diff
@@ -38,41 +34,104 @@ class Task_Mode(str, Enum):
...
+
+    Returns:
+        ContextManager: Instance of context manager used to verify the operators.
+    """
+    if (not use_op_by_op_verifier) or (device_type != Device.QAIC):
         return nullcontext()
 
-    filter_config = qaic_debug.DispatchFilterConfig.default(train_device)
+    # Lazily imported qaic_debug when it is actually needed.
+    import torch_qaic.debug as qaic_debug  # noqa: F401
```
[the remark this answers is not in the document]
quic-meetkuma: check other instances as well

quic-swatia (approved):

quic-akuruvil (approved):

QEfficient/finetune/utils/helper.py
```diff
@@ -38,41 +34,104 @@ class Task_Mode(str, Enum):
...
+
+    Returns:
+        bool: Flag to inidicate whether the current process is in rank-0.
+    """
     return get_rank() == 0
 
 
-def get_world_size():
+def get_world_size() -> int:
+    """Get total number of DDP devices being used. Note, in case of non-DDP use
```
quic-mamta, line 70: it returns number of processes instead of devices.(both are different in case of pp+ddp)

quic-meetkuma: Updated in latest.

PR #520, opened 2025-07-21
[QEff. Finetune] Updated handling of custom dataset in FT. Updated finetune.md readme file.

quic-meetkuma:
- Introduced the handling of custom dataset via --dataset_config argument. This argument expects a json file which has parameters to enable custom preprocessing for any dataset. 
- Updated the docs to reflect the changes in the interface of custom dataset usage.

### DIFF
--- a/docs/source/finetune.md
+++ b/docs/source/finetune.md
@@ -104,7 +104,7 @@ tensorboard --logdir runs/<file> --bind_all
         ```json
         "preproc_file": "sample_dataset_preproc.py:preprocessing_fn"
         ```
-    -   The preprocessing function must follow the structure below. The signature of the function should not be altered. The sample illustrates `apply_prompt_template` and `tokenize` as sub-functions, but we can define our own sub-functions as needed. For reference, check the example files in the [./QEfficient/finetune/dataset/](https://github.com/quic/efficient-transformers/tree/main/QEfficient/finetune/dataset) directory.
+    -   The preprocessing function must follow the structure below. The function parameters and the return type of the function should not be altered. The sample illustrates `apply_prompt_template` and `tokenize` as sub-functions, but we can define our own sub-functions as needed. For reference, check the example files in the [./QEfficient/finetune/dataset/](https://github.com/quic/efficient-transformers/tree/main/QEfficient/finetune/dataset) directory.
         ```python
         def preprocessing_fn(dataset_config, tokenizer, split, context_length=None):
             # Load the dataset or read from the disk
@@ -139,12 +139,10 @@ tensorboard --logdir runs/<file> --bind_all
         -   In the sample preprocessing function above, the `split` variable takes its value from `data_config.json`. For the training dataset, the value will be taken from the `"train_split"` key, and for the evaluation/test dataset, it will be taken from the `"test_split"` key.
         -   Additional arguments needed for the preprocessing function can be passed in `data_config.json` and will be available via the `dataset_config` variable within the function. For instance, in the sample config above, `"test_split_ratio"` and `"disc_style"` keys can be used in the preprocessing function to define the test split ratio and style of the dataset. These values are accessed through the `dataset_config` variable. Check out the sample preprocessing file at [./QEfficient/finetune/dataset/custom_dataset/sample_dataset_preproc.py](https://github.com/quic/efficient-transformers/tree/main/QEfficient/finetune/dataset/custom_dataset/sample_dataset_preproc.py).
 
-
-
 3.  **Custom Collate Function for Batching**
     -   When using a batch size greater than 1, we may need to override the default collate (batching different samples together in a batch) behavior by including a `"collate_file"` key in `data_config.json`.
     -   Use the same `"file.py:function"` format. If omitted, the default Hugging Face `DataCollatorForSeq2Seq` is typically used, which pads sequences to the longest length in the batch.
-    -   A custom collate function must have the following signature. The signature of the function should not be altered:
+    -   A custom collate function must follow the structure below. The function parameters and the return type of the function should not be altered:
         ```python
         def get_data_collator(tokenizer):
             # Define and return a custom collate_fn here


QEfficient/finetune/utils/config_utils.py
```diff
@@ -101,7 +105,20 @@ def generate_dataset_config(dataset_name: str) -> Any:
...
     assert dataset_name in supported_datasets, f"Given dataset '{dataset_name}' is not supported."
     # FIXME (Meet): Replace below logic by creating using auto registry of datasets.
-    dataset_config = {k: v for k, v in inspect.getmembers(datasets)}[dataset_name]()
+    dataset_config = {k: v for k, v in inspect.getmembers(qeff_datasets)}[dataset_name]()
+    if dataset_name == "custom_dataset":
+        if custom_dataset_config is None:
+            logger.raise_error(
+                "For 'custom_dataset', please provide dataset config file via 'custom_dataset_config' flag.",
+                RuntimeError,
+            )
```
quic-swatia, line 114: Why are we not handling this in the update_config() instead?

QEfficient/finetune/dataset/custom_dataset/disc_preproc.py
```diff
@@ -0,0 +1,87 @@
+# -----------------------------------------------------------------------------
```
quic-swatia, line 1: If we are keeping this file in the PR, it should be named as the sample file.

docs/source/finetune.md
```diff
@@ -75,38 +75,67 @@ tensorboard --logdir runs/<file> --bind_all
...
+   - A custom collate function must have the following signature:
+     ```python
+     def get_data_collator(tokenizer):
+         # Define and return a custom collate_fn here
+     ```
+
+4. **Passing Additional Configuration Parameters**  
+   You can add custom arguments in `data_config.json`, which will be accessible via the `dataset_config` argument inside your `get_custom_dataset()` function.
+
+5. **Example `data_config.json` File**
```
quic-swatia, line 123: Since, there's  already sample_dataset_config.json created, we can remove the sample from here to keep it the documentation crisp. Either of the two should be added.

QEfficient/finetune/configs/dataset_config.py
```diff
@@ -41,7 +41,5 @@ class imdb_dataset:
 @dataclass
 class custom_dataset:
     dataset: str = "custom_dataset"
-    file: str = "dataset/custom_dataset.py"
     train_split: str = "train"
     test_split: str = "validation"
-    data_path: str = ""
```
quic-swatia, line 47: Why are we removing some of these? These can be overwritten through the json file.

docs/source/finetune.md
```diff
@@ -75,38 +75,67 @@ tensorboard --logdir runs/<file> --bind_all
...
+             # Apply prompt formatting to each datapoint
+
+         def tokenize():
+             # Tokenize the formatted datapoint
+
+         # Apply functions to dataset using map
+         dataset = dataset.map(apply_prompt_template, ...)
+         dataset = dataset.map(tokenize, ...)
+         
+         return dataset
```
quic-swatia, line 108: The existing template looks better to me as it briefly talks about what each function does followed by function call/def

docs/source/finetune.md
```diff
@@ -75,38 +75,67 @@ tensorboard --logdir runs/<file> --bind_all
...
+   }
+   ```
+
+6. **Implementing Custom Preprocessing Logic**  
+   Within your dataset loader function, define `apply_prompt_template()` to manipulate raw data into desired prompt format, and `tokenize()` to convert it into token IDs using the tokenizer.
+
+7. **Reference for Dataset Utilities**  
+   You can refer to existing implementations in the [dataset directory of this repository](https://github.com/quic/efficient-transformers/tree/main/QEfficient/finetune/dataset).
+
+---
```
quic-swatia, line 141: It's better to first explain these points (6 and 7) above and then followed by the sample templates .

docs/source/finetune.md
```diff
@@ -75,38 +75,67 @@ tensorboard --logdir runs/<file> --bind_all
     1) Gradient accumulation: By default, gradient accumulation happens for 4 steps. To update this value, command line argument gradient_accumulation_steps has to be passed. (Example: '--gradient_accumulation_steps 8')
     2) Gradient Checkpointing: By default, gradient checkpointing is disabled. To enable it, command line argument gradient_accumulation_steps has to be passed.
 
-## Fine-Tuning on custom dataset
 
-To run fine tuning for any user specific dataset, prepare the dataset using the following steps:
-
```
quic-swatia, line 81: Most of these points should not be removed, specially  4-7.

QEfficient/finetune/utils/config_utils.py
quic-meetkuma: Here, the dataset_config which is defined at L108 is not a dict but a dataclass. In order to update the params of the existing dataclass we need to convert that into dict, add new keys from custom_dataset_config file and update it and then again construct the dataclass. This code it doing that.

docs/source/finetune.md
```diff
@@ -75,38 +75,67 @@ tensorboard --logdir runs/<file> --bind_all
...
+4. **Passing Additional Configuration Parameters**  
+   You can add custom arguments in `data_config.json`, which will be accessible via the `dataset_config` argument inside your `get_custom_dataset()` function.
+
+5. **Example `data_config.json` File**
+   ```json
+   {
+     "train_split": "train",
+     "test_split": "test",
+     "test_split_ratio": 0.15,
+     "preproc_file": "disc_preprocd.py:get_preprocessed_disc",
```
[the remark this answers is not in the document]
quic-swatia: Please give an example of the use case which will not be satisfied.

docs/source/finetune.md
```diff
@@ -75,38 +75,84 @@ tensorboard --logdir runs/<file> --bind_all
...
+        "test_split": "test",
+        "test_split_ratio": 0.15,
+        "preproc_file": "sample_dataset_preproc.py:preprocessing_fn",
+        "collate_file": "sample_dataset_preproc.py:data_collate_fn",
+        "disc_style": "sarcasm_more"
+        }
+        ```
+
+2.  **Specifying the Preprocessing Function**
+    -   In `data_config.json`, include a `"preproc_file"` key to define the path to your preprocessing Python file and the function within it.
```
quic-swatia, line 101: Here, we are asking the user to give the name of the function (which will do prompt creation and tokenization) in the pre processing file. 

Now, if the user doesn't give the function name over here then  by default QEfficient/finetune/dataset/custom_dataset.py file expects the function named 'preproc_fn' to be present in the pre processing file which is again defined by the user. 

Now, if the user has not defined the function with name 'preproc_fn' in the pre processing file, then it will throw an error that the function is missing. 

Simpler design would be to ask the user to use the same name of function each time in the pre processing file and then there's no need to pass it in the json file as well.

docs/source/finetune.md
```diff
@@ -75,38 +75,84 @@ tensorboard --logdir runs/<file> --bind_all
...
+            # the keys should be 'input_ids', 'attention_mask', and 'labels'.
+            return dataset
+        ```
+        -   In the sample preprocessing function above, the `split` variable takes its value from `data_config.json`. For the training dataset, the value will be taken from the `"train_split"` key, and for the evaluation/test dataset, it will be taken from the `"test_split"` key.
+        -   Additional arguments needed for the preprocessing function can be passed in `data_config.json` and will be available via the `dataset_config` variable within the function. For instance, in the sample config above, `"test_split_ratio"` and `"disc_style"` keys can be used in the preprocessing function to define the test split ratio and style of the dataset. These values are accessed through the `dataset_config` variable. Check out the sample preprocessing file at [./QEfficient/finetune/dataset/custom_dataset/sample_dataset_preproc.py](https://github.com/quic/efficient-transformers/tree/main/QEfficient/finetune/dataset/custom_dataset/sample_dataset_preproc.py).
+
+
+
+3.  **Custom Collate Function for Batching**
```
quic-swatia, line 144: Here, if the user defines a function for data collation in the pre processing file and misses to provide its name in the json file then the current flow of code will not flag it and will run the default huggingface API. 

Hence, here also moving to above suggested  design will fix the issue: ' ask the user to use the same name of function each time in the pre processing file and then there's no need to pass it in the json file as well'

docs/source/finetune.md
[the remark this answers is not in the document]
quic-swatia: Yes, I agree json file would help in the extra arguments.  But why are we giving the flexibility to change the name of the pre processing function when the signature is fixed. It is creating confusion as explained in my below comments. :)

docs/source/finetune.md
quic-meetkuma: Discussed offline and come to a conclusion where we will enforce the file name and function name as mandatory arguments. That will leave no room for confusion.

QEfficient/finetune/dataset/custom_dataset/sample_dataset_preproc.py
```diff
@@ -0,0 +1,87 @@
...
+
+import datasets
+from transformers.data import DataCollatorForSeq2Seq
+
+
+def get_data_collator(tokenizer):
+    return DataCollatorForSeq2Seq(tokenizer)
+
+
+def get_preprocessed_disc(dataset_config, tokenizer, split, context_length=None):
```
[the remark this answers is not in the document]
quic-meetkuma: Had an offline discussion and decided to go ahead with the current approach, because using kwargs here and then unpacking dataset_config from it and then unpacking user provided (via json) arguments from dataset_config does not look like a better design. Hence we are thinking of changing this entirely during the refactoring work where the kwargs will take all the values directly from dataset config json file.

docs/source/finetune.md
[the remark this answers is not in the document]
quic-meetkuma: Had some offline discussion and decided to go ahead by keeping function name as mandatory argument.

docs/source/finetune.md
```diff
@@ -75,38 +75,84 @@ tensorboard --logdir runs/<file> --bind_all
...
+        ```
+
+2.  **Specifying the Preprocessing Function**
+    -   In `data_config.json`, include a `"preproc_file"` mandatory key to define the path to your preprocessing Python file and the function within it.
+    -   Use the format `"filename.py:function_name"`. The filename and function name both are required. 
+        _Example:_
+        ```json
+        "preproc_file": "sample_dataset_preproc.py:preprocessing_fn"
+        ```
+    -   The preprocessing function must follow the structure below. The signature of the function should not be altered. The sample illustrates `apply_prompt_template` and `tokenize` as sub-functions, but we can define our own sub-functions as needed. For reference, check the example files in the [./QEfficient/finetune/dataset/](https://github.com/quic/efficient-transformers/tree/main/QEfficient/finetune/dataset) directory.
```
quic-swatia, line 107: Since, the name of the function has become flexible in this PR, we don't want the whole function signature to be fixed. Parameter list and return type should not be altered.

docs/source/finetune.md
```diff
@@ -75,38 +75,84 @@ tensorboard --logdir runs/<file> --bind_all
...
+
+
+
+3.  **Custom Collate Function for Batching**
+    -   When using a batch size greater than 1, we may need to override the default collate (batching different samples together in a batch) behavior by including a `"collate_file"` key in `data_config.json`.
+    -   Use the same `"file.py:function"` format. If omitted, the default Hugging Face `DataCollatorForSeq2Seq` is typically used, which pads sequences to the longest length in the batch.
+    -   A custom collate function must have the following signature. The signature of the function should not be altered:
```
quic-swatia, line 147: Similarly as above, we don't want the whole function signature to be fixed. Parameter list and return type should not be altered.

quic-meetkuma: Thanks, updated in latest.

docs/source/finetune.md
quic-meetkuma: Thanks, updated in latest.

PR #521, opened 2025-07-22
Deleted model after export to save memory

asmigosw:
Deleting the model after export is complete and storing only necessary params required either at compile or generation.

### DIFF
--- a/QEfficient/base/modeling_qeff.py
+++ b/QEfficient/base/modeling_qeff.py
@@ -5,6 +5,7 @@
 #
 # ----------------------------------------------------------------------------
 
+import gc
 import inspect
 import logging
 import shutil
@@ -62,7 +63,7 @@ class QEFFBaseModel(ABC):
         self.model_architecture = (
             (arch := getattr(self.model.config, "architectures", None)) and len(arch) > 0 and arch[0]
         ) or None
-        
+
         # Flag for checking if weights are offloaded
         self._is_weights_offloaded: bool = False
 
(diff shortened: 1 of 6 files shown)


QEfficient/base/modeling_qeff.py
```diff
@@ -141,6 +156,17 @@ def _export(
             self.onnx_path = onnx_path
             return onnx_path
 
+        # Storing state_dict to load model's weights if export called again
+        # if not any(name for name, param in self.model.named_parameters() if param.is_meta):
+        #     self.state_dict = self.model.state_dict()
+
+        # Loading model if weights are in meta state from state_dict
+        if any(name for name, param in self.model.named_parameters() if param.is_meta):
```
ochougul, line 164: can't we do this check without the for loop, this will loop over all the model params, we can check only one right?
or add a attribute to a model. and check based on that.

ochougul (changes requested):
Changes look good to me as first PR.
The re-loading part needs to be figured out, are you planning to do this in next PR?

Please clean the code, we are good to go.

quic-hemagnih (approved):
Looks good to me, please run the CI. We can merge this now.

QEfficient/base/modeling_qeff.py
quic-rishinr: I have added a variable and added an or condition in the latest code

quic-hemagnih (approved):
LGTM

PR #522, opened 2025-07-23
[QEff. Finetune]: For grad checkpointing, changed the preserve_rng_state to True

quic-meetkuma:
- When grad checkpointing enabled, at that time we had provided preserve_rng_state as False because keeping it True was failing due to not setting rng state. Now that issue is solved hence we can set preserve_rng_state as True. 
- With this the runs with and without --gradient_checkpointing flag should produce identical results.

### DIFF
--- a/QEfficient/cloud/finetune.py
+++ b/QEfficient/cloud/finetune.py
@@ -174,7 +174,7 @@ def load_model_and_tokenizer(
     if train_config.gradient_checkpointing:
         # Note: below attribute and method is only available in HuggingFace Transformer models.
         if hasattr(model, "supports_gradient_checkpointing") and model.supports_gradient_checkpointing:
-            model.gradient_checkpointing_enable(gradient_checkpointing_kwargs={"preserve_rng_state": False})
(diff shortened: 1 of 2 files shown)


QEfficient/cloud/finetune.py
```diff
@@ -174,7 +174,7 @@ def load_model_and_tokenizer(
     if train_config.gradient_checkpointing:
         # Note: below attribute and method is only available in HuggingFace Transformer models.
         if hasattr(model, "supports_gradient_checkpointing") and model.supports_gradient_checkpointing:
-            model.gradient_checkpointing_enable(gradient_checkpointing_kwargs={"preserve_rng_state": False})
+            model.gradient_checkpointing_enable(gradient_checkpointing_kwargs={"preserve_rng_state": True})
         else:
```
quic-akuruvil, line 178: rng state was set, by initializing the seed, is it?

quic-akuruvil: Okay, thanks Meet

PR #527, opened 2025-08-01
[QEff. Finetune]: Correcting num_steps trained as per max_train_step and displaying non scaled loss value on console.

quic-swatia:


QEfficient/finetune/utils/train_utils.py
```diff
@@ -164,7 +164,7 @@ def train(
                     continue
             total_train_steps += 1
 
-            if train_config.max_train_step > 0 and total_train_steps >= train_config.max_train_step:
```
quic-meetkuma, line 167: why this is required?

total_train_steps starts from 0. If user mentions that he/she wants to run 20 steps then we should stop the iterations at the total_train_steps == 20 not at 21.

Let me know if I am missing something.

QEfficient/finetune/utils/train_utils.py
```diff
@@ -265,7 +265,7 @@ def train(
                     )
 
             pbar.set_description(
-                f"Training Epoch: {epoch + 1}/{train_config.num_epochs}, step {step + 1}/{len(train_dataloader)} completed (loss: {loss.detach().float()})"
+                f"Training Epoch: {epoch + 1}/{train_config.num_epochs}, step {step + 1}/{len(train_dataloader)} completed (loss: {(loss * num_samples_in_cur_update).detach().float()})"
```
quic-meetkuma, line 268: For batch_size=4, with 3rd sample being a padded sample, the losses are L1, L2, L3 and L4. The loss for L3 is zeroed because it is padded sample. Now, here the loss variable is the avg of all 4 losses. To make it an average of 3 values, we should be multiplying the loss with 4 and divide it with 3.

Let me know is this the problem you are trying to solve? If so, then how your solution helps?

quic-meetkuma (changes requested):

quic-meetkuma:
Can you prefix the title with "[QEff. Finetune]:" so that we can filter out whenever needed in future.

QEfficient/finetune/utils/train_utils.py
quic-swatia: Please check line #165. total_train_steps doesn't start from 0, it's starts from 1. When total_train_steps =1, the first step gets executed. 

In case of '--max_train_step 20',   when step == 19, then total_train_steps == 20 and the previous condition becomes True  and hence it stops the execution. So the training happens only from step 0-18, i.e. 19 steps.

QEfficient/finetune/utils/train_utils.py
quic-swatia: This is not the issue. The use case which you have explained is already taken care at line #200. 

In case of gradient_accumulation, since the per step loss is scaled down by the value of gradient_accumulation_steps, and loss is printed on the console only after the step finishes, the scaled down value was being displayed on the console. Corrected it with the above change.

QEfficient/finetune/utils/train_utils.py
```diff
@@ -326,7 +325,7 @@ def train(
             )
 
             if is_rank_zero():
-                tensorboard_updates.add_scalars("loss", {"eval": eval_epoch_loss}, total_train_steps)
+                tensorboard_updates.add_scalars("loss", {"eval": eval_epoch_loss}, total_train_steps - 1)
```
quic-meetkuma, line 328: I think the better design is to properly use this total_train_steps variable. I suggest below changes to make the design cleaner.

- Remove the updation of total_train_steps at L130.
- At the L152, at the start of each iteration over dataloader update the total_train_steps variable: total_train_steps = len(train_dataloader) * (epoch + 1) + step. This will be helpful throughout each step of dataloader.
- Replace the condition at L157 with the condition of L162. Both are doing the same thing. 
- Remove L158 updation of total_train_steps.
- Remove L164 updation of total_train_steps.
- The condition at L166 should include >= instead of >.
- At L212, the tensorboard should take total_train_steps not 'total_train_steps - 1'.
- At L328, the tensorboard should take total_train_steps not 'total_train_steps - 1'.


With this there will not be a manual +1 or -1 into the total_train_steps. This will make it more maintainable and understandable.

quic-meetkuma (approved):
LGTM, thanks for quick fix. :)

PR #532, opened 2025-08-05
corrections made in jupyter notebooks

smedhe:
The device_group arg is deprecated from compile and added in generate method,

### DIFF
--- a/notebooks/QEfficientMPT.ipynb
+++ b/notebooks/QEfficientMPT.ipynb
@@ -29,12 +29,26 @@
    "execution_count": null,
    "id": "c21f82d5-17df-4fc9-a180-05edd032f02d",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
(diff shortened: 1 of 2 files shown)


notebooks/QEfficientMPT.ipynb
```diff
@@ -91,11 +91,8 @@
     "# Compile the model for provided compilation arguments\n",
     "# Please use platform SDK to Check num_cores for your card.\n",
     "\n",
-    "qeff_model.compile(\n",
-    "    num_cores=14,\n",
-    "    mxfp6=True,\n",
-    "    device_group=[0],\n",
-    ")"
+    "qeff_model.compile(num_cores=14, mxfp6=True)\n",
+    "# the device_group arg is deprecated from compile and added in generate method"
```
quic-rishinr, line 95: its not deprecated instead renamed to device_id. However, you can skip setting device_id explicitly as the device will be automatically selected using the auto device picker during execution

PR #533, opened 2025-08-05
Bug fix for model auto for vision

qcdipankar:
Fix for qeff vision models export is failing for dual qpc method when we providing onnx_dir option to export API.


quic-rishinr:
The same issue appears in other classes like QEffCausalLMForTextImageToTextModel and _QEffAutoModelForImageTextToTextDualQPC. Please review the _export and _compile calls for all these classes to verify that all positional arguments are being passed correctly

quic-hemagnih (commented):
Please check here, if we need to assign export_dir

quic-hemagnih (approved):

PR #537, opened 2025-08-12
Cleaned and Rebased PR for (#481) to change the hash creation module …

quic-dhirajku:
…for all models in Qefficient.

This PR contains changes made to the modelling_qeff, modeling_auto to allow usage of certain export parameters and kwargs passed during model creation. The hashing module is now made independant of the calling class and the test scripts are updated accordingly to test for this functionality. Added functionality to have an overarching parent directory in cache to contain all different exported model configs belonging to the same architecture. In case the architecture isn't present in the config of the model, we instead proceed with  self.model_name based parent directory creation. Hash is now created during export, so as to incorporate all the additional params needed for unique hash creation thus, the test scripts have been modified to test hashing functionalities accordingly. We maintain an Exclusion list of params for kwargs to be discarded during hashing parameter selection. We'll need to look into the alternate approach of maintaining an Inclusion list instead. There was a comment to use MetaClasses to handle raising a warning whenever someone loads a model without using 'from_pretrained' method but the current class architecture of VLMs and SpeechSeq2Seq models don't allow for this, this use case will be handled in a different PR.

### DIFF
--- a/QEfficient/utils/_utils.py
+++ b/QEfficient/utils/_utils.py
@@ -536,13 +536,15 @@ def export_wrapper(func):
         self.export_hash = export_hash
 
         # _EXPORT CALL
-        func(self, *args, **kwargs)
+        onnx_path = func(self, *args, **kwargs)
 
         # POST-PROCESSING
         # Dump JSON file with hashed parameters
         hashed_params_export_path = export_dir / "hashed_export_params.json"
         create_json(hashed_params_export_path, filtered_hash_params)
         logger.info("Hashed parameters exported successfully.")
+
+        return onnx_path
 
     return wrapper
 


QEfficient/transformers/models/modeling_auto.py
```diff
@@ -955,7 +908,7 @@ def __init__(
             self.model.config.vision_config.use_flash_attn = "false"
         else:
             self.model.config.text_config.use_cache = True
-        self.pretrained_model_name_or_path = kwargs.get("pretrained_model_name_or_path", None)
+        self.hash_params["qeff_auto_class"] = self.__class__.__name__
```
quic-hemagnih, line 911: Can we think of moving this to a common place, I see it in all the Auto Classes.

QEfficient/utils/constants.py
```diff
@@ -28,6 +28,47 @@
 # Compiler defaults
 DEFAULT_AIC_NUM_CORES = 16
 DEFAULT_AIC_MXPF6_MATMUL = False
+# Hashing defaults
+HASH_HEXDIGEST_STR_LEN = 16
+# Why not use an Inclusion list instead?
+KWARGS_EXCLUSION_LIST = [
```
quic-hemagnih, line 34: As discussed please prepare the inclusion list instead of exclusion list. This approach will have full control on the set of parameters which are included in hash computation. Exclusion list may lead to surprises.

QEfficient/utils/_utils.py
```diff
@@ -749,6 +751,41 @@ def make_serializable(obj):
     create_json(qconfig_file_path, qconfigs)
 
 
+def filter_and_create_export_hash(**kwargs):
+    """
+    This Method prepares all the model params required to create the hash for export directory.
+    """
+    # TODO: Check and confirm if we want an exclusion list or an inclusion list
+    filtered_params = kwargs["model_params"]
```
quic-hemagnih, line 759: should we use .get() here for model_params

QEfficient/utils/hash_utils.py
```diff
@@ -0,0 +1,43 @@
...
+    if isinstance(obj, set):
+        return sorted(obj)
+    raise TypeError(f"Object of type {obj.__class__.__name__} is not JSON serializable")
+
+
+def to_hashable(obj) -> bytes:
+    """
+    Converts obj to bytes such that same object will result in same hash
+    """
+    return json.dumps(
```
quic-hemagnih, line 25: Should we wrap json.dump in try-except block to catch serialization errors

quic-hemagnih (changes requested):
As discussed please prepare the inclusion list instead of exclusion list. This approach will have full control on the set of parameters which are included in hash computation. Exclusion list may lead to surprises.

QEfficient/base/modeling_qeff.py
```diff
@@ -366,6 +387,10 @@ def _compile(
                     ]
                 )
             )
+        # Dump JSON file with hashed parameters
```
quic-amitraj, line 390: With single decorator you will be able to handle both export and compile. You can put generic name like- @wrap_with_hook, which will be for all the generic preporcess and postprocess of a function.

quic-amitraj (changes requested):

tests/utils/test_hash_utils.py
```diff
@@ -0,0 +1,99 @@
...
+        allow_nan=False,
+        indent=None,
+        separators=(",", ":"),
+        default=json_serializable,
+        sort_keys=True,
+    ).encode()
+    assert to_hashable(obj_with_set) == expected_with_set
+
+
+def test_hash_dict_params():
```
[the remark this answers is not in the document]
quic-dhirajku: Just an additional test to verify that hash_dict_params method is working well. I was asked to add a test for the hashing methods.

PR #538, opened 2025-08-12
Added memory optimization for onnx transforms

quic-rishinr:
Added periodic memory cleanup to FP16ClipTransform and SplitTensorsTransform to reduce memory usage during large tensor processing. Also avoids redundant external data loading when already present.

### DIFF
--- a/QEfficient/transformers/models/modeling_auto.py
+++ b/QEfficient/transformers/models/modeling_auto.py
@@ -3201,7 +3201,7 @@ class QEFFAutoModelForSpeechSeq2Seq(QEFFTransformersBase, MultimodalUtilityMixin
         super().__init__(model, **kwargs)
         self.num_layers = model.config.num_hidden_layers
         self.hash_params["qeff_auto_class"] = self.__class__.__name__
-    
+
     @property
     def model_name(self) -> str:
         """
@@ -3216,7 +3216,7 @@ class QEFFAutoModelForSpeechSeq2Seq(QEFFTransformersBase, MultimodalUtilityMixin
         if mname.startswith("QEff") or mname.startswith("QEFF"):
             mname = mname[4:]
         return mname
-      
+
     @property
     def get_model_config(self) -> dict:
         """


ochougul:
Do we have before and after memory utilization for this change?

QEfficient/base/onnx_transforms.py
```diff
@@ -35,162 +46,218 @@ def apply(cls, model: ModelProto, **kwargs) -> Tuple[ModelProto, bool]:
...
         model: ModelProto,
         *,
-        model_name: str,
+        transforms: List[str],
+        model_name: str = "",
         onnx_base_dir: Optional[str] = None,
-        file_chunk_size: int = 10 * 2**30,  # 10 GiB
+        file_chunk_size: int = 10 * 2**30,
         size_threshold: int = 1024,
+        opset_version: int = 17,
```
quic-rishinr: lets set the default value from constants.py

QEfficient/base/onnx_transforms.py
```diff
@@ -35,162 +46,218 @@ def apply(cls, model: ModelProto, **kwargs) -> Tuple[ModelProto, bool]:
...
 
+        try:
+            cls._load_external_data(model, onnx_base_dir)
+            tensors = external_data_helper._get_all_tensors(model)
 
-class CustomOpTransform(OnnxTransform):
-    """
-    Transform to register custom operations and add their function protos to the ONNX model.
-    """
+            TensorInfo = namedtuple("TensorInfo", ["tensor", "tsize"])
```
quic-rishinr: Do we need this creation of tensor_infos? if we are only extracting the tensor size cant we do it inside the process_tensor method? 
pass enumerate(tensors) directly to process_tensor method instead of tensor_infos and then extract the tensor size by 
tsize = len(tensor.raw_data) if tensor.HasField("raw_data") else 0

QEfficient/base/onnx_transforms.py
```diff
@@ -35,162 +46,218 @@ def apply(cls, model: ModelProto, **kwargs) -> Tuple[ModelProto, bool]:
...
-        transformed = False
+                if "FP16ClipTransform" in requested_transforms:
+                    if cls._clip_tensor(tensor, fp16_min, fp16_max):
+                        local_applied.append("FP16ClipTransform")
 
-        # Register all custom op symbolic functions with torch.onnx
-        for op_name, (func_class, _) in cls._custom_ops.items():
-            if hasattr(func_class, "symbolic"):
-                torch.onnx.register_custom_op_symbolic(f"::{op_name}", func_class.symbolic, opset_version)
+                if "SplitTensorsTransform" in requested_transforms and tsize > size_threshold:
```
quic-rishinr: 1. Can we do split tensor transform in parallel? it would require file_num_tracker["size"] to grow sequentially right?
2. if there are multiple threads trying to update file_num_tracker in parallel it could lead to race conditions, we need to use a thread lock while updating file_num_tracker

QEfficient/base/onnx_transforms.py
abhishek-singh591: all the three values?

QEfficient/base/onnx_transforms.py
```diff
@@ -35,162 +46,218 @@ def apply(cls, model: ModelProto, **kwargs) -> Tuple[ModelProto, bool]:
...
 
-        return model, transformed
+        logger.info("External data and cache cleaned up.")
 
 
-class SplitTensorsTransform(OnnxTransform):
-    """
-    Split external tensors file
-    """
+class OnnxTransform(BaseOnnxTransform):
```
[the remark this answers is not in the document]
abhishek-singh591: done

QEfficient/base/onnx_transforms.py
abhishek-singh591: done.

QEfficient/base/onnx_transforms.py
```diff
@@ -35,162 +46,218 @@ def apply(cls, model: ModelProto, **kwargs) -> Tuple[ModelProto, bool]:
...
+            applied_transforms = {name: False for name in requested_transforms}
 
-    @classmethod
-    def apply(cls, model: ModelProto, *, opset_version: int = 17, **kwargs) -> Tuple[ModelProto, bool]:
-        """
-        Apply custom op registration and add all function protos to the model.
+            def process_tensor(index_info: Tuple[int, TensorInfo]) -> List[str]:
+                idx, info = index_info
+                tensor, tsize = info
+                local_applied = []
```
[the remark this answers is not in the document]
abhishek-singh591: changed it dictionary.

abhishek-singh591:
Please refer to PR #640 for the latest code.

PR #539, opened 2025-08-14
Optimized onnx transform class via multithreading

abhishek-singh591:
###  ONNX Transform Optimization with Thread Pool

Refactored the ONNX transform class to use a thread pool for parallelizing tensor operations, replacing the previous iterative loop. This resulted in a **notable performance boost** in the FP16Clip transform and a **marginal improvement** in `split_tensor()`, which may have further optimization potential.

####  Performance (LLaMA 3.1 8B)

| Operation      | Without Thread Pool | With Thread Pool |
|----------------|---------------------|------------------|
| FP16 Transform | 51.94 sec           | 38.78 sec        |
| Split Tensor   | 15.30 sec           | 13.73 sec        |

Thread count is hardcoded to `os.cpu_count() * 4` to better handle I/O-intensive workloads. Performance may vary depending on the machine's threading capabilities, so results may not be exactly reproducible across environments.


abhishek-singh591:
## LLaMA 3.1 70B Performance Comparison

### Without Thread Pooling
- **Execution time for `f16`**: `461.2928 seconds`
- **Execution time for `Split`**: `540.5107 seconds`

### With Thread Pooling
- **Execution time for `f16`**: `300.3154 seconds`
- **Execution time for `Split`**: *Still running, expected to be lower.*

> Note: I attempted to run `Split` with thread pooling as well, but due to high memory usage from other processes on the machine, the RAM was exhausted and the process was terminated, i will edit this as soon as i get these numbers.

PR #544, opened 2025-08-20
[QEff finetune] : fix qaic device for pp+ddp

quic-mamta:
- Fix qaic device for pp+ddp
- This fix is required for sdk version 1.21
- Without this fix , pp+ddp was working fine on 1.20.0.194


quic-swatia (approved):
Please add the SDK on which it is tested for future reference.

PR #546, opened 2025-08-23
Optimized ONNX Transform via Class Merging and Thread Pooling

abhishek-singh591:
### Optimized ONNX Transform via Class Merging and Thread Pooling

This PR follows up on #539 – Optimized ONNX transform class via multithreading.

It merges the FP16 and Split ONNX transform classes into a single implementation to eliminate redundant tensor loading and iteration. Additionally, the transform logic has been refactored to use a **thread pool**, replacing the previous sequential loop to parallelize tensor operations.

####  Performance Benchmarks:-

| Model           | Original Duration (s) | Optimized Duration (s) |
|----------------|------------------------|-------------------------|
| LLaMA 3.1 8B    | 88.35                  | 58.55                   |
| LLaMA 3.1 70B   | 1029.82                | 727.37                  |

> **Note:** Thread count is set to `os.cpu_count() * 4` to better handle I/O-bound workloads. Performance may vary depending on system hardware and threading capabilities.

### DIFF
--- a/QEfficient/base/onnx_transforms.py
+++ b/QEfficient/base/onnx_transforms.py
@@ -42,7 +42,7 @@ class ClipAndSplitTransform(OnnxTransform):
         cls,
         model: ModelProto,
         *,
-        model_name: str,
+        model_name: str = "",
         onnx_base_dir: Optional[str] = None,
         apply_clip: bool = True,
         apply_split: bool = True,
--- a/tests/base/test_onnx_transforms.py
+++ b/tests/base/test_onnx_transforms.py
@@ -32,7 +32,7 @@ def test_fp16clip_transform():
     }
     """)
     onnx.checker.check_model(test_onnx, True, True, True)
-    transformed_onnx, transformed = ClipAndSplitTransform.apply(test_onnx, model_name="", apply_split=False)
+    transformed_onnx, transformed = ClipAndSplitTransform.apply(test_onnx, apply_split=False)
     assert transformed
     assert onnx.numpy_helper.to_array(transformed_onnx.graph.initializer[0]) == 65504.0
     assert onnx.numpy_helper.to_array(transformed_onnx.graph.initializer[1]) == 2147483647
@@ -64,7 +64,7 @@ def test_fp16clip_transform_external(tmp_path):
     onnx.checker.check_model(onnx_path, True, True, True)
 
     transformed_onnx, transformed = ClipAndSplitTransform.apply(
-        test_onnx, model_name="", onnx_base_dir=str(tmp_path), apply_split=False
+        test_onnx, onnx_base_dir=str(tmp_path), apply_split=False
     )
     assert transformed
     assert onnx.numpy_helper.to_array(transformed_onnx.graph.initializer[0]) == -65504.0


QEfficient/base/onnx_transforms.py
```diff
@@ -5,97 +5,88 @@
...
     @classmethod
     def apply(cls, model: ModelProto, **kwargs) -> Tuple[ModelProto, bool]:
-        """
-        Override this class to apply a transformation.
-        :param model: The model's ONNX graph to transform
-        :param kwargs: Parameters needed for specific transforms. All transforms should take **kwargs to ignore unneeded kwargs.
-
-        :returns: ONNX graph after applying the transform
-        :returns: Boolean indicating whether transform was applied
-        """
```
ochougul, line 31: don't remove this

QEfficient/base/onnx_transforms.py
```diff
@@ -5,97 +5,88 @@
...
     @classmethod
     def apply(
         cls,
         model: ModelProto,
         *,
         model_name: str,
         onnx_base_dir: Optional[str] = None,
-        file_chunk_size: int = 10 * 2**30,  # 10 GiB
+        apply_clip: bool = True,
+        apply_split: bool = True,
```
ochougul, line 35: add warning when both are False and return early without loading external data.

QEfficient/base/onnx_transforms.py
```diff
@@ -5,97 +5,88 @@
...
-        transformed = False
         external_data_helper.load_external_data_for_model(model, onnx_base_dir)
-        for tensor in external_data_helper._get_all_tensors(model):
-            if tensor.HasField("raw_data") and ((tsize := len(tensor.raw_data)) > size_threshold):
+        tensors = external_data_helper._get_all_tensors(model)
+
+        TensorInfo = namedtuple("TensorInfo", ["tensor", "tsize"])
+        tensor_infos = [
+            TensorInfo(tensor, len(tensor.raw_data) if tensor.HasField("raw_data") else 0) for tensor in tensors
+        ]
```
ochougul, line 46: how much time does this line take?
This is actually another loop over tensors; can we not do this? and get this info inside the `process_tensor` func itself?

ochougul (changes requested):

QEfficient/base/onnx_transforms.py
abhishek-singh591: As discussed, this step accounts for roughly 25% of the total ONNX transformation time. The bottleneck is due to the expensive access to "tensor.raw_data". Calculating the size upfront is beneficial for downstream processing, since the list TensorInfo datatype is iterated over later this behavior was also observed empirically.

QEfficient/base/onnx_transforms.py
```diff
@@ -32,70 +36,74 @@ def apply(cls, model: ModelProto, **kwargs) -> Tuple[ModelProto, bool]:
...
-                    file_num += 1
-                    current_file_size = tsize
-                external_data_helper.set_external_data(tensor, f"{model_name}_{file_num}.onnx.data")
-        return model, transformed
+
+            return transformed
+
+        with ThreadPoolExecutor(max_workers=os.cpu_count() * 4) as executor:
+            transformed_flags = list(executor.map(process_tensor, tensor_infos))
+        return model, any(transformed_flags)
```
ochougul, line 89: we should use `return model, all(transformed_flags)`
If any of the transform failed we should return False.

ochougul (changes requested):
LGTM. can merge if CI is passing

QEfficient/base/onnx_transforms.py
abhishek-singh591: In the earlier implementation, the transformed was set to true if both ONNX operations were applied to at least one tensor, even if each operation was applied to a different tensor. To maintain consistency with that logic in the merged function, the transformed_flags should similarly be set to true as long as both ONNX operations have been applied across the tensors regardless of whether they were applied to the same tensor or separate ones. Therefore, using any or all directly would not capture this behavior accurately. I’ll update the code accordingly, unless you'd prefer to revise the original logic instead.

abhishek-singh591: I've updated the code as per our discussion. In the latest commit, the transformed_flag will be set to true when the corresponding ONNX transformation is applied to at least one tensor. This ensures the logic aligns with the intended behavior.

QEfficient/base/onnx_transforms.py
```diff
@@ -32,70 +36,77 @@ def apply(cls, model: ModelProto, **kwargs) -> Tuple[ModelProto, bool]:
...
-        :param size_threshold: Only tensors greater than this threshold (in bytes) will be saved externally.
-        """
-        file_num = 0
-        current_file_size = 0
-        transformed = False
+        if not apply_clip and not apply_split:
+            warnings.warn("Both apply_clip and apply_split are False. Skipping transformation.")
+            return model, False
+
         external_data_helper.load_external_data_for_model(model, onnx_base_dir)
```
quic-rishinr, line 57: Even though combining the transform save time in this case, it also reduces the flexibility we have over multiple transforms. In future if we need to add more transforms the condition would become more complex and if its a new transform would need to load the tensors again. I have added few changes as part #538 of the memory clean and reducing the peak memory usage. Can you check if the same concepts can be used here?

quic-rishinr (changes requested):

QEfficient/base/onnx_transforms.py
abhishek-singh591: Yeah, its kind off tradeoff between time and flexibility, let me check that will get back to you.

PR #549, opened 2025-09-01
Removed redundant input_length column from dataset.

quic-meetkuma:
- This is required because, whenever dataset padding is performed there will be an additional input being fed to the model. For causal lm models somehow this did not gave any error. But for BERT kind of models this gave error.


quic-swatia (approved):
Change looks good to me. Meet confirmed that change works well with PFT FT as well. :)

PR #554, opened 2025-09-08
Onboarding Molmo Model

mohiso22:


QEfficient/transformers/models/molmo/modeling_molmo.py
```diff
@@ -0,0 +1,802 @@
...
+        x: torch.Tensor,
+        attention_bias: Optional[torch.Tensor] = None,
+        position_ids: Optional[torch.Tensor] = None,
+        layer_past: Optional[Cache] = None,
+        batch_index: Optional[torch.LongTensor] = None,
+        use_cache: bool = False,
+        **kwargs,
+    ) -> Tuple[torch.Tensor, Optional[Tuple[torch.Tensor, torch.Tensor]]]:
+        if not self.config.norm_after:
+            if self._activation_checkpoint_fn is not None:
```
quic-amitraj, line 238: Remove all the training related code throughout this file.

quic-amitraj (changes requested):
Remove all the training related code throughout modelling file.

QEfficient/transformers/models/molmo/modeling_molmo.py
```diff
@@ -0,0 +1,776 @@
...
+        img_size: int,
+        kv_offload: bool = False,
+        **compiler_options,
+    ):
+        prefill_seq_len = prefill_seq_len if prefill_seq_len else 1024
+        ctx_len = ctx_len if ctx_len else constants.INTERN_CTX_LEN
+
+        img_size = 588
+        img_tile = 576
+        num_images = 5
```
quic-xiyushi, line 580: Could you make num_images a parameter instead of hardcoding it? The value of num_images here is essentially max_crops (which is one of the image processor kwargs) plus 1. That way, in VLLM, we can apply the same preprocessing logic as we do for llama4

PR #560, opened 2025-09-12
[Qwen2_5_vl] - Onboarding Qwen2_5_vl model in QEfficient

mohiso22:


QEfficient/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py
```diff
@@ -0,0 +1,871 @@
...
+        ctx_len: int,
+        img_size: None,
+        height: int = None,
+        width: int = None,
+        kv_offload: bool = False,
+        **compiler_options,
+    ):
+        if height is None or width is None:
+            height = 1365
+            width = 2048
```
quic-xiyushi, line 719: When I tried compiling the model without explicitly setting the height and width (i.e., using the default values: height=1365 and width=2048), the compilation failed. Could you please try it on your end as well? If it fails, consider updating the default values to ones that work.

examples/qwen2_5_vl_example.py
```diff
@@ -0,0 +1,178 @@
...
+import transformers
+from PIL import Image
+from qwen_vl_utils import process_vision_info
+from transformers import AutoConfig, AutoProcessor, TextStreamer
+
+from QEfficient import QEFFAutoModelForImageTextToText
+
+## For AWQ model update pytorch version to 2.8.*
+model_id = "Qwen/Qwen2.5-VL-32B-Instruct"
+config = AutoConfig.from_pretrained(model_id)
```
quic-rishinr: Do we need to pass the config?

tests/transformers/models/qnn_config.json
```diff
@@ -0,0 +1,10 @@
+{
```
quic-rishinr: Why do we need this file?

quic-rishinr:
can we add the tests for qwen 2.5

quic-hemagnih:
CI is passed

PR #573, opened 2025-09-23
Fix llama model o_proj lora_ids passing for finite lorax in release/v1.20.0

quic-jouachen:
This is regarding the issue reported in https://github.com/quic/efficient-transformers/issues/572

The finite lorax feature failed to execute when testing on a llama adapter ([jumip/llama-lora-adapter](https://huggingface.co/jumip/llama-lora-adapter)) that contains o_proj as target module.

Work in progress -- test to be added to avoid future regression


quic-rishinr:
Hey @quic-jouachen 1.20 branch is frozen. can we close this PR?

PR #579, opened 2025-09-29
[QEff Finetune] Correction in data type of loss

quic-swatia:
In case of DDP, if size of dataset (train dataset or validation  dataset) is smaller than the degree of DDP, then the loss value for padded samples was coming out to be in float. It's handled with this change.


quic-meetkuma (commented):
LGTM, Make a JIRA and reference the gerrit on the JIRA. Post that we are good to go.

quic-akuruvil (approved):
LGTM

quic-meetkuma (approved):
Thanks for spotting the issue and fixing it. :)

PR #592, opened 2025-10-17
Wan 2.2 support

tv-karthikeya:


tv-karthikeya:
will be rebasing, repush

PR #595, opened 2025-10-18
Modified qwen_2.5 modelling file to allow replicate_kv_script to work for custom num_kv_heads.

quic-dhirajku:
Edited the replicate_kv_heads script to allow loading the VLM and export properly after KV_heads update.

## Export goes through successfuly, need to export full model for it to work due to TF version issues.


quic-hemagnih:
Hi @quic-dhirajku  as discussed please raise a ticket for SIT to verify the Perf hit. In case if we are good from perf side and CI then lets go ahead and merge it.

In parallel we can work on below PR
https://github.com/quic/efficient-transformers/pull/625

PR #598, opened 2025-10-24
Qwen2.5_VL Example Script Update

mohiso22:


QEfficient/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py
```diff
@@ -779,10 +779,10 @@ def get_specializations(
         **compiler_options,
     ):
         if height is None or width is None:
-            height = 1365
-            width = 2048
+            height = 354
```
quic-rishinr: Can we keep the constants in the constant file?

quic-hemagnih: Agree, lets put both height and width in the constants file and use macro here.

mohiso22: Done

quic-hemagnih (approved):

tests/transformers/models/test_image_text_to_text_models.py
```diff
@@ -134,6 +134,17 @@
         "Can you describe the image in detail.",
         1,
     ),
+    (
+        "Qwen/Qwen2.5-VL-32B-Instruct",
```
vbaddi: We could test smallest from the qwen2.5? maybe 3B?

PR #603, opened 2025-11-03
New PR for GPTOSS decode-only model

ochougul:


examples/gpt_oss.py
```diff
@@ -0,0 +1,35 @@
...
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# -----------------------------------------------------------------------------
+
+from transformers import AutoTokenizer, TextStreamer
+
+from QEfficient import QEFFAutoModelForCausalLM
+
+model_id = "openai/gpt-oss-20b"
```
quic-rishinr: Can we add a comment mentioning we no longer need to convert the weights to fp32 to avoid the confusion?

QEfficient/utils/generate_inputs.py
```diff
@@ -87,13 +87,17 @@ def prepare_pytorch_inputs(self):
 
         if self.full_batch_size:
             inputs["input_ids"] = input_ids
-            inputs["position_ids"] = torch.arange(input_len).view(1, input_len)
-            inputs["batch_index"] = torch.arange(1).view(-1, 1)
+            inputs["position_ids"] = position_ids
+            inputs["batch_index"] = torch.arange(self.full_batch_size).view(-1, 1)
 
         past_key_values = []
+        sliding_padding_shape = self.padding_shape[:2] + [self.config.sliding_window] + [self.padding_shape[-1]]
```
quic-rishinr: Shouldnt we have check if we config have sliding window parameter here?

QEfficient/utils/generate_inputs.py
```diff
@@ -168,9 +169,15 @@ def prepare_ort_inputs(self):
                 inputs["past_key." + str(i)] = np.zeros((cache_shape), dtype=np.float32)
                 inputs["past_value." + str(i)] = np.zeros((cache_shape), dtype=np.float32)
         else:
+            sliding_padding_shape = self.padding_shape[:2] + [self.config.sliding_window] + [self.padding_shape[-1]]
```
quic-rishinr: Same as above

QEfficient/transformers/quantizers/quantizer_utils.py
```diff
@@ -378,3 +378,71 @@ def repack_zeros(qzeros, bits):
         break
     qzeros = qzeros.T
     return qzeros
+
+
+FP4_VALUES = [
```
quic-rishinr: Can we move this to constants.py?

quic-rishinr (changes requested):

QEfficient/transformers/quantizers/quantizer_utils.py
ochougul: This constant is specific to this file, and won't be used elsewhere at all. So I think its okay for it to be here.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -2641,6 +2654,11 @@ def compile(
                 for kv in ["key", "value"]:
                     custom_io[f"past_{kv}.{i}{suffix}"] = kv_cache_dtype
 
+        # HACK for now
```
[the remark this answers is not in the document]
ochougul: makes sense

PR #610, opened 2025-11-05
Continuous Batching for VLMs

asmigosw:
Adding CB support for VLMs:
1. Llava
2. Llava_Next
3. Gemma3
4. Mistral3
5. InternVL2_5
6. InternVL3_5
7. Molmo

### DIFF
--- a/QEfficient/generation/embedding_handler.py
+++ b/QEfficient/generation/embedding_handler.py
@@ -232,7 +232,9 @@ class VisionHandler:
                 if "mistral3" in self._qeff_model.model.config.model_type:
                     image = image.resize((constants.MISTRAL3_IMAGE_HEIGHT, constants.MISTRAL3_IMAGE_WIDTH))
                 if "llava_next" in self._qeff_model.model.config.model_type:
-                    image = image.resize((constants.GRANITEVISION_IMG_SIZE_HEIGHT, constants.GRANITEVISION_IMG_SIZE_WIDTH))
+                    image = image.resize(
+                        (constants.GRANITEVISION_IMG_SIZE_HEIGHT, constants.GRANITEVISION_IMG_SIZE_WIDTH)
+                    )
 
             # Prepare conversation format
             conversation = [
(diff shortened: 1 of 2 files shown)


QEfficient/generation/embedding_handler.py
```diff
@@ -95,6 +223,9 @@ def prepare_vlm_inputs(self, image_url: str, query: str, prefill_seq_len: int) -
             else:
                 image = Image.open(image_url)
 
+            if "mistral3" in self._qeff_model.model.config.model_type:
```
quic-mamta: same as above. please update the args for this function's docstrings.

tests/transformers/models/test_image_text_to_text_models.py
```diff
@@ -521,13 +625,64 @@ def check_intern_image_text_to_text_pytorch_vs_kv_vs_ort_vs_ai100(
...
     assert (pytorch_hf_tokens == qpc_tokens).all(), "Tokens don't match for pytorch HF output and QPC output"
+
+    # testing for CB models
+    if not kv_offload:  # CB not yet enabled for Single QPC
+        return
+
+    full_batch_size = 4
+    queries = [query] * full_batch_size
+
+    pytorch_hf_tokens = [pytorch_hf_tokens] * 4
```
quic-mamta: is there any reason we are not using run_vlm_hf_model_pytorch_CB here?

QEfficient/generation/embedding_handler.py
```diff
@@ -70,6 +78,126 @@ def is_available(self) -> bool:
...
+        if not self.is_available():
+            raise ValueError("Vision handler not properly initialized. Need both vision_session and processor.")
+
+        try:
+            # Download image
+            if image_url.startswith(("http://", "https://")):
+                image = Image.open(requests.get(image_url, stream=True).raw)
+            else:
+                image = Image.open(image_url)
+            image = image.resize((536, 354))
```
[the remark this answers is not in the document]
asmigosw: No, actually these models currently only support this config, if we change it, it breaks. So, I have made it fixed. Apart from InternVL we do not have that much freedom in any other VLM to have user specific resize.

PR #613, opened 2025-11-09
WIP: Feat: Add ONNX Sub Functions Export Feature

vbaddi:
This PR introduces support for exporting ONNX modules as functions, enabling more efficient model compilation and execution on hardware.

### Changes
- Added new environment variable `QEFF_USE_ONNX_FUNCTIONS` to control ONNX function export behavior
- Integrated ONNX function export capability into the inference pipeline

### Enable ONNX Functions Export
Set the environment variable before running inference:

```bash
export QEFF_USE_ONNX_FUNCTIONS=true
```

### Export and Execute with ONNX Functions
```bash
python -m QEfficient.cloud.infer \
  --model-name gpt2 \
  --num-cores 16 \
  --device-group "[0]" \
  --prompt "My name is" \
  --num-layers 2
```

#### Backward Compatibility
This feature is opt-in and requires explicit environment variable. Existing workflows remain unaffected when the flag is disabled.

### DIFF
--- a/QEfficient/transformers/cache_utils.py
+++ b/QEfficient/transformers/cache_utils.py
@@ -24,6 +24,7 @@ from QEfficient.customop import (
     CtxScatterFuncCB3D,
 )
 
+
 def _get_invalid_idx_value():
     """
     Get the appropriate invalid index value for CtxGather operations.
@@ -46,6 +47,7 @@ def _get_invalid_idx_value():
     else:
         # For runtime: use 0
         return 0
+
 
 class QEffDynamicLayer(DynamicLayer):
     def read_only(self, cache_kwargs):


QEfficient/base/onnx_transforms.py
```diff
@@ -99,3 +102,164 @@ def apply(
...
+        onnx_base_dir: Optional[str] = None,
+        **kwargs,
+    ) -> Tuple[ModelProto, bool]:
+        """
+        :param enable_onnx_slim_transform: If True, applies onnx-slim transformations.
+        :param temp_onnx_path: Path to save the slimmed ONNX model.
+        """
+        transformed = False
+        onnx_slim_transform = True  # kwargs.get("enable_onnx_slim_transform", False)
+        temp_onnx_path = kwargs.get("temp_onnx_path", None)
```
quic-rishinr: Can we make it as a mandiatory argument? and onnx_base_dir is unused here

QEfficient/base/onnx_transforms.py
```diff
@@ -99,3 +102,164 @@ def apply(
...
+        *,
+        onnx_base_dir: Optional[str] = None,
+        **kwargs,
+    ) -> Tuple[ModelProto, bool]:
+        """
+        :param enable_onnx_slim_transform: If True, applies onnx-slim transformations.
+        :param temp_onnx_path: Path to save the slimmed ONNX model.
+        """
+        transformed = False
+        onnx_slim_transform = True  # kwargs.get("enable_onnx_slim_transform", False)
```
quic-rishinr: if OnnxSlimTransform is called do you need to again have a flag for onnx_slim_transform = True? and then check it on line 130? expectation should be to apply the onnxslimtransform right?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -2359,7 +2371,7 @@ def export(self, export_dir: Optional[str] = None) -> str:
                 for kv in ["key", "value"]:
                     example_inputs["past_key_values"][i].append(torch.zeros(pkv_cache[0][0].shape, dtype=torch.float32))
                     dynamic_axes[f"past_{kv}.{i}"] = pkv_dynamic_axes
-                    output_names.append(f"past_{kv}.{i}_RetainedState")
+                    output_names.append(f"past_{kv}.{i}_InternalRetainedState")
```
quic-rishinr: Why we are renaming it? if we are renaming _RetainedState to _InternalRetainedState wouldnt the chages need to added on text_generation_inference and other places we are skipping the bufferes? Even if we are not enabling the subfunction this would impact regular execution

QEfficient/base/modeling_qeff.py
```diff
@@ -252,6 +260,8 @@ def _export(
                 output_names=output_names,
                 dynamic_axes=dynamic_axes,
                 opset_version=constants.ONNX_EXPORT_OPSET,
+                export_modules_as_functions=decoder_layer_classes,
+                do_constant_folding=True,
```
quic-rishinr: Do we need it to be true by default if we are enabling it via env variable?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -2111,7 +2117,13 @@ class QEFFAutoModelForCausalLM(QEFFBaseModel):
         SplitGateUpWeightsTransform,
         KVCacheExternalModuleMapperTransform,
     ]
-    _onnx_transforms = [FP16ClipTransform, SplitTensorsTransform]
+    _onnx_transforms = [
+        FP16ClipTransform,
+        CustomOpTransform,
```
quic-rishinr: do we need to apply the CustomOpTransform again after export?

quic-rishinr (changes requested):

ochougul:
It might be more intuitive and flexible to have a dedicated flag in the export configuration—something like `use_subfunctions` or `export_submodules_as_functions`
Using an environment variable makes it harder to switch between the two approaches dynamically, especially during development or testing. A flag would offer clearer intent and better usability.
We can add this flag to the export API in auto classes

QEfficient/base/onnx_transforms.py
inisis: Can you provide the error log? I think there should be a 5% performance gain with onnxslim.

abhishek-singh591: Not sure about GPT OSS but for Qwen 2.5 VL we observed that onnxslim removes identical nodes which lead to creation of dummy nodes.

QEfficient/transformers/models/modeling_auto.py
abhishek-singh591: @ochougul.

QEfficient/base/onnx_transforms.py
inisis: > Not sure about GPT OSS but for Qwen 2.5 VL we observed that onnxslim removes identical nodes which lead to creation of dummy nodes.

@abhishek-singh591 Hi, what's that dummy nodes, it should not be created by onnxslim, the removes of identical nodes is generally known as CSE, it reduces extra computation, I thinks it's very useful.

QEfficient/utils/patches.py
```diff
@@ -0,0 +1,120 @@
...
+
+def apply_torch_patches():
+    """Apply all necessary torch patches for ONNX export."""
+    # Monkey patch the function
+    onnx_utils._setup_trace_module_map = _setup_trace_module_map_patched
+
+    if hasattr(onnx_utils, "_get_module_attributes"):
+        onnx_utils._get_module_attributes = _get_module_attributes
+
+    print("Applied torch ONNX export patches for export_modules_as_functions compatibility")
```
ochougul: use logger here.

vbaddi:
Continuing in #621.
This will be closed.

PR #620, opened 2025-11-17
[Proxy]: Adding support for exporting proxy Model

abukhoy:
This feature adds support for exporting a proxy model, which disables the Embedding Layer and LM Head of the model.

Set `enable_proxy = True` to export the proxy model.
Set `write_io = True` to save input/output files during the generation stage.

Refer to the example script for implementation details.

## Testing

1. Text Models
2. Embedding Models
3. Vision Models
4. Audio Models

Note: Check the Example Script for the same.


vbaddi:
Thanks @abukhoy 
Can we extend it to VLM also? Assuming Vision can have full model, the `decoder_wrapper` can have `enable_proxy` flag?

examples/proxy/audio_model.py
```diff
@@ -0,0 +1,68 @@
+# -----------------------------------------------------------------------------
```
quic-rishinr: move this to script folder and rename the folder

quic-rishinr (changes requested):

PR #597, opened 2025-10-24
Extend on-device sampling support for dual QPC VLMs

quic-xiyushi:
**Overview**

On-device sampling can significantly reduce host overhead and improve inference throughput; however, so far it has only been implemented for `QEffForCausalLM` models. This PR extends on-device sampling support to the language decoder of dual QPC vision language models, `QEffCausalLMForTextImageToTextModel`. In addition, it fixes the bug in gumbel noise so that it correctly simulates a multinomial distribution for random sampling.

**Implementation details**

```
class _QEffAutoModelForImageTextToTextDualQPC:

def __init__(
        self,
        model: nn.Module,
        continuous_batching: bool = False,
        qaic_config: Optional[dict] = None,
        **kwargs,
    ):
        # Omitting unchanged parts
        self.lang_model = QEffCausalLMForTextImageToTextModel(model, qaic_config=qaic_config, **kwargs)
        # ---Sampling---
        # Note: SamplerTransform should be applied after all other transforms
        # are done. The role of the sampler is to just add nodes at the output of the
        # previous transform function.
        self.lang_model.model, _ = SamplerTransform.apply(self.lang_model.model, qaic_config, **kwargs)
```

**Usage**

The usage is the similar to enable on-device sampling for `QEffForCausalLM`.

```
from QEfficient import QEFFAutoModelForImageTextToText

model_id = "Qwen/Qwen2.5-VL-3B-Instruct"

qeff_model = QEFFAutoModelForImageTextToText.from_pretrained(
    model_id,
    attn_implementation="eager",
    kv_offload=True,
    continuous_batching=True,
    qaic_config={
        "include_sampler": True,
        "return_pdfs": False,
        "max_top_k_ids": 512,
    },
)
```

PR #624, opened 2025-11-19
Add Support for Guided Decoding to On Device Sampling

quic-sanising:
### 📌 Overview

This PR introduces **guided decoding** capabilities in On Device Sampling for `QEffForCausalLM` and `QEffCausalLMForTextImageToTextModel` models.

</br>
</br>

### 🚀 Motivation

As outlined in [this blog on structured decoding](https://blog.vllm.ai/2025/01/14/struct-decode-intro.html), structured decoding represents a fundamental shift in controlling LLM outputs. Instead of relying on post-processing, constraints are enforced during token generation via **logits manipulation**. This approach ensures:

*   **Format compliance** at generation time.
*   Reduced error rates for structured outputs.
*   Performance improvements through optimized backends like **XGrammar**, which can deliver up to **5× faster token generation under load**.

The constraints are provided through `token_bitmasks` which is a Boolean matrix of shape `(batch_size, vocab_size)`. Here, each element indicates whether a token should be kept (1) or masked (0). During sampling, this mask is applied to the logits before token selection, ensuring that only allowed tokens are considered. 

By performing this operation directly on the device, we eliminate host-device transfers, reduce latency, and improve throughput for structured decoding workloads.

</br>
</br>

### 🛠️ Implementation Details

The guided decoding logic is injected via `include_guided_decoding=True` during model loading. No changes to the model architecture are required.

```python
from QEfficient import QEFFAutoModelForCausalLM as AutoModelForCausalLM

# Load model with On Device Sampler enabled
qeff_model = AutoModelForCausalLM.from_pretrained(
    "meta-llama/Llama-3.1-8B",
    continuous_batching=True,
    qaic_config={
        "include_sampler": True,
        "return_pdfs": False,
        "max_top_k_ids": 512,
        "include_guided_decoding": True,
    },
)

# Compile as usual
qeff_model.compile(
    prefill_seq_length=128,
    ctx_len=256,
    full_batch_size=16,
    num_devices=4,
    num_speculative_tokens=0,
    mxint8_kv_cache=True,
    mxfp6_matmul=True,
)
```

To disable guided decoding, simply set `include_guided_decoding=False`.

### DIFF
--- a/scripts/Jenkinsfile
+++ b/scripts/Jenkinsfile
@@ -169,6 +169,7 @@ pipeline {
                     cd /efficient-transformers &&
                     . preflight_qeff/bin/activate &&
                     pip install /opt/qti-aic/integrations/torch_qaic/py310/torch_qaic-0.1.0-cp310-cp310-linux_x86_64.whl &&
+                    pip install torch==2.9.0 torchvision==0.24.0 torchaudio==2.9.0 --index-url https://download.pytorch.org/whl/cpu &&
                     mkdir -p $PWD/cli_qaic_finetuning &&
                     export TOKENIZERS_PARALLELISM=false &&
                     export QEFF_HOME=$PWD/cli_qaic_finetuning &&
--- a/tests/transformers/test_subfunction.py
+++ b/tests/transformers/test_subfunction.py
@@ -4,9 +4,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 # ----------------------------------------------------------------------------
+from collections import Counter
 
-import hashlib
-
+import onnx
 import pytest
 import torch
 from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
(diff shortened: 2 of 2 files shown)


quic-sanising:
Depends on #597

QEfficient/transformers/sampler/sampler.py
```diff
@@ -122,6 +125,9 @@ def sampler_forward(
     top_ps: Optional[torch.Tensor] = None,
     min_ps: Optional[torch.Tensor] = None,
     random_numbers: Optional[torch.Tensor] = None,
+    vision_embeds: Optional[torch.Tensor] = None,
```
quic-mamta: please keep dtype of these 2 consistent as per lines 27-28. also update function docstring for these newly added args.

tests/transformers/sampler/test_sampler.py
```diff
@@ -46,79 +75,162 @@
         20,  # generation_len
         4,  # full_batch_size
         1,  # spec_length
+        False,  # is_vlm
+    ),
+    # pytest.param(
```
quic-mamta: Any reason for this test to be disabled?

tests/transformers/sampler/test_sampler.py
```diff
@@ -132,55 +244,78 @@ def test_sampler_transform(
...
-    model_w_sampler = QEFFAutoModelForCausalLM.from_pretrained(
+    additional_configs = {}
+    additional_params = {}
+    if is_vlm:
+        config = AutoConfig.from_pretrained(model)
+        config.text_config.num_hidden_layers = 4
+        additional_configs["config"] = config
+        additional_configs["kv_offload"] = True
+        qeff_class = QEFFAutoModelForImageTextToText
+        assert isinstance(prompts, tuple)
```
quic-mamta: Please add error message in this case for vlms.

tests/transformers/sampler/test_sampler.py
```diff
@@ -233,48 +370,62 @@ def test_greedy_sampling(
...
+    if is_vlm:
+        additional_configs["kv_offload"] = True
+        qeff_class = QEFFAutoModelForImageTextToText
+        assert isinstance(prompts, tuple)
+        additional_params["images"] = prompts[0]
+        additional_params["processor"] = AutoProcessor.from_pretrained(model)
+        prompts = prompts[1]
+    else:
+        qeff_class = QEFFAutoModelForCausalLM
+        spec_length -= 1
```
quic-mamta: num_hidden_layers should be specified here or does this test require all model layers?

tests/transformers/sampler/test_sampler.py
```diff
@@ -385,3 +541,126 @@ def test_random_sampling(
...
+        },
+        **additional_configs,
+    )
+    model_w_sampler_wo_guided_decoding = qeff_class.from_pretrained(
+        model,
+        continuous_batching=True,
+        qaic_config={
+            "include_sampler": True,
+            "return_pdfs": False,
+            "max_top_k_ids": 1024,
```
quic-mamta: can we define this test parameter in constants and then reuse?

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -711,10 +713,12 @@ class SamplerTransform:
         QEffGPTJForCausalLM,
         QEffGraniteForCausalLM,
         QEffGraniteMoeForCausalLM,
+        QEffInternDecoderWrapper,
         QEffLlamaForCausalLM,
         QEffMptForCausalLM,
         QEffPhi3ForCausalLM,
         QEffQwen2ForCausalLM,
+        QEffQwen_2_5_vl_DecoderWrapper,
```
quic-mamta: Is this supported for Intern and Qwen models only?

quic-xiyushi: Currently, only Intern and Qwen models have been tested. Other models should also be supported if we add their decoder wrapper here.

examples/performance/on_device_sampling.py
```diff
@@ -21,13 +21,16 @@ def main(args, **kwargs):
     include_sampler = None
     return_pdfs = None
     max_top_k_ids = None
+    include_guided_decoding = None
```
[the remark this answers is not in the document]
quic-xiyushi: It’s set to None instead of a boolean because, in lines 46–54, include_guided_decoding will only be added to qaic_config if it is not None; otherwise, the default value defined in QEfficient/utils/sampler_utils.py will be used.

tests/transformers/sampler/test_sampler.py
quic-xiyushi: We want to use different `max_top_k_ids values`, and `include_guided_decoding` may or may not be present. Therefore, even if we define this test parameter in constants, it won’t be very reusable and is not generalized.

QEfficient/transformers/models/pytorch_transforms.py
quic-xiyushi: Added support for Llava, Garnite, Gemma, and Llama 4, and verified that they work correctly. Thanks!

PR #625, opened 2025-11-19
Created ReplicateKVHeadTransform to integrate KV-heads replication module within Qefficient library.

quic-dhirajku:
The Transform enables KV-head replication for CausalLMs and VLMs as well. 
The feature is enabled by passing n_kv_head_repeat parameter during initialization of the QEff wrapper class for the corresponding model. 
`n_kv_head_repeat` param acts as the multiplier for the number of repeats to be done to original count of KV heads. This operation also causes the config and the hash params of the respective model to update the num_key_value_heads parameter and add a paramter orig_kv_heads to it; It allows us to export the same model with different number of kv_heads without causing a hash conflict. 
Added tests for both CausalLMs and VLMs with this functionality to compare outputs of Pytorch HF model and the AIC model. Two new optional paramters `n_kv_head_repeat` and `test_kv_replicate` are added for testing purpose. Setting `test_kv_replicate` to True performs a KV-head replication of every model such that the number of KV-heads and attention heads becomes equal. This was done to ensure tests don't fail due to misalignment issues when we simply repeat num_key_value_heads twice and thus cause a divisibility error on hum_heads.


tests/transformers/models/image_text_to_text/test_image_text_to_text_models.py
```diff
@@ -428,6 +443,8 @@ def check_intern_image_text_to_text_pytorch_vs_kv_vs_ort_vs_ai100(
     num_devices: int = 1,
     enable_qnn: Optional[bool] = False,
     qnn_config: Optional[str] = None,
+    n_kv_head_repeat: Optional[int] = None,
```
vbaddi: can we rename this to "num_kv_heads_replicate" instead?

vbaddi (changes requested):

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -2384,6 +2395,9 @@ def __init__(
 
         setattr(model.config, "max_seq_len_cached", max_seq_len_cached)
         super().__init__(model, qaic_config=qaic_config, **kwargs)
+        self.model, replicate_kv_transformed = ReplicateKVHeadTransform.apply(self.model, **kwargs)
+        if replicate_kv_transformed:
+            self.hash_params["config"] = model.config.to_diff_dict()
```
ochougul: better add it to `_pytorch_transforms` if we are always going to call it.

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -686,6 +691,164 @@ class RevertPrefillOnlyTransform(ModuleMappingTransform):
     }
 
 
+class ReplicateKVHeadTransform:
```
ochougul: Make this inherit `ModuleMutatorTransform` 
You may need to implement mutate method which is similar to apply here

ochougul (changes requested):
Write a test that makes sure onnx hash is different when different number of kv heads are passed.

quic-rishinr:
closing this PR changes are part of #997

PR #626, opened 2025-11-19
[QEff. Finetune]: Added initial folder structure and files for HF trainer based FT code.

quic-meetkuma:
- The code is based on HF Trainer infrastructure. We have added a skeleton code. Upon merger of this, different team member will work parallelly on different components.
- The code will be an experimental and will sit along side the existing finetuning code. Once we achieve parity and stability with the HF Trainer based code, we will deprecate the current finetuning code and adopt the HF Trainer–based code as mainstream.


quic-akuruvil:
Looks good to merge. The skeleton folder structure looks solid for the initial version.

PR #636, opened 2025-11-25
[QEff Finetune]: Made a separate list of dependencies for Finetuning.

quic-meetkuma:
- With this change the FT dependencies should be explicitly installed via "pip install -e .[ft]".
- Added this so that eager stack can have different pytorch and transformers dependencies than the AOT stack.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,8 +51,6 @@ infer = [
     "torch@https://download.pytorch.org/whl/cpu/torch-2.7.0%2Bcpu-cp310-cp310-manylinux_2_28_x86_64.whl ; python_version=='3.10' and platform_machine=='x86_64'",
 ]
 ft = [
-    "torch_qaic @ file:///opt/qti-aic/integrations/torch_qaic/py310/torch_qaic-0.1.0-cp310-cp310-linux_x86_64.whl ; python_version=='3.10' and platform_machine=='x86_64'",
-    "torch_qaic @ file:///opt/qti-aic/integrations/torch_qaic/py311/torch_qaic-0.1.0-cp310-cp310-linux_x86_64.whl ; python_version=='3.11' and platform_machine=='x86_64'",
     "accelerate @ file:///opt/qti-aic/integrations/accelerate/py310/accelerate-1.10.1-py3-none-any.whl ; python_version=='3.10' and platform_machine=='x86_64'",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -51,6 +49,14 @@ test = ["pytest","pytest-mock"]
 docs = ["Sphinx==7.1.2","sphinx-rtd-theme==2.0.0","myst-parser==3.0.1","sphinx-multiversion"]
 quality = ["black", "ruff", "hf_doc_builder@git+https://github.com/huggingface/doc-builder.git"]
 
+# New FT dependencies
+ft = [
```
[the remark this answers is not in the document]
quic-akuruvil: That would require explicit installation again by the user right. If it can be included here it would be good. When we do `pip install .[ft] ` all the dependencies required for FT are installed in one go.

pyproject.toml
```diff
@@ -35,21 +35,28 @@ dependencies = [
...
+infer = [
+    "torch==2.7.0; platform_machine=='aarch64'",
+    # Specifying torch cpu package URL per python version, update the list once pytorch releases whl for python>3.11
+    "torch@https://download.pytorch.org/whl/cpu/torch-2.4.1%2Bcpu-cp38-cp38-linux_x86_64.whl ; python_version=='3.8' and platform_machine=='x86_64'",
+    "torch@https://download.pytorch.org/whl/cpu/torch-2.7.0%2Bcpu-cp39-cp39-manylinux_2_28_x86_64.whl ; python_version=='3.9' and platform_machine=='x86_64'",
+    "torch@https://download.pytorch.org/whl/cpu/torch-2.7.0%2Bcpu-cp310-cp310-manylinux_2_28_x86_64.whl ; python_version=='3.10' and platform_machine=='x86_64'",
+]
+ft = [
+    "tensorboard ; python_version>='3.10' and python_version<'3.12' and platform_machine=='x86_64'",
+    "transformers==4.55.0 ; python_version>='3.10' and python_version<'3.12' and platform_machine=='x86_64'",
```
quic-swatia: Dependencies already has the same version of transformers at line number 22 of this file. Why is this needed here?

quic-meetkuma: Eager support python 3.10 and 3.12.
Explicitly adding transformers here to simplify future modifications to the FT’s transformer.

PR #640, opened 2025-11-26
Added memory and time optimization for onnx transforms

abhishek-singh591:
### Memory Optimization

Added periodic memory cleanup to FP16ClipTransform and SplitTensorsTransform to reduce memory usage during large tensor processing. Also avoids redundant external data loading when already present.

### Time Optimized ONNX Transform via Class Merging and Thread Pooling

It merges the FP16 and Split ONNX transform classes into a single implementation to eliminate redundant tensor loading and iteration. Additionally, the transform logic has been refactored to use a **thread pool**, replacing the previous sequential loop to parallelize tensor operations.

####  Performance Benchmarks:-

| Model           | Original Duration (s) | Optimized Duration (s) |
|----------------|------------------------|-------------------------|
| LLaMA 3.1 8B    | 88.35                  | 58.55                   |
| LLaMA 3.1 70B   | 1029.82                | 727.37                  |

> **Note:** Thread count is set to `os.cpu_count() * 4` to better handle I/O-bound workloads. Performance may vary depending on system hardware and threading capabilities.

### DIFF
--- a/tests/base/test_onnx_transforms.py
+++ b/tests/base/test_onnx_transforms.py
@@ -102,7 +102,6 @@ def test_split_tensors_transform(tmp_path):
     onnx_transforms = OnnxTransformPipeline(transforms=[SplitTensorsTransform])
     trans_onnx, transformed = onnx_transforms.apply(
         test_onnx,
-        test_onnx,
         model_name="test_split",
         onnx_base_dir=str(tmp_path),
         file_chunk_size=32 * 4,


QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3208,6 +3203,21 @@ def __init__(self, model: nn.Module, **kwargs):
...
+        Returns
+        -------
+        str
+            The model's class name, with "QEff" or "QEFF" prefix removed if present.
+        """
+        mname = self.model.__class__.__name__
+        if mname.startswith("QEff") or mname.startswith("QEFF"):
+            mname = mname[4:]
+        return mname
+
```
ochougul: why is this added here?

scripts/memory_profiling/__init__.py
ochougul: remove, this was just for our experiments

QEfficient/base/onnx_transforms.py
```diff
@@ -142,80 +98,173 @@ class CustomOpTransform(OnnxTransform):
...
+                removed_initializers.append(i)
+
+        if transformed:
+            for i in sorted(removed_initializers, reverse=True):
+                model.graph.initializer.pop(i)
+
         return model, transformed
+
+
+class OnnxTransformPipeline(BaseOnnxTransform):
```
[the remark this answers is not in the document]
abhishek-singh591: done.

PR #652, opened 2025-12-03
[QEff. Finetune] Adding callback and its test cases.

tchawada:
Adding a Script for Registering and Retrieving Callback Classes
It has create_callback() function which creates an instance of callback.
Additionally, there is a test_callbacks.py script that validates the functionality and retrieval process.

### DIFF
--- a/QEfficient/finetune/experimental/tests/test_callback.py
+++ b/QEfficient/finetune/experimental/tests/test_callback.py
@@ -50,11 +50,8 @@ def test_callbacks(callback_name):
     except ValueError as e:
         assert "Unknown callback" in str(e)
         return
-    if hasattr(callback_inst, "callback"):
-        assert callback_inst.callback is not None
-    else:
-        assert callback_inst is not None
-        assert isinstance(callback_inst, TrainerCallback)
+    assert callback_inst is not None
+    assert isinstance(callback_inst, TrainerCallback)
 
 
 @pytest.mark.parametrize("callback_name,callback_class", REGISTRY_CALLBACK_CONFIGS.items())


QEfficient/finetune/experimental/core/callbacks.py
```diff
@@ -4,3 +4,185 @@
...
+                if isinstance(v, str) and len(v) > self.max_str_len:
+                    shallow_logs[k] = (
+                        f"[String too long to display, length: {len(v)} > {self.max_str_len}. "
+                        "Consider increasing `max_str_len` if needed.]"
+                    )
+                else:
+                    shallow_logs[k] = v
+            _ = shallow_logs.pop("total_flos", None)
+            # round numbers so that it looks better in console
+            if "epoch" in shallow_logs:
```
quic-meetkuma: FYI:https://github.com/huggingface/transformers/blob/f8e69286fb4b7f4cc778514e1c83c8a1f579f328/src/transformers/trainer_callback.py#L657

Minor comment can be added like this: Overriding existing on_log function in order to print epoch, loss and lr

QEfficient/finetune/experimental/tests/test_callback.py
```diff
@@ -0,0 +1,79 @@
...
+# -----------------------------------------------------------------------------
+
+import pytest
+from transformers import TrainerCallback
+
+from QEfficient.finetune.experimental.core.callbacks import create_callbacks
+from QEfficient.finetune.experimental.core.component_registry import registry
+
+
+class ModelSummaryCallback(TrainerCallback):
```
quic-meetkuma: This is a good implementation. May be we can use it as other defined callbacks and put it along those. For now it is fine. 

Also, you don't any implementation as this is just class registry.

quic-meetkuma (changes requested):
Looks good. Minor corrections. Post that let us merge it.

PR #658, opened 2025-12-09
[QEff. Finetune]: Adding base class and HF class

quic-swatia:
-  Added Base Model class and HF model class.
- Base Model class will support FT for any custom model and will be a common skeleton for any model, including HF. 
- Added unit tests for these.

### DIFF
--- a/QEfficient/finetune/experimental/core/model.py
+++ b/QEfficient/finetune/experimental/core/model.py
@@ -5,19 +5,19 @@
 #
 # -----------------------------------------------------------------------------
 
-import logging
+import warnings
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Optional, Type
 
 import torch.nn as nn
-from transformers import AutoTokenizer
 import transformers
-from transformers.utils.logging import get_logger
+from transformers import AutoTokenizer
 
 from QEfficient.finetune.experimental.core.component_registry import registry
+from QEfficient.finetune.experimental.core.logger import Logger
 from QEfficient.finetune.experimental.core.utils.dataset_utils import insert_pad_token
 
-logger = get_logger(__name__)
+logger = Logger(__name__)
 
 
 class BaseModel(nn.Module, ABC):
@@ -48,7 +48,8 @@ class BaseModel(nn.Module, ABC):
 
     def load_tokenizer(self) -> Any:
         """Override if the model exposes a tokenizer."""
-        raise NotImplementedError(f"{type(self).__name__} does not provide a tokenizer.")
+        warnings.warn(f"{type(self).__name__} does not provide a tokenizer.", category=UserWarning)
+        return None
 
     # Lazy accessors
     @property
@@ -67,19 +68,6 @@ class BaseModel(nn.Module, ABC):
     def forward(self, *args, **kwargs):
         return self.model(*args, **kwargs)
 
-    def get_input_embeddings(self):
-        if hasattr(self.model, "get_input_embeddings"):
-            return self.model.get_input_embeddings()
-        logger.info(f"Model {self.model_name} does not expose input embeddings", logging.WARNING)
-        return None
-
-    def resize_token_embeddings(self, new_num_tokens: int) -> None:
-        if hasattr(self.model, "resize_token_embeddings"):
-            self.model.resize_token_embeddings(new_num_tokens)
-        else:
-            logger.info(f"Model {self.model_name} cannot resize token embeddings", logging.WARNING)
-
-    # optional
     def to(self, *args, **kwargs):
         self.model.to(*args, **kwargs)
         return self
@@ -129,12 +117,13 @@ class HFModel(BaseModel):
 
     def configure_model_kwargs(self) -> Dict[str, Any]:
         """Hook for subclasses to tweak HF `.from_pretrained` kwargs."""
+
         extra = dict(self.model_kwargs)
         # extra["quantization_config"] = self._build_quant_config()
         return extra
 
     def load_model(self) -> nn.Module:
-        logger.info(f"Loading HuggingFace model '{self.model_name}' via {self.auto_class.__name__}")
+        logger.log_rank_zero(f"Loading HuggingFace model '{self.model_name}' via {self.auto_class.__name__}")
 
         return self.auto_class.from_pretrained(
             self.model_name,
@@ -143,7 +132,7 @@ class HFModel(BaseModel):
 
     def load_tokenizer(self) -> AutoTokenizer:
         """Load Hugging Face tokenizer."""
-        logger.info(f"Loading tokenizer '{self.tokenizer_name}'")
+        logger.log_rank_zero(f"Loading tokenizer '{self.tokenizer_name}'")
         tokenizer = AutoTokenizer.from_pretrained(self.tokenizer_name)
         insert_pad_token(tokenizer)
         return tokenizer
--- a/QEfficient/finetune/experimental/tests/test_model.py
+++ b/QEfficient/finetune/experimental/tests/test_model.py
@@ -5,16 +5,15 @@
 #
 # -----------------------------------------------------------------------------
 
+from unittest import mock
+
 import pytest
 import torch
 import torch.nn as nn
-from unittest import mock
 
-import transformers
 from QEfficient.finetune.experimental.core import model
-from QEfficient.finetune.experimental.core.model import BaseModel, HFModel
-from QEfficient.finetune.experimental.core.component_registry import registry
-from QEfficient.finetune.experimental.core.component_registry import ComponentFactory
+from QEfficient.finetune.experimental.core.component_registry import ComponentFactory, registry
+from QEfficient.finetune.experimental.core.model import BaseModel
 
 
 class TestMockModel(nn.Module):
@@ -67,8 +66,9 @@ def test_tokenizer_lazy_loading():
 def test_to_moves_inner_and_returns_self():
     m = ComponentFactory.create_model("testcustom", "dummy")
     with mock.patch.object(TestMockModel, "to", autospec=True) as mocked_to:
-        ret = m.to("cuda:0")
-    mocked_to.assert_called_once_with(m.model, "cuda:0")
+        ret = m.to("cpu:0")
+    assert mocked_to.call_args[0][0] is m.model
+    assert mocked_to.call_args[0][1] == "cpu:0"
     assert ret is m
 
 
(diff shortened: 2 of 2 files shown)


QEfficient/finetune/experimental/core/model.py
```diff
@@ -4,3 +4,145 @@
...
+            raise TypeError(f"load_model() must return nn.Module, got {type(module)}")
+        obj._model = module
+        return obj
+
+    @abstractmethod
+    def load_model(self) -> nn.Module:
+        """Create and return the underlying torch.nn.Module."""
+        ...
+
+    def load_tokenizer(self) -> Any:
```
quic-meetkuma: Specific to LLMs. We can not put it here.

QEfficient/finetune/experimental/core/model.py
```diff
@@ -4,3 +4,145 @@
...
+        module = obj.load_model()
+        if not isinstance(module, nn.Module):
+            raise TypeError(f"load_model() must return nn.Module, got {type(module)}")
+        obj._model = module
+        return obj
+
+    @abstractmethod
+    def load_model(self) -> nn.Module:
+        """Create and return the underlying torch.nn.Module."""
+        ...
```
quic-meetkuma: use "pass" as it is explicit.

QEfficient/finetune/experimental/core/model.py
```diff
@@ -4,3 +4,145 @@
...
+import transformers
+from transformers.utils.logging import get_logger
+
+from QEfficient.finetune.experimental.core.component_registry import registry
+from QEfficient.finetune.experimental.core.utils.dataset_utils import insert_pad_token
+
+logger = get_logger(__name__)
+
+
+class BaseModel(nn.Module, ABC):
```
quic-meetkuma: Is there any reason why below methods are not implemented? It was given in a reference code as well.

- parameters
- named_parameters
- state_dict
- load_state_dict

Reference code: https://github.com/quic-meetkuma/LightningLLMs/blob/hf_trainer/LightningLLM/components/model.py

QEfficient/finetune/experimental/core/model.py
```diff
@@ -4,3 +4,145 @@
...
+    def train(self, mode: bool = True):
+        self.model.train(mode)
+        return super().train(mode)
+
+    def eval(self):
+        return self.train(False)
+
+
+@registry.model("hf")
+class HFModel(BaseModel):
```
quic-meetkuma: How can I get the PEFT config file from this model?

Any reason to avoid adding that method?

Reference code had that functionality for some reason: https://github.com/quic-meetkuma/LightningLLMs/blob/hf_trainer/LightningLLM/components/model.py#L272

QEfficient/finetune/experimental/tests/test_model.py
```diff
@@ -0,0 +1,143 @@
+import pytest
```
quic-meetkuma: Copy write header missing

QEfficient/finetune/experimental/tests/test_model.py
```diff
@@ -0,0 +1,143 @@
...
+    m.train()
+    assert m.training is True
+    assert m.model.training is True
+
+
+def test_resize_token_embeddings_and_get_input_embeddings_warn(monkeypatch):
+    m = TestCustomModel.create("dummy")
+
+    # resize_token_embeddings: underlying model lacks the method, should warn and not raise
+    with mock.patch("QEfficient.finetune.experimental.core.model.logger.info") as mocked_log:
```
quic-meetkuma: logger.info should be changed to logger.log_rank_zero based on earlier comments.

QEfficient/finetune/experimental/tests/test_model.py
```diff
@@ -0,0 +1,143 @@
...
+        "AutoTokenizer",
+        mock.Mock(from_pretrained=mock.Mock(return_value=fake_tok)),
+    )
+    monkeypatch.setattr(
+        "QEfficient.finetune.experimental.core.model.insert_pad_token",
+        mock.Mock(),
+        raising=False,
+    )
+
+    m = HFModel.create("hf-name")
```
quic-meetkuma: No need to call individual class's create method. There is a reason to use a registry functionality.

QEfficient/finetune/experimental/tests/test_model.py
```diff
@@ -0,0 +1,143 @@
...
+    assert isinstance(m.model, FakeAuto)
+
+    # load tokenizer
+    tok = m.load_tokenizer()
+
+    # tokenizer was loaded and pad token inserted
+    model.AutoTokenizer.from_pretrained.assert_called_once_with("hf-name")
```
quic-meetkuma: Dont write such tests filled with assert_called_once etc.. We have written such tests in past but that was not an appropriate thing. It was a makeshift arrangement because of monolith structure of code. Write extensive and proper tests. If a function has made some changes to the model's structure then use that to test rather then counting how many times the function gets called.

QEfficient/finetune/experimental/core/model.py
quic-swatia: IMO the reference code is wrong and misses some careful observations about pytorch nn.Module and how it works. nn.Module already provides these and because the wrapped model is registered, so state_dict / parameters work as-is. Adding trivial passthroughs would be redundant, is noisy and a future maintenance hazard. Let me know if you have difficulty understanding this, I can explain offline.

QEfficient/finetune/experimental/tests/test_model.py
```diff
@@ -0,0 +1,143 @@
...
+        super().__init__(model_name)
+        print("init of custom class")
+
+    def load_model(self) -> nn.Module:
+        return TestMockModel()
+
+    def load_tokenizer(self):
+        return "dummy-tokenizer"
+
+
```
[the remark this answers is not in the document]
quic-swatia: There is no PEFT code, hence no test cases.

Replied above for PEFT code.

QEfficient/finetune/experimental/tests/test_model.py
```diff
@@ -0,0 +1,143 @@
...
+    out = m(torch.zeros(1, 2))
+    assert out.shape == (1, 2)
+
+
+def test_tokenizer_lazy_loading():
+    m = TestCustomModel.create("dummy")
+    assert m._tokenizer is None
+    tok = m.tokenizer
+    assert tok == "dummy-tokenizer"
+    assert m._tokenizer == tok
```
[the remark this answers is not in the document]
quic-swatia: If you look carefully they are different. tokenizer is a property and _tokenizer is a private variable

QEfficient/finetune/experimental/core/model.py
```diff
@@ -4,3 +4,145 @@
...
+    def __init__(self, model_name: str, **model_kwargs: Any) -> None:
+        super().__init__()
+        self.model_name = model_name
+        self.model_kwargs: Dict[str, Any] = model_kwargs
+        self._model: Optional[nn.Module] = None
+        self._tokenizer: Any = None  # HF tokenizers are not nn.Modules.
+
+    # Factory constructor: load model after __init__ finishes
+    @classmethod
+    def create(cls, model_name: str, **model_kwargs: Any) -> "BaseModel":
```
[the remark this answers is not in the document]
quic-meetkuma: I understand your concern. Basically, you want to make sure there will be one liner initialization of model. The reference code does it in 2 lines : https://github.com/quic-meetkuma/LightningLLMs/blob/da2f6b39e8533cbd05d563ca68113828f783e73e/LightningLLM/main.py#L131C34-L131C46

I suggest better to move this create into the ComponentRegistry class and create a "create_model" method over there. The reason being that class has a sole responsibility of storing references of particular classes and creating instances of classes which the user want. This way all the registry and instance creation is encapsulated in one place and we only need to deal with ComponentRegistry and not other individual classes.

QEfficient/finetune/experimental/tests/test_model.py
quic-meetkuma: Move it to ComponentFactory and then instantiate from there.

QEfficient/finetune/experimental/core/model.py
```diff
@@ -4,3 +4,145 @@
...
+
+class BaseModel(nn.Module, ABC):
+    """Shared skeleton for every finetunable model in the system."""
+
+    def __init__(self, model_name: str, **model_kwargs: Any) -> None:
+        super().__init__()
+        self.model_name = model_name
+        self.model_kwargs: Dict[str, Any] = model_kwargs
+        self._model: Optional[nn.Module] = None
+        self._tokenizer: Any = None  # HF tokenizers are not nn.Modules.
```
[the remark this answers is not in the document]
quic-meetkuma: Let us use some generic name e.g. _preprocessor (even in HFModel class)?

QEfficient/finetune/experimental/core/model.py
```diff
@@ -4,3 +4,145 @@
...
+    def forward(self, *args, **kwargs):
+        return self.model(*args, **kwargs)
+
+    def get_input_embeddings(self):
+        if hasattr(self.model, "get_input_embeddings"):
+            return self.model.get_input_embeddings()
+        logger.info(f"Model {self.model_name} does not expose input embeddings", logging.WARNING)
+        return None
+
+    def resize_token_embeddings(self, new_num_tokens: int) -> None:
```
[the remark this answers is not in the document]
quic-meetkuma: This method looks specific to NLP. I dont find reference of this method in VIT or Whisper model of HF. I think the else case is taking care of it. So it is fine for now.

QEfficient/finetune/experimental/core/model.py
```diff
@@ -4,3 +4,145 @@
...
+    def tokenizer(self) -> Any:
+        if self._tokenizer is None:
+            self._tokenizer = self.load_tokenizer()
+        return self._tokenizer
+
+    # nn.Module API surface
+    def forward(self, *args, **kwargs):
+        return self.model(*args, **kwargs)
+
+    def get_input_embeddings(self):
```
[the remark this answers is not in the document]
quic-swatia: This method is also accessible via PreTrainedClass. method implementation here is not helping. 

What was the thought process behind adding this one in the reference code? I also think we should remove this.

QEfficient/finetune/experimental/core/model.py
quic-swatia: Updated in latest, doesn't make any difference though. Both syntax are valid.

QEfficient/finetune/experimental/core/model.py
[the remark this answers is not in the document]
quic-swatia: I meant this method also* (just like resize_token_embeddings) is accessible via PreTrainedModel class. Check reply for resize_token_embeddings one.

QEfficient/finetune/experimental/core/model.py
quic-swatia: PEFT config should be added in the Trainer classes as it is part of training config. trl repo has also added it in the trainer classes. We should not deviate from that. It will make code reuse difficult. Refer: https://github.com/huggingface/trl/blob/main/trl/trainer/sft_trainer.py

Hence, it should be added in class BaseTrainer in PR #660. 

Also, in your reference code, you are assuming that if peft_config  is provided, it is for LoRA. Why is that done? 

 We should keep the flexibility to accept other PEFT techniques configuration as well.

QEfficient/finetune/experimental/core/model.py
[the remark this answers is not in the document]
quic-swatia: Going ahead with existing approach  and not adding load_model in the init as the  current approach follows the HF approach of factory method from_pretrained() of class PreTrainedModel.

PR #663, opened 2025-12-10
Add automatic CCL list generation for prefill and decode when user does not provide lists

vjanfaza:
For using CCL feature during prefilling and decoding, the user needs to pass two lists of context lengths to be used during these processes. Here, we add an option to generate these lists automatically when they are not provided by the user manually. These list generation is suitable for a general-purpose application which considers both prefilling and decoding processes and generates CCL lists for both.

### DIFF
--- a/QEfficient/utils/check_ccl_specializations.py
+++ b/QEfficient/utils/check_ccl_specializations.py
@@ -156,7 +156,7 @@ def process_ccl_specializations(ccl_prefill, ccl_decode, ctx_len, prefill_seq_le
                     if val >= 0:
                         ccl_prefill.append(val)
                 ccl_prefill.sort()
-            
+
     logger.info("CCL Configuration:")
     logger.info(f"  - Prefill context lengths: {ccl_prefill}")
     logger.info(f"  - Decode context lengths: {ccl_decode}")
--- a/QEfficient/generation/text_generation_inference.py
+++ b/QEfficient/generation/text_generation_inference.py
@@ -819,7 +819,9 @@ class QEffTextGenerationBase:
                 inputs["lora_ids"] = np.array(batch_lora_ids, dtype=np.int64).reshape(self.batch_size, 1)
 
         if self.comp_ctx_lengths_prefill is not None:
-            self.list_of_comp_ctx_lengths_prefill = [np.zeros(length, dtype=np.int8) for length in self.comp_ctx_lengths_prefill]
+            self.list_of_comp_ctx_lengths_prefill = [
+                np.zeros(length, dtype=np.int8) for length in self.comp_ctx_lengths_prefill
+            ]
             prefill_ccl_id = 0
             inputs["comp_ctx_lengths"] = self.list_of_comp_ctx_lengths_prefill[prefill_ccl_id]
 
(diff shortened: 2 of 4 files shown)


QEfficient/utils/check_ccl_specializations.py
```diff
@@ -5,40 +5,221 @@
...
+def build_doubling_set(start: int, limit: int, max_elements: int) -> Set[int]:
+    """
+    Build a STRICT doubling set: {start, start*2, start*4, ...} up to 'limit',
+    collecting at most 'max_elements' values. Returns a set; caller will sort.
+    """
+    values: Set[int] = set()
+    if max_elements <= 0 or start <= 0 or limit <= 0:
+        return values
+
+    v = start
```
quic-rishinr: use proper variable naming

QEfficient/utils/check_ccl_specializations.py
```diff
@@ -5,40 +5,221 @@
...
+        # Passthrough if either provided
+        if comp_ctx_lengths_decode is not None or comp_ctx_lengths_prefill is not None:
+            return (
+                comp_ctx_lengths_prefill if comp_ctx_lengths_prefill is not None else [],
+                comp_ctx_lengths_decode if comp_ctx_lengths_decode is not None else [],
+                mapped_cl,
+            )
+
+        # Due to limitations in the number of specializations during compilation, we set the maximum number of elements in comp_ctx_lengths_decode and comp_ctx_lengths_prefill lists to 5.
+        max_elems = 5
```
quic-rishinr: Add it as constant in constant.py

quic-rishinr (changes requested):

QEfficient/utils/check_ccl_specializations.py
```diff
@@ -5,40 +5,190 @@
...
+    if ccl_prefill is None and ccl_decode is None:
+        # Generate optimized context length lists for prefill and decode based on ctx_len
+        # Due to compiler limitations, ccl_prefill and ccl_decode must have distinct values
+        ccl_prefill, ccl_decode, ctx_len = automatic_ccl_generation(ctx_len, prefill_seq_len, ccl_prefill, ccl_decode)
+    else:
+        if prefill_seq_len == 1:
+            if ccl_prefill is not None and ccl_decode is not None:
+                # both prefill and decode ccl can share the same specializations since prefill_seq_len=1. So, a sorted union of both lists can be used for both of them.
+                ccl_union_all = sorted(set(ccl_prefill + ccl_decode))
+                ccl_union_all = [min(x, ctx_len) for x in ccl_union_all]
```
quic-rishinr: this would create multiple ctx_len value in the list if ctx len is less than values passed in ccl_prefill or ccl_decode. is it ok?
for example 
`ccl_prefill = [5000, 8000, 10000]
ccl_decode = [6000, 10000, 12000]
ctx_len = 8192
output would be [5000, 6000, 8000, 8192, 8192]`
maybe we can use if we do not want to create duplicate 
`ccl_prefill=  ccl_decode  =sorted({min(x, ctx_len) for x in (ccl_prefill + ccl_decode)})`

QEfficient/utils/check_ccl_specializations.py
```diff
@@ -5,40 +5,190 @@
...
+                    while val in ccl_decode or val in ccl_prefill:
+                        val -= 1
+                        if val < 0:
+                            break  # Prevent negative values
+                    if val >= 0:
+                        ccl_prefill.append(val)
+
+                # Step 4: Sort both lists
+                ccl_prefill.sort()
+                ccl_decode.sort()
```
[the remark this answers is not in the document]
vjanfaza: Yes, since we don't know how the user passed the list

QEfficient/utils/check_ccl_specializations.py
```diff
@@ -5,40 +5,171 @@
...
+    # Handle non-positive CL
+    if ctx_len <= 0:
+        mapped_cl = next_multiple_of_1024(1)
+        seq = [mapped_cl]
+        return seq, seq, mapped_cl
+
+    mapped_cl = next_multiple_of_1024(ctx_len)
+
+    # Early small-ctx_len case for identical lists
+    if mapped_cl <= 4096:
```
quic-rishinr: Can we move 4096 also to constants.py? update the same in line 113 as well. Sorry missed to post this comment last time

quic-rishinr (approved):

PR #675, opened 2025-12-19
Logger Module For Efficient Transformers

abhishek-singh591:
This PR is to add support for Logger Module in Efficient Transformer library. QEFFLogger will run in a low priority thread and shall have a thread safe queue waiting for logs coming from different modules running in single or multi threaded enviornment. This shall support capturing logs classified at various log levels and user can set different log levels for different namespace so that user can have control of logging and thereby easing the debugging. These logs are dumped in the JSON format so that post processing can be done easily.


abhishek-singh591:
Continuing on PR #696 hence closing this.

PR #682, opened 2025-12-22
Adding dissagg mode support to Qwen3Moe

qcdipankar:
#672 

**Adding disagg support to Qwen3Moe** 

> Config used

PL =128

CL=128*3

image

### DIFF
--- a/QEfficient/transformers/models/qwen3_moe/modeling_qwen3_moe.py
+++ b/QEfficient/transformers/models/qwen3_moe/modeling_qwen3_moe.py
@@ -142,7 +142,7 @@ class QEffPrefillChunkedQwen3MoeSparseMoeBlock(Qwen3MoeSparseMoeBlock):
             gate = x @ W_g  # [T, I]
             up = x @ W_u  # [T, I]
             down = (up * self.experts[e].act_fn(gate)) @ W_d  # [T, H]
-            masked_down = torch.where(routing_weight > 0, down * routing_weight, torch.zeros_like(expert_out))
+            masked_down = down * routing_weight
             expert_out += masked_down
         return expert_out.view(B, S, H), router_logits
 


tests/transformers/models/test_disagg_mode.py
```diff
@@ -30,8 +33,9 @@
 prompts = [prompt1, prompt2]
 
 
+@pytest.mark.skip(reason="only works for gpt-oss right now")
```
ochougul: remove this skip as well

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -2602,7 +2603,11 @@ def export(
                 if self.model.config.model_type in SPECIALIZED_PREFILL_ONLY_MODEL_ARCH
                 else seq_len
             )
-            kv_cache_shape[2] = seq_len + self.model.config.sliding_window if enable_chunking else seq_len
+            kv_cache_shape[2] = (
+                seq_len + (self.model.config.sliding_window if self.model.config.sliding_window is not None else 0)
```
ochougul: better to use 
```
seq_len + getattr(self.model.config, "sliding_window", 0)
```

examples/qwen3moe_disagg_mode_with_chunking.py
ochougul: New file is not needed
Just add a comment in the gpt_oss file.

ochougul (changes requested):

QEfficient/transformers/models/modeling_auto.py
qcdipankar: This is a better to use I agree but it will fail in case of models where the config has the parameter sliding_window but it is set to none.

QEfficient/transformers/models/qwen3_moe/modeling_qwen3_moe.py
```diff
@@ -118,53 +117,50 @@ def eager_attention_forward(
...
-            gate = W_g(x)  # [T, I]
-            up = W_u(x)  # [T, I]
-            down = W_d(up * self.experts[e].act_fn(gate))  # [T, H]
-
+            W_g, W_u = self.experts[e].gate_proj.weight.T, self.experts[e].up_proj.weight.T  # [H, I], [H, I]
+            W_d = self.experts[e].down_proj.weight.T  # [I, H]
+            gate = x @ W_g  # [T, I]
+            up = x @ W_u  # [T, I]
+            down = (up * self.experts[e].act_fn(gate)) @ W_d  # [T, H]
             masked_down = torch.where(routing_weight > 0, down * routing_weight, torch.zeros_like(expert_out))
```
ochougul: This torch.where condition is not required IMO
Can you check how much is perf different with and without this?

ochougul (changes requested):

ochougul (approved):

PR #694, opened 2025-12-24
[QEff. Finetune]: Ft integrated tests

quic-dhirajku:
Added test script to perform end to end Finetuning tests for SFT dataset. Need to add changes to the repo for Seq completion task as well. Current run uses CPU to perform finetuning.

### DIFF
--- a/QEfficient/finetune/experimental/tests/test_integrated.py
+++ b/QEfficient/finetune/experimental/tests/test_integrated.py
@@ -368,16 +368,16 @@ def run_inference_causal_lm(model, tokenizer):
         model: Model instance
         tokenizer: Tokenizer instance
     """
-    test_prompt = "Test prompt for generation"
-    inputs = tokenizer(test_prompt, return_tensors="pt")
-
-    with torch.no_grad():
+    test_prompt = "Test prompt for generation."
+    texts = tokenizer(test_prompt, return_tensors="pt")
+    texts = texts.to(model.device)
+    with torch.inference_mode():
         outputs = model.generate(
-            **inputs,
+            **texts,
+            temperature=0.4,
             max_new_tokens=10,
             do_sample=False,
         )
-
     generated_text = tokenizer.decode(outputs[0], skip_special_tokens=True)
     logger.warning(f"Generated text: {generated_text}")
 
@@ -441,6 +441,11 @@ class TestCausalLMIntegration:
         model, tokenizer = create_model_and_tokenizer(pipeline)
         logger.warning(f"Model loaded: {model_config['model_name']}")
 
+        callbacks = []
+        optimizer = pipeline._create_optimizer()
+        logger.warning(f"Optimizer created: {type(optimizer[0]).__name__}")
+        # Create trainer using ComponentRegistry
+        training_config = pipeline._prepare_training_config()
         # Load and prepare dataset
         train_dataset, test_dataset = load_and_prepare_dataset(
             dataset_config_test=dataset_config,
@@ -448,11 +453,6 @@ class TestCausalLMIntegration:
             pipeline=pipeline,
         )
         logger.warning(f"Dataset loaded: {len(train_dataset)} samples")
-        callbacks = []
-        optimizer = pipeline._create_optimizer()
-        logger.warning(f"Optimizer created: {type(optimizer[0]).__name__}")
-        # Create trainer using ComponentRegistry
-        training_config = pipeline._prepare_training_config()
         trainer = create_sft_trainer(
             model=model,
             tokenizer=tokenizer,


QEfficient/finetune/experimental/tests/test_integrated.py
```diff
@@ -0,0 +1,332 @@
...
+from QEfficient.finetune.experimental.core.model import HFModel
+from QEfficient.finetune.experimental.core.trainer.sft_trainer import SFTTrainerModule
+from QEfficient.utils.logging_utils import logger
+
+
+class TestParametrizedConfigurations:
+    """Parametrized tests for different model and dataset configurations."""
+
+    @pytest.fixture(autouse=True)
+    def setup_and_cleanup(self):
```
quic-meetkuma: Should we use def setup and def teardown predefined methods? These are executed before and after each tests.

QEfficient/finetune/experimental/tests/test_integrated.py
```diff
@@ -0,0 +1,332 @@
...
+        if os.path.exists(self.test_output_dir):
+            try:
+                shutil.rmtree(self.test_output_dir)
+            except Exception as e:
+                logger.warning(f"Warning: Failed to clean up {self.test_output_dir}: {e}")
+
+    @pytest.mark.parametrize(
+        "model_name,task_type,max_eval_step,max_train_step,dataset_name,data_path_fixture,use_peft,config_name",
+        [
+            pytest.param(
```
quic-meetkuma: Can we convert these configs and function's input argument into some dataclass structure and define those dataclasses like constants at the start of file and use here?

QEfficient/finetune/experimental/tests/test_integrated.py
```diff
@@ -0,0 +1,332 @@
...
+        """
+        from trl import SFTConfig
+
+        # # Get data path if fixture is specified
+        # data_path = None
+        # if data_path_fixture:
+        #     data_path = request.getfixturevalue(data_path_fixture)
+
+        # Determine auto_class_name based on task type
+        if task_type == "CAUSAL_LM":
```
quic-meetkuma: I think there has to be some type of enums in our code base for this.

QEfficient/finetune/experimental/tests/test_integrated.py
```diff
@@ -0,0 +1,332 @@
...
+                "CAUSAL_LM",
+                10,
+                20,
+                "alpaca",
+                None,
+                True,
+                "llama_3.2_1B_config_alpaca_single_device",
+                id="llama_config_alpaca",
+            ),
+            # pytest.param(
```
quic-meetkuma: For BERT kind of models, I guess we may need to come up with separate PR so disabling BERT test case is fine in this PR.

QEfficient/finetune/experimental/tests/test_integrated.py
```diff
@@ -0,0 +1,332 @@
...
+
+        # Load Model
+        model_config = config_manager.get_model_config()
+
+        # Adjust model config for faster testing
+        model_config_kwargs = {}
+        model_config_kwargs["num_hidden_layers"] = 2
+
+        # Create HFModel instance without PEFT config
+        hf_model = HFModel(
```
quic-meetkuma: Use factory

QEfficient/finetune/experimental/tests/test_integrated.py
```diff
@@ -0,0 +1,332 @@
...
+                return prompt["prompt"] + prompt["completion"]
+
+        else:
+            pytest.fail(f"Unknown dataset: {dataset_name}")
+
+        logger.warning(f"Dataset loaded: {len(dataset)} samples")
+
+        # Create SFT Training Config
+        training_config = master_config.training
+        sft_config = SFTConfig(
```
quic-meetkuma: As this is integration file, we should not directly initialize SFTConfig class. I think we should rely on ComponentRegistry and ComponentFactory classes to give us correct class instance or reference.

QEfficient/finetune/experimental/tests/test_integrated.py
```diff
@@ -0,0 +1,332 @@
...
+            args=sft_config,
+            train_dataset=dataset.dataset,
+            processing_class=tokenizer,
+            peft_config=peft_config,
+            formatting_func=formatting_func,
+        )
+        logger.warning("Trainer instantiated")
+        # Run Training
+        logger.warning(f"Starting training for {config_name}...")
+        train_result = trainer.train()
```
quic-meetkuma: add a try catch.

QEfficient/finetune/experimental/tests/test_integrated.py
```diff
@@ -0,0 +1,332 @@
...
+        logger.warning(f"Training completed for {config_name}!")
+
+        # Verify Training Results
+        assert train_result is not None
+        assert hasattr(train_result, "training_loss")
+        logger.warning(f"Training loss: {train_result.training_loss:.4f}")
+
+        # Test Inference
+        if task_type == "CAUSAL_LM":
```
quic-meetkuma: Instead of having if else condition for different task_type, should we split the tests for each tasks? The code duplication should be handled properly by having small reusable function. What do you think?

QEfficient/finetune/experimental/tests/test_trainer.py
```diff
@@ -0,0 +1,299 @@
...
+        trainer_info = registry.get_trainer_module("sft")
+        trainer_cls = trainer_info["trainer_cls"]
+
+        # Instantiate trainer with PEFT config
+        trainer = trainer_cls(
+            model=model,
+            args=sft_config,
+            train_dataset=dummy_dataset,
+            processing_class=tokenizer,
+            peft_config=peft_config,
```
quic-meetkuma: Here we are expecting the SFTTrainer class to convert the model into peft model. So after this, check whether model has been modified to become a peft model? e.g. trainable params, presence of lora weights etc.

QEfficient/finetune/experimental/tests/test_trainer.py
```diff
@@ -0,0 +1,299 @@
...
+        assert trainer.tokenizer is not None
+
+    def test_sft_trainer_with_peft_model(self, model_config, peft_model_config, dummy_dataset):
+        """Test that SFTTrainer works with PEFT-enabled models."""
+
+        # Load model and tokenizer
+        hf_model = HFModel(**model_config)
+        model = hf_model.load_model()
+        # Load PEFT Config
+        peft_config = LoraConfig(peft_model_config)
```
quic-meetkuma: Try to move this to some utility file

QEfficient/finetune/experimental/tests/test_integrated.py
```diff
@@ -0,0 +1,332 @@
...
+    SchedulerConfig,
+    TrainingConfig,
+)
+from QEfficient.finetune.experimental.core.dataset import SFTDataset
+from QEfficient.finetune.experimental.core.model import HFModel
+from QEfficient.finetune.experimental.core.trainer.sft_trainer import SFTTrainerModule
+from QEfficient.utils.logging_utils import logger
+
+
+class TestParametrizedConfigurations:
```
quic-meetkuma: These tests are integration tests. In future we may need to write tests which also checks the individual steps loss values. How to do that? How to reuse this code when we write those comparative tests?

PR #707, opened 2026-01-07
Added changes to load and export Llama model in bfloat16/float16 precision

quic-dhirajku:
test_bf16_export.py script can be used to see how to export a bf16 model.


quic-dhirajku:
Changes done in a separate PR along with other models as well.

PR #708, opened 2026-01-07
Adding support for multi-node DDP training

smedhe:
Add support for multi-node Distributed Data Parallel (DDP) training to the QEfficient finetuning pipeline. This enables scaling training across multiple nodes while keeping the existing single-node behavior unchanged.

Commands for DDP across 2 servers:
For the Master Addr or the Primary Machine, use node-rank as 0:
QAIC_VISIBLE_DEVICES=0,1,2,3 torchrun --nnodes=2 --nproc-per-node=4 --seed 0 --node-rank=0 --master_addr=<MASTER_NODE_IP> --master_port=8000 -m QEfficient.cloud.finetune --device qaic --enable_ddp --model_name "meta-llama/Llama-3.2-1B" --dataset alpaca_dataset --train_batch_size 1 --val_batch_size 1 --num_epochs 1 --max_train_step 200 --max_eval_step 50

For Node 1, use node-rank as 1:
QAIC_VISIBLE_DEVICES=0,1,2,3 torchrun --nnodes=2 --nproc-per-node=4 --seed 0 --node-rank=1 --master_addr=<MASTER_NODE_IP> --master_port=8000 -m QEfficient.cloud.finetune --device qaic --enable_ddp --model_name "meta-llama/Llama-3.2-1B" --dataset alpaca_dataset --train_batch_size 1 --val_batch_size 1 --num_epochs 1 --max_train_step 200 --max_eval_step 50

### DIFF
--- a/QEfficient/cloud/finetune.py
+++ b/QEfficient/cloud/finetune.py
@@ -6,7 +6,6 @@
 # -----------------------------------------------------------------------------
 
 import logging
-import os
 import random
 import warnings
 from typing import Any, Optional, Union
@@ -29,7 +28,7 @@ from QEfficient.finetune.utils.config_utils import (
 )
 from QEfficient.finetune.utils.dataset_utils import get_dataloader, get_longest_seq_length
 from QEfficient.finetune.utils.device_map import get_device_map
-from QEfficient.finetune.utils.helper import Task_Mode, get_world_size
+from QEfficient.finetune.utils.helper import Task_Mode, get_local_rank, get_local_world_size, get_rank, get_world_size
 from QEfficient.finetune.utils.logging_utils import logger
 from QEfficient.finetune.utils.parser import get_finetune_parser
 from QEfficient.finetune.utils.train_utils import print_model_size, print_trainable_parameters, train
@@ -92,13 +91,10 @@ def setup_distributed_training(train_config: TrainConfig) -> None:
     assert torch_device.index is None, f"DDP requires only device type (qaic/cuda), got: {torch_device}"
 
     # Torchrun-provided env vars
-    world_size = int(os.environ.get("WORLD_SIZE", "1"))
-    rank = int(os.environ.get("RANK", "0"))
-    local_rank = int(os.environ.get("LOCAL_RANK", "0"))
-    local_world_size = int(os.environ.get("LOCAL_WORLD_SIZE", "1"))
-
-    # Sanity check vs helper
-    assert world_size == get_world_size(), f"WORLD_SIZE={world_size} but get_world_size()={get_world_size()}"
+    world_size = get_world_size()
+    rank = get_rank()
+    local_rank = get_local_rank()
+    local_world_size = get_local_world_size()
 
     # Per-node device validation
     num_available_devices = getattr(torch, torch_device.type).device_count()
--- a/QEfficient/finetune/utils/helper.py
+++ b/QEfficient/finetune/utils/helper.py
@@ -55,6 +55,15 @@ def get_rank() -> int:
     return int(os.getenv("RANK", 0))
 
 
+def get_local_rank() -> int:
+    """Get the current local rank of the process.
+
+    In DDP, this should correspond to the 'LOCAL_RANK' environment variable set by torchrun.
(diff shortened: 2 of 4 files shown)


QEfficient/cloud/finetune.py
```diff
@@ -67,32 +66,60 @@ def setup_distributed_training(train_config: TrainConfig) -> None:
...
+
     dist_backend_map = {"cpu": "gloo", "qaic": "qccl", "cuda": "gloo"}
-    dist.init_process_group(backend=dist_backend_map[torch_device.type])
+    dist.init_process_group(dist_backend_map[torch_device.type], rank=rank, world_size=world_size)
+
+    # Set the base device index for this process on this node
+    # For PP: each process controls num_pp_stages devices starting from base_device_index
+    base_device_index = local_rank * train_config.num_pp_stages
     # from here onward "qaic/cuda" will automatically map to "qaic:i/cuda:i", where i = process rank
-    getattr(torch, torch_device.type).set_device(dist.get_rank() * train_config.num_pp_stages)
```
quic-akuruvil: Test whether this change would affect the PP pipeline.

QEfficient/finetune/utils/helper.py
```diff
@@ -47,13 +47,12 @@ def enum_names(enum_cls: Enum) -> List[str]:
 
 
 def get_rank() -> int:
```
quic-swatia: Please add another method ( get_local_rank() )which returns local rank which was earlier returned by the current method.

QEfficient/cloud/finetune.py
```diff
@@ -67,32 +66,60 @@ def setup_distributed_training(train_config: TrainConfig) -> None:
...
         return
 
+    # ---- DDP path (single- or multi-node) ----
     assert torch_device.type != "cpu", "Host doesn't support single-node DDP"
-    assert torch_device.index is None, f"DDP requires only device type, got: {torch_device}"
+    assert torch_device.index is None, f"DDP requires only device type (qaic/cuda), got: {torch_device}"
+
+    # Torchrun-provided env vars
+    world_size = int(os.environ.get("WORLD_SIZE", "1"))
+    rank = int(os.environ.get("RANK", "0"))
```
quic-swatia: Why are we not fetching it from the method present in helper file? Please explain if I am missing something here. 
IMO rank and world_size should be fetched directly from the helper methods while there should be methods defined for local_rank and local_world_size inside the helper.py

QEfficient/cloud/finetune.py
```diff
@@ -67,32 +66,60 @@ def setup_distributed_training(train_config: TrainConfig) -> None:
...
+    local_rank = int(os.environ.get("LOCAL_RANK", "0"))
+    local_world_size = int(os.environ.get("LOCAL_WORLD_SIZE", "1"))
+
+    # Sanity check vs helper
+    assert world_size == get_world_size(), f"WORLD_SIZE={world_size} but get_world_size()={get_world_size()}"
```
quic-swatia: both world_size and get_world_size() are assigned os.environ.get("WORLD_SIZE", "1"), so it's not required.

QEfficient/finetune/utils/helper.py
smedhe: Resolved.

quic-swatia:
> For PP, device_map.py uses get_rank() which was earlier returning local rank but now returns rank. This will not break PP as of now, but it's a wrong practice. In case PP is used with multi node, it would break. Please call the get_local_rank() in device_map once it is defined.

After this change, we can also test if PP + DDP for multi node is working with this PR. It seems like it might work. It's okay if it doesn't, we can still go ahead with the PR.

quic-meetkuma (approved):
Looks fine to me. Please update the finetune.md file. Imagine or other teams refer the commands from there. They will need to change command at their side of the code.

QEfficient/cloud/finetune.py
quic-akuruvil: @smedhe Please push the changes and then mark as resolve

quic-akuruvil (commented):
Start CI run also for this. So CI passes before merge

PR #712, opened 2026-01-09
Updating instructions to load 2-layer Wan model

tv-karthikeya:
Readme update for Wan T2V model


tv-karthikeya:
will be raising a Cherry pick from #715 mainline

PR #720, opened 2026-01-12
Updating MDP partition config: prioritizing dump over load

asmigosw:
QEfficient should ignore providing `-mdp-load-partition-config` when `-mdp-dump-partition-config` is provided in compiler_options of compile API.

### DIFF
--- a/QEfficient/base/modeling_qeff.py
+++ b/QEfficient/base/modeling_qeff.py
@@ -445,6 +445,10 @@ class QEFFBaseModel(ABC):
         user_provided_load_config = False
 
         if mdp_dump_json_path:
+            if mdp_ts_json_path:
+                logger.warning(
+                    "Loading and Dumping partition is not supported at the same time. Prioritizing dump config over load config!"
+                )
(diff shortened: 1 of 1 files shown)


QEfficient/base/modeling_qeff.py
```diff
@@ -438,8 +438,21 @@ def _compile(
...
+        if mdp_dump_json_path:
+            command.append(f"-mdp-dump-partition-config={mdp_dump_json_path}")
+        elif mdp_ts_json_path:
             command.append(f"-mdp-load-partition-config={mdp_ts_json_path}")
+            mdp_ts_json = load_json(str(mdp_ts_json_path))
+        elif mdp_ts_num_devices > 1:
+            # Generate mdp config only if neither dump nor load is provided and num_devices > 1
+            mdp_ts_json = generate_mdp_partition_config(
+                mdp_ts_num_devices, compiler_options.get("aic_num_cores", constants.DEFAULT_AIC_NUM_CORES)
+            )
```
quic-hemagnih: Can we put an error log in the else case, so that we handle the failure scenarios.

PR #725, opened 2026-01-13
Fixing the default value of CCL in infer.py

vjanfaza:
Since CCL is deactivated by default, the value of CCL lists (ccl_prefill and ccl_decode) should be None by default. In infer.py script the value of these lists wasn't None and it caused the problem of ccl activation by default. In this PR we addressed this issue.

### DIFF
--- a/QEfficient/cloud/infer.py
+++ b/QEfficient/cloud/infer.py
@@ -138,6 +138,7 @@ def main(
     enable_qnn: Optional[bool] = False,
     qnn_config: Optional[str] = None,
     trust_remote_code: Optional[bool] = False,
+    ccl_enabled: Optional[bool] = False,
     **kwargs,
 ) -> None:
     """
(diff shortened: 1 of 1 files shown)


QEfficient/cloud/infer.py
```diff
@@ -343,15 +347,21 @@ def main(
...
     )
     parser.add_argument(
         "--comp-ctx-lengths-decode",
         type=lambda comp_ctx_lengths_decode: [int(x) for x in comp_ctx_lengths_decode.split(",")],
-        default=[2048],
+        default=None,
         help="Define ccl list in csv format (e.g.,--comp-ctx-lengths 512,1024,2048).",
     )
+    parser.add_argument(
+        "--ccl_enabled",
```
quic-rishinr: its better to keep the flag name as enable_ccl

quic-rishinr (changes requested):

QEfficient/cloud/infer.py
vjanfaza: We have used this flag name in the ccl implementation. I think it's better to keep it consistent with other scripts.

PR #699, opened 2026-01-05
Added support of subfunction for VLMs

abhishek-singh591:

PR #733, opened 2026-01-16
Fix for Qwen 2.5 VL with subfunction

abhishek-singh591:


abhishek-singh591:
This is temporary fix for someone who want to test qwen 2.5 VL with subfunction, there is generalized PR #699. pls refer that making it draft for time being.

PR #745, opened 2026-01-20
Adding blocked kv and skip softmax for gpt oss

kdulla:


kdulla:
Moved to updated generic interface, PR no longer needed

PR #731, opened 2026-01-16
[QEff. Finetuning] Adding finetune_experiemental.py and related files

quic-swatia:
1) Added FinetuningPipeline (finetune_experiemental.py) which integrates all the components added for HF-trainer and enable running fine tuning through it. 
2) Added files to handle PEFT and training config. 
3) Made changes in the config_manager and callbacks files. 
4) Added unit tests for the FinetuningPipeline (test_finetune.py)
5) Updated tests in test_callback and test_config_manager based on above changes.

PR #747, opened 2026-01-21
['QEff.finetuning'] Changing some params from training config to model config

tchawada:
This PR contain:
1.documentation for new finetune experimental stack.
2. Updates inconfig_manager.py


quic-akuruvil (commented):
Add a sample working config in configs/ folder.

docs/source/hf_finetune.md
```diff
@@ -0,0 +1,438 @@
...
+#### **1. Alpaca (yahma/alpaca-cleaned)**
+
+```yaml
+dataset:
+  tokenizer_name: "meta-llama/Llama-3.2-1B"
+  dataset_type: "seq_completion"
+  dataset_name: "yahma/alpaca-cleaned"
+  train_split: "train"
+  test_split: "test"
+  max_seq_length: 512
```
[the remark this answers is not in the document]
quic-akuruvil: okay, what if alpaca dataset has some sample which has seq_len > 512. On what basis is it set to 512 here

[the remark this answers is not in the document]
quic-akuruvil: I think this default we set, (for each dataset) should not be any random value. It should be dependent on length of samples in the dataset, or based on which is the max the hardware can support etc. Might need an analysis on the samples of the dataset. We might lose some samples if its length exceeds 512.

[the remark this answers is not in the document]
quic-akuruvil: Anyways, we can work on setting defaults after further analysis later. Now this would be okay to complete the end to end testing. But please note this down for future resolution.

docs/source/hf_finetune.md
```diff
@@ -0,0 +1,438 @@
...
+```bash
+# torch_qaic (example wheel path — adjust to your environment)
+pip install /opt/qti-aic/integrations/torch_qaic/py310/torch_qaic-0.1.0-cp310-cp310-linux_x86_64.whl
+
+# accelerate 
+pip install /opt/qti-aic/integrations/accelerate/py310/accelerate-1.10.0-py3-none-any.whl
+```
+
+> **Note**  
+> If you’re using the `qeff-env` Docker environment, `torch_qaic` and `accelerate` may already be installed.
```
quic-meetkuma: I believe qeff_env should not be used. We should only rely on torch_qaic_env because torch_qaic_env is for FT and it may have different pytorch requirement than qeff_env.

docs/source/hf_finetune.md
```diff
@@ -0,0 +1,438 @@
...
+
+```python
+# registry/datasets.py
+import json
+from torch.utils.data import Dataset
+from .base import register  # your registry base
+
+@registry.dataset( "my_custom_dataset")
+class MyCustomDataset(Dataset):
+    def __init__(self, path: str, prompt_template: str = None):
```
quic-meetkuma: Is the signature of __init__ is correct? How does it take the yaml's dataset's different attribute ? I believe we were taking kwargs to allow user to send any arguments from yaml to init. Fixing this might be blocking that behavior. What do you think?

docs/source/hf_finetune.md
```diff
@@ -0,0 +1,438 @@
...
+In your config, reference an HF dataset and a template function name:
+
+```yaml
+data:
+  hf_dataset: "tatsu-lab/alpaca"
+  split: "train"
+  prompt_template: "alpaca_func:format_alpaca"
+```
+
+Define the function (e.g., in `preprocess/alpaca_func.py`):
```
quic-meetkuma: e.g. preprocess/alpaca_func.py --> This is given but in your above example yaml snippet, you missed preprocess directory. I think that might create confusion.

docs/source/hf_finetune.md
```diff
@@ -0,0 +1,438 @@
...
+
+```python
+def format_alpaca(example):
```
quic-meetkuma: Add a line similar to L50 indicating file path.

docs/source/hf_finetune.md
```diff
@@ -0,0 +1,438 @@
...
+  fsdp_config: "./configs/accelerate/fsdp_tp_parallelism_config.yaml"
+```
+- **Pipeline Parallelism**: Split model layers across devices.
+- **Tensor Parallelism**: Split tensors across devices.
+
+***
+## Component Registry
+The training script uses a component registry to manage different components like models, optimizers, and datasets. This allows for easy swapping of components without modifying core logic.
+
+To register a new component, use the `@registery` decorator.
```
quic-meetkuma: I think prepare data should be little below of all these config related section. It uses component factory and here we are actually introducing component factory. One option is to move all the documentation to config.md file so any user who want to get detailed documentation of all the knobs, can read that file. That way this entire main md file will not look bloated.

QEfficient/finetune/experimental/core/config_manager.py
```diff
@@ -113,6 +114,22 @@ class DatasetConfig:
         default=4,
         metadata={"help": "Number of workers for dataset processing."},
     )
+    prompt_template: str = field(
+        default=None,
+        metadata={"help": "Template for formatting prompts (e.g., 'User: {input} Assistant: ')."},
+    )
+    prompt_func: str = field(
+        default=None,
+        metadata={"help": "Function for formatting prompts (e.g., 'User: {input} Assistant: ')."},
```
quic-meetkuma: e.g. should include path of the function. Isn't it?

QEfficient/finetune/experimental/core/config_manager.py
```diff
@@ -460,47 +477,59 @@ class MasterConfig:
...
+                raise ValueError(f"Expected a .yaml/.yml file, got: {config_path}")
+
+            try:
+                config = self.load_config(config_path)
+            except Exception as e:
+                raise ValueError(f"Failed to parse YAML config '{config_path}': {e}")
+        else:
+            self.config = MasterConfig()
+
+        self.config = asdict(self.config)
```
quic-meetkuma: L531 and L532 are not needed in new flow.

QEfficient/finetune/experimental/core/config_manager.py
```diff
@@ -598,16 +626,20 @@ def validate_config(self) -> None:
         """
         Validate configuration parameters for MasterConfig.
         """
+        cfg = self.config
```
quic-meetkuma: Do you think we should add a validation code to validate some of the params. e.g. "qaic" device type is given but torch_qaic itself is not loaded then we can not move forward.

QEfficient/finetune/experimental/core/config_manager.py
quic-akuruvil: The code uses both List[str] and list[str] for type annotations. Standardize on one style (preferably list[str])

QEfficient/finetune/experimental/core/config_manager.py
```diff
@@ -632,47 +664,39 @@ def validate_config(self) -> None:
...
 
         # Epochs / steps
-        n_epochs = training.get("num_train_epochs", 0)
-        max_steps = training.get("max_steps", -1)
+        n_epochs = training.get("num_train_epochs", 1)
         self._push(
             errors,
-            n_epochs <= 0 and max_steps <= 0,
+            n_epochs <= 0,
             "Either training.num_train_epochs > 0 or training.max_steps > 0 must be set.",
```
quic-meetkuma: Why max_steps is removed? If it is not needed update the comment as well.

QEfficient/finetune/experimental/tests/test_config.yaml
```diff
@@ -100,5 +95,5 @@ callbacks:
   early_stopping:
     early_stopping_patience: 3
     early_stopping_threshold: 0.001
-  tensorboard:
```
quic-meetkuma: Why it is removed?

QEfficient/finetune/experimental/core/config_manager.py
```diff
@@ -460,47 +477,59 @@ class MasterConfig:
...
-
+    master_config = MasterConfig()
     return master_config
 
 
 class ConfigManager:
     """Manages configuration loading, validation, and updates."""
 
-    def __init__(self, config: MasterConfig):
+    def __init__(self, config: MasterConfig, config_path: Optional[str] = None):
```
quic-meetkuma: The __init__ should take config only if user wants to override its values and pass it to ConfigManager. Same way any use cases where user want to use a config stored at config_path, then config_path argument is used. 

For our use case where we are invoking finetuning from CLI, the ConfigManager should not be given anything because it parses CLI argument within its __init__.

Accordingly, changes should be made in #731.

CC: @quic-swatia

QEfficient/finetune/experimental/core/config_manager.py
tchawada: It depends on dataset we are using, I make it default for alpaca

PR #779, opened 2026-02-05
removed duplication of `mdp_json_path` in compilation command (#706)

ochougul:
Needed for passing custom config via vllm.

---------

### DIFF
--- a/QEfficient/transformers/models/modeling_auto.py
+++ b/QEfficient/transformers/models/modeling_auto.py
@@ -2998,10 +2998,12 @@ class QEFFAutoModelForCausalLM(QEFFBaseModel):
         Dict[str, Union[int, str]]
             A dictionary defining the prefill specialization.
         """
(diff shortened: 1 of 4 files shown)


QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3031,7 +3031,7 @@ def compile(
                     )
                 )
 
-        if prefill_only is None or not prefill_only:
+        if (prefill_only is None or not prefill_only) and prefill_seq_len != 1:
```
quic-rishinr: Can we set the prefill_only to false as default value?

quic-rishinr (changes requested):

QEfficient/transformers/models/modeling_auto.py
ochougul: this leads to single specialization when prefill_seq_len>1

PR #783, opened 2026-02-09
Updating Wan Vae pytest, Export to 45p, minor issues

tv-karthikeya:


quic-amitraj:
@tv-karthikeya lets create the dummy model for the wan2.2 and flux. Same as the other transformers models we have done in pytest. @abukhoy please take a look into this.

`This will significantly decrease the pytest model loading time.`

PR #804, opened 2026-02-23
Updating dependencies for upgrading TF version 5.0.0

asmigosw:


asmigosw:
Will be requesting for TF version 5.5

PR #845, opened 2026-03-10
[QEff. Finetuning]: Fixed Data Parallel issue

quic-swatia:
1) Fixed Data Parallel issue.
2) Removed sample config for PP as the changes of config manager are now merged. 
3) Updated the PP documentation accordingly.

### DIFF
--- a/QEfficient/finetune/experimental/core/trainer/sft_trainer.py
+++ b/QEfficient/finetune/experimental/core/trainer/sft_trainer.py
@@ -12,7 +12,7 @@ from QEfficient.finetune.experimental.core.config_manager import PeftConfig
 
 @registry.trainer_module(name="sft", args_cls=SFTConfig, required_kwargs={"peft_config": PeftConfig})
 class SFTTrainerModule(SFTTrainer):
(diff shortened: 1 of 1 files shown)


docs/source/hf_finetune.md
```diff
@@ -248,15 +248,15 @@ training:
...
 
 ### Launch commands
 
 **PP only — single process, 2 stages (via YAML)**
 ```bash
-python -m QEfficient.cloud.finetune_experimental configs/sample_pp_config.yaml
+python -m QEfficient.cloud.finetune_experimental configs/sft_single_device_gsm8k_config.yaml
 ```
-where `sample_pp_config.yaml` contains `pp_degree: 2` under `training:`.
+where `sft_single_device_gsm8k_config.yaml` contains `pp_degree: 2` under `training:`.
```
quic-akuruvil: For `pp_degree 2` , 2 devices are required right? But the config file name is ambiguous here. Can we not create a different file? And keep this file for single_device run?

quic-swatia: As discussed offline, since we only need one extra  param ( i.e. pp_degree ) to be set >1, it will be better to reuse the existing config file.

PR #889, opened 2026-03-26
Add old stack logging support to new stack

quic-abhamidi:


quic-akuruvil (commented):
Also check PP, DDP and Multinode DDP APIs are logging properly, with the new change.

quic-akuruvil (changes requested):

quic-akuruvil (changes requested):

quic-akuruvil (changes requested):

quic-akuruvil (dismissed):
LGTM

quic-abhamidi:
Adding this change to v1 branch for merging.

PR #801, opened 2026-02-20
Qwen3Vl

qcdipankar:
1. Added Support for Qwen3-VL
2. Added Support for Qwen3-VL-Moe
3. Added Support for Disaggregated VL MOE Model. (Three QPC )

PR #900, opened 2026-04-01
added interleave changes to qwen3vl

qcdipankar:
1. Added cache changes of rotary aligned to 4.57.3
2. Added interleave method


qcdipankar:
implemented in #801

PR #905, opened 2026-04-03
feat(Llama GQA): optional shared-KV matmul rewrite for attention. WIP: (env-gated)

vbaddi:
This PR adds an optional Llama GQA attention rewrite that avoids explicit KV expansion in the normal eager attention path. The optimization is gated by an environment variable and is OFF by default, so baseline behavior is unchanged unless enabled. 

  **How to enable for testing**
`export QEFF_LLAMA_GQA_SHARED_KV=1`


quic-rishinr:
Closing this PR as there is no progress since last few week. Please reopen the PR if required.

PR #907, opened 2026-04-06
[QEff. Finetuning] Adding TP+DDP support in HF Trainer stack

smedhe:
Adding TP+DDP support in the new hf trainer stack. It is still an experimental feature due to transformers and accelerate versions mismatch.
Further TO-DO:
load_in_4_bits flag is not supported in v5.1.0, commented out since it is not used
need to add steps for running tp+ddp in the config.md


docs/source/hf_finetune.md
```diff
@@ -324,6 +324,130 @@ QAIC_VISIBLE_DEVICES=0,1 python -m QEfficient.cloud.finetune_experimental \
...
+
+>  **Note:** TP requires a specific set of dependencies. Follow the steps below carefully.
+
+```bash
+python -m venv finetune_env
+source finetune_env/bin/activate
+
+git clone https://github.com/quic/efficient-transformers.git
+cd efficient-transformers
+git checkout ft_experimental      # Can remove this once merged to mainline
```
quic-akuruvil, line 363: ft_v1

quic-akuruvil (changes requested):

quic-akuruvil (changes requested):
Add inference post the checkpoints are saved. And for checkpoint saving stick to native HF checkpoints saving. So it is easier to load model, and add tuned peft adapters.

PR #912, opened 2026-04-10
revert(export): Revert proxy-only ONNX transform gating and restore default export behavior

vbaddi:
## Summary
  Reverts the ONNX transform changes from the related branches and restores the previous mainline behavior.

  ## Changes

  - Restore default SplitTensorsTransform and FP16ClipTransform behavior
  - Remove proxy-only ONNX transform gating
  - Remove the embedding-only FP16 special-case path

  ## Validation
  `python -m pytest -q tests/unit_test/models/test_model_quickcheck.py -n auto`


quic-rishinr:
Closing the PR as PR is no longer need on release branch

PR #914, opened 2026-04-13
Removed redundancies from QEFFHybridCache and QEFFHybridChunkedCache

quic-mamta:
Removed redundancies from QEFFHybridCache and QEFFHybridChunkedCache, It should improve perf for models that use sliding window like gemma, mistral etc.


vbaddi:
@mamtsing can you check why does the quick check workflow fail and fix it pls. thanks

quic-mamta:
not relevent

PR #918, opened 2026-04-17
[Tests]: gemma3 tests are enabled

abukhoy:


tests/configs/image_text_model_configs.json
```diff
@@ -76,7 +76,7 @@
       "img_size": 896,
       "img_url": "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/datasets/cat_style_layout.png",
       "text_prompt": "Can you describe the image in detail.",
-      "num_layers": 6,
+      "num_layers": 2,
```
quic-hemagnih, line 79: This model has to run for minimum 6 layers

tests/configs/image_text_model_configs.json
```diff
@@ -591,7 +596,7 @@
...
```
quic-hemagnih, line 599: same here 6 layers minimum

tests/configs/image_text_model_configs.json
```diff
@@ -88,10 +88,15 @@
       "full_batch_size": 2,
       "additional_params": {
         "text_config": {
-        "_sliding_window_pattern": 2,
+        "sliding_window_pattern": 2,
```
asmigosw, line 91: Won't this differ from original config file? And since you are updating this in code itself below, do we need this param here?

tests/configs/image_text_model_configs.json
quic-hemagnih: Layer 0 to 5 : full attention and layer 6: sliding window

tests/configs/image_text_model_configs.json
abukhoy: No, it will not differ the original config. I have checked the config when I tested it.

tests/configs/image_text_model_configs.json
```diff
@@ -88,10 +88,15 @@
...
+        "head_dim": 256,
```
quic-hemagnih, line 92: How head_dim :256 is computed? Also why sliding_window_pattern is 2?

tests/configs/image_text_model_configs.json
abukhoy: No need of 6 layers, I have changed the layer types. In 2 layers, we can check the model with full_attention and sliding_attention.

tests/configs/image_text_model_configs.json
abukhoy: No need of 6 layers, I have changed the layer types. In 2 layers, we can check the model with full_attention and sliding_attention.

tests/transformers/models/image_text_to_text/test_continuous_batching.py
```diff
@@ -68,6 +68,9 @@ def check_image_text_to_text_pytorch_vs_kv_vs_ort_vs_ai100_CB(
             model_name, trust_remote_code=True, padding=model_name not in ModelConfig.MOLMO_MODELS
         )
         config = set_num_layers_vlm(config, n_layer=n_layer)
+        if hasattr(config, "model_type") and config.model_type in ["gemma3"]:
+            config.text_config._sliding_window_pattern = 2
```
quic-hemagnih, line 72: Original model has slidding window length as 6

quic-hemagnih (changes requested):

tests/transformers/models/image_text_to_text/test_continuous_batching.py
abukhoy: Yes, that's true. But we can modify, right?

tests/configs/image_text_model_configs.json
abukhoy: sliding window pattern, we can change according to the layer types.

tests/transformers/models/image_text_to_text/test_continuous_batching.py
abukhoy, line 72: @asmigosw  as you can see that I am setting _sliding_window_pattern here.

asmigosw (approved):
LGTM

PR #927, opened 2026-04-20
CB Bug fix for Qwen3VL Dense and basic cleaning of example script and Model File

qcdipankar:
Issue: When running CB with single image multiple time with different prompts it was working whereas with multiple images and prompts it was failing.

Resolution: The hidden states calculation was done with index [0] every time as a result when the broadcasting was happening with batch size 4 it was taking the first image only and we were getting wrong output in CB. Taking full hidden_states ensures the error does not occur.


quic-rishinr:
@qcdipankar please include the test for full_batch_size ≥ 2 with multiple images

quic-rishinr (approved):

PR #938, opened 2026-04-22
Logging support added for HF Trainer stack

quic-abhamidi:

Note: Sample output logger file for reference.
[new_stack_dpp_4_device_run_report.txt](https://github.com/user-attachments/files/26963021/new_stack_dpp_4_device_run_report.txt)


quic-akuruvil (changes requested):
Rebase against ft_v1 branch. Not up to date with latest tip of ft_v1

QEfficient/finetune/experimental/core/callbacks.py
```diff
@@ -31,6 +38,113 @@
 registry.callback("default_flow")(DefaultFlowCallback)
 registry.callback("tensorboard")(TensorBoardCallback)
 
+logger = Logger(__name__)
+# Setting the path for dumping the log file
+output_dir = Path(ConfigManager().config.training["output_dir"])
+log_file_name = os.path.join(output_dir, f"training_logs_{datetime.now().strftime('%Y%m%d_%H%M%S')}.txt")
```
quic-akuruvil: Since default log file is mentioned here. Check the functionality, if user passes custom log_file path, will it be overridden/taken?

quic-akuruvil (changes requested):

QEfficient/finetune/experimental/core/callbacks.py
```diff
@@ -31,6 +38,113 @@
 registry.callback("default_flow")(DefaultFlowCallback)
 registry.callback("tensorboard")(TensorBoardCallback)
 
+logger = Logger(__name__)
+# Setting the path for dumping the log file
+output_dir = Path(ConfigManager().config.training["output_dir"])
```
quic-akuruvil: At import time ConfigManager().config may reflect defaults, not the final config, and override will be missed, leading to wrong output_dir

quic-akuruvil (changes requested):

QEfficient/finetune/experimental/core/callbacks.py
quic-abhamidi: No, this was checked, when user provides the output_dir in the .yaml file, that path is taken.

quic-akuruvil (commented):
Make sure to run all the local FT pytests from the project repo. And it is passed

QEfficient/finetune/experimental/core/callbacks.py
quic-akuruvil: Okay, test the case where output_dir is given as CLI flag as well

quic-akuruvil (changes requested):
I can see AOT files in this PR. It should not come. Avoid rebasing with main. I will do that

quic-abhamidi:
Closing this PR as it was rebased against main.

PR #943, opened 2026-04-27
Logging support added for HF Trainer stack

quic-abhamidi:


quic-abhamidi:
Hi @quic-akuruvil, please use this PR for logger as it is rebased with ft_experimental_v1 branch and all the comments of PR #938 have been addressed here. 
This PR also passes all the pytests

quic-akuruvil (approved):
LGTM

PR #946, opened 2026-04-27
Compile_only flag removal

quic-rishinr:
Removed compile_only flag from compilation configs and helper


tv-karthikeya (approved):

quic-hemagnih (changes requested):
Can we also remove from ./QEfficient/compile/compile_helper.py:142:

quic-hemagnih (approved):

PR #952, opened 2026-04-28
fix: improve weight offloading to handle plain tensor attrs and use to_empty()

quic-rishinr:
fix: improve weight offloading to handle plain tensor attrs and use to_empty()

Replace manual storage resizing with `to_empty(device="meta")` for
parameters/buffers and explicitly handle plain tensor attributes (e.g.
stacked expert weights in MoE models) that are not registered as
parameters or buffers. This ensures all tensors are properly moved to
the meta device, reducing memory usage after ONNX export.

Add unit tests for plain tensor attribute clearing


quic-rishinr:
@ochougul @abhishek-singh591 please review the PR

vbaddi (approved):

abhishek-singh591 (approved):
LGTM after the last commit, Validated with llama 3 8B model.

PR #957, opened 2026-04-30
Optimize attention blocking nested loops 

anujgupt-github:
changed the code from doing the exact same math repeatedly.


kdulla (approved):
Looks good to me

quic-rishinr (approved):

PR #960, opened 2026-05-04
[QEff. Finetuning] TP+DDP for transformers upgrade to v5.5.4 

smedhe:
This pull request adds TP + DDP support and updates the codebase to remain compatible with newer versions of the dependent libraries. The changes have been validated against the following library versions:

transformers 5.5.4
peft 0.19.1
trl 1.2.0
accelerate 1.13.0


quic-akuruvil (changes requested):
There are so many redundant file changes. Take transforemrs V5.5.4 upgrade branch from models' team, and then raise a PR, keeping that as the base branch.

PR #976, opened 2026-05-11
Automatic Blocking Configurator Bugfix and Enabling skip_kv by default

kdulla:
Fixed a minor issue in automatic blocking configurator that would drop head blocking when requested. Added a unit test to validate blocking modes after automatic configuration to catch similar bugs in the future. 

Also set skip_kv to be true by default since that is optimal behavior.

### DIFF
--- a/QEfficient/blocking/blocking_configurator.py
+++ b/QEfficient/blocking/blocking_configurator.py
@@ -42,6 +42,10 @@ def _infer_data_bytes(compile_config: Dict[str, Any]) -> int:
 
 def _normalize_attention_mode(raw_mode: str) -> str:
     mode = raw_mode.lower()
+    if "h" in mode and ("q" in mode or "kv" in mode):
+        return "hqkv"
+    if "h" in mode:
+        return "h"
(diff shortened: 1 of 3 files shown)


QEfficient/blocking/blocking_configurator.py
```diff
@@ -42,6 +42,10 @@ def _infer_data_bytes(compile_config: Dict[str, Any]) -> int:
 
 def _normalize_attention_mode(raw_mode: str) -> str:
     mode = raw_mode.lower()
+    if "h" in mode and ("q" in mode or "kv" in mode):
+        return "hqkv"
```
ochougul: does this mean there can't be separate `hkv` or `hq` blocking?

kdulla: Yes, right now we do hkv and hq blocking by passing them through the hqkv blocked attention forward, with num_q_blocks = 1 or num_kv_blocks = 1 if needed.

ochougul (approved):

vbaddi (approved):
LGTM, thanks 👍

PR #981, opened 2026-05-13
Onboarding DFlash: Block Diffusion Speculative Decoding

fannanya:
In this FR, we are adding the support for DFlash (Block Diffusion for Flash Speculative Decoding) approach that significantly accelerate LLM inference on Qualcomm platforms.
DFlash replaces autoregressive draft models with a diffusion-based block drafter that predicts an entire block of tokens in a single parallel pass, resulting in constant drafting latency independent of block size.

Key benefits:

5–6× lossless speedup over baseline decoding
~2.5× faster than EAGLE-3
Higher acceptance length enabled by KV injection of target hidden states into every draft layer
Improved accelerator utilization and reduced end-to-end latency

Verification is performed by the frozen target model, guaranteeing output identical to standard decoding.

Use-case(s)
High-throughput and low-latency LLM serving
Long chain-of-thought reasoning workloads (e.g., math and planning)
Code generation and agentic inference loops
Platform differentiation via advanced speculative decoding on AI100
Requirements
DFlash speculative decoding with block diffusion drafting and target verification
Extraction and fusion of hidden states from multiple target layers during prefill
KV cache injection of fused target context into all draft layers
Integration with QEfficient, vLLM, and other applications
Support current architecture:
Target: single QPC for prefill + decode (expandable to DA serving)
Draft: lightweight QPCs with small KV cache (e.g., 16 tokens)


quic-rishinr:
@fannanya is this PR ready for review?

fannanya:
Yes Rishin. It is I have performed some test cases for reviewing this PR, according to my understanding. Let me know if you need any information or give suggestions on this PR.

PR #984, opened 2026-05-13
support multiple TLM decode specializations via num_speculative_tokens list 

eplatero97:
## Description

Speculative decoding on QAIC requires statically compiled shapes. Previously, a TLM
could only be compiled for a single proposal length K, forcing every decode step to
run the full `seq_len=K+1` kernel regardless of how many tokens the draft model
actually proposed. This PR allows compiling multiple decode specializations in one
QPC so the runtime can dispatch to the smallest kernel that covers the actual proposal
count, reducing unnecessary compute on short proposals without changing correctness.

### Summary

- `QEFFAutoModelForCausalLM.compile()` now accepts `num_speculative_tokens` as a
  `List[int]`. Each value K compiles one TLM decode specialization (`seq_len=K+1`,
  `num_logits_to_keep=K+1`), enabling per-step dispatch to the cheapest kernel that
  covers the actual proposal count.
- Plain `int` input still works (backward compatible — treated as `[K]`).
- Removes `enable_fallback_decode_spec`; equivalent behavior is `[0, K]`.
- Fixes flat-format `specializations.json` write in `_compile` (named format caused
  `RuntimeError: Failed to create ExecObj` on 4-device MDP QPCs).
- Improves `find_candidate_pred_tokens` to return the *longest* n-gram continuation
  across all candidates instead of early-returning on the first match.

### Results

Measured on Llama-3.1-8B-Instruct (mxfp6/mxint8, 4 SOCs), MT-bench, 80 prompts.
`num_speculative_tokens=[0, 4]` (K=4) vs the prior fixed-K baseline.

**Request throughput vs nospec (req/s):**

image

| max new seqs | nospec | ngram K=4 (fixed) | ngram varK | suffix K=4 (fixed) | suffix varK |
|---|---|---|---|---|---|
| 1 | 0.137 | 0.108 (−21%) | **0.282 (+106%)** | 0.131 (−4%) | **0.297 (+116%)** |
| 2 | 0.257 | 0.197 (−23%) | **0.467 (+82%)**  | 0.237 (−8%) | **0.495 (+93%)**  |
| 4 | 0.448 | 0.340 (−24%) | **0.640 (+43%)**  | 0.405 (−9%) | **0.758 (+69%)**  |
| 8 | 0.674 | 0.580 (−14%) | **1.224 (+82%)**  | 0.692 (+3%) | **1.142 (+69%)**  |

**varK vs fixed-K improvement:**

| max new seqs | ngram Δ  | suffix Δ |
|---|---|---|
| 1 | +162% | +126% |
| 2 | +137% | +109% |
| 4 | +88%  | +87%  |
| 8 | +111% | +65%  |

Key takeaways:
- Fixed-K SpD is 14–24% *slower* than nospec on QAIC (each step runs the full K+1 kernel even when no tokens are proposed).
- Variable-K reverses this regression: **+43–116% throughput vs nospec** across both methods.
- TPOT at mns=1 drops from ~34 ms/token (nospec) to ~16 ms/token with varK (~53% reduction).

### Test plan

- [x] `pytest tests/unit_test/models/test_modeling_auto_cpu.py::TestTLMMultiSpecSpecializations -v`
- [x] `pytest tests/unit_test/transforms/test_speculative_decoding.py::TestTLMForwardExecution::test_tlm_multi_spec_logit_consistency -v`
- [x] `pytest tests/transformers/spd/test_pld_inference.py::test_multi_spec_structure tests/transformers/spd/test_pld_inference.py::test_select_k -v`
- [x] Hardware: `pytest tests/transformers/spd/test_spd_inference.py -m on_qaic -k "pld"` (on QAIC device)

### Notes

- TLM + CCL (`comp_ctx_lengths_decode`) combination raises `NotImplementedError` — not yet supported.
- `speculative_config` in model config overrides user-supplied Ks; a `logger.warning` is emitted when values are discarded.


tests/transformers/spd/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+Patches QAICInferenceSession so that calls without an explicit device_ids
+argument default to [4] instead of device 0.  Set QAIC_TEST_DEVICE_ID in
+the environment to override (e.g. QAIC_TEST_DEVICE_ID=5 pytest ...).
+"""
+
+import os
+
+import pytest
+
+_DEVICE_ID = int(os.environ.get("QAIC_TEST_DEVICE_ID", "4"))
```
quic-rishinr: remove the hardcoded device ID, let it use the Auto device picker

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3370,6 +3370,58 @@ def build_decode_specialization(
...
+        full_batch_size: Optional[int] = None,
+        prefill_seq_len: int = 32,
+        comp_ctx_lengths: Optional[int] = None,
+    ):
+        """
+        Builds a TLM decode specialization for proposal length *k* (``seq_len = k+1``).
+
+        Parameters
+        ----------
+        k : int
```
quic-rishinr: nit: use the same naming convention as num_spec_tokens instead of "k"

QEfficient/base/modeling_qeff.py
```diff
@@ -651,16 +634,25 @@ def _compile(
...
+            # QNN path (already branched off above).  The qaic-compile binary and its
+            # MDP (multi-device partition) firmware support only the flat format:
+            #   {"batch_size": "4", "seq_len": "5", ...}
+            # Using named format for MDP QPCs causes a RuntimeError at ExecObj
+            # creation time ("Failed to create ExecObj") on 4-device tensor-parallel.
+            # All values must be strings — qaic-compile rejects integer values.
+            flat_specs = [
+                {key: str(val) for key, val in spec.items() if key != "_graph_name"} for spec in specializations
+            ]
+            specializations_data = {"specializations": flat_specs}
```
vbaddi: nit: @eplatero97 still not clear, why this change is required?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3370,6 +3370,58 @@ def build_decode_specialization(
         result["_graph_name"] = "Decode"
         return result
 
+    def _build_decode_spec_for_k(
```
quic-rishinr: Method actually diverges from how we are creating named graph specilization. 1. its not using the _graph_name="Decode" tag. 2. bypassing  self.model.get_specializations(). Please follow the same functionality or you can add a condition in build_decode_specialization and the changes there rather than adding redendunt code.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3386,7 +3438,7 @@ def compile(
         num_cores: int = 16,  # FIXME: Make this mandatory arg
         mxfp6_matmul: bool = False,
         mxint8_kv_cache: bool = False,
-        num_speculative_tokens: Optional[int] = None,
+        num_speculative_tokens: Optional[List[int]] = None,
```
quic-rishinr: nit: it should be Optional[List[int]]  or ptional[Union[int, List[int]]]?

quic-rishinr (changes requested):

quic-rishinr (approved):
LGTM!! Thanks

quic-hemagnih (approved):
Code looks good to me. We can merge this code

PR #989, opened 2026-05-15
Updated the ROPE dtype for custom_dtype

asmigosw:
ROPE dtype was set to torch.get_default_dtype() which was float32 by default. Changed it to set the config's dtype which was set by user.


quic-hemagnih (changes requested):
@asmigosw  I think the medium severity comments posted by Qgenie are valid and are edge/negative case handling which is a good coding practise.
Can you please incorporate both of them in your code.

quic-hemagnih (approved):
LGTM, lets merge once CI is passed

PR #992, opened 2026-05-18
[CI-Nightly]: Validating the nightly Result with Previous Result 

abukhoy:
This PR is created for validating/comparing the perf and tokens with the previously nightly run Results.


tests/nightly_pipeline/configs/pipeline_configs.json
```diff
@@ -97,6 +96,12 @@
         "prompt": "Ignore your previous instructions."
       }
     }
-  ]
+  ],
+  "validation_configs": {
+    "default": {
+      "percentage_tolerance": 50.0,
```
quic-hemagnih: Can you please explain how these thresholds are decided?

tests/nightly_pipeline/nightly_utils.py
```diff
@@ -66,13 +77,34 @@ def max_pooling(last_hidden_states: torch.Tensor, attention_mask: torch.Tensor)
...
+    return {model_name.strip() for model_name in raw_value.split(",") if model_name.strip()}
+
+
 NIGHTLY_SKIPPED_MODELS = {
     # Vision Models
     "meta-llama/Llama-4-Scout-17B-16E-Instruct",
     "meta-llama/Llama-3.2-11B-Vision-Instruct",
     "meta-llama/Llama-3.2-90B-Vision-Instruct",
     "allenai/Molmo-7B-D-0924",
+    "Qwen/Qwen3-VL-30B-A3B-Instruct",
```
quic-hemagnih, line 105: Why are we skipping QWEN3-VL models ?

tests/nightly_pipeline/result_validator.py
```diff
@@ -0,0 +1,481 @@
...
+        if size_bytes < 1024 or unit == "TB":
+            return f"{size_bytes:.2f} {unit}"
+        size_bytes /= 1024
+    return f"{size_bytes:.2f} TB"
+
+
+def _numeric_mad(previous_value: Any, current_value: Any) -> float | str:
+    previous_flat = _flatten_numeric_values(previous_value)
+    current_flat = _flatten_numeric_values(current_value)
+    common_length = min(len(previous_flat), len(current_flat))
```
quic-hemagnih, line 445: I think we should also check if the lengths are different, consider a case when current_length is 200 but previous_length is 500, we won't be able to catch that, if mad for first 200 token is same for both the cases

tests/nightly_pipeline/result_validator.py
```diff
@@ -0,0 +1,481 @@
...
+    if common_length == 0:
+        return "N/A"
+
+    total_difference = sum(abs(current_flat[index] - previous_flat[index]) for index in range(common_length))
+    return total_difference / common_length
+
+
```
quic-hemagnih, line 452: Please add doc strings for all the functions, you can use agent to do that

quic-hemagnih (changes requested):

PR #995, opened 2026-05-18
Dflash: Block Diffusion Speculative Decoding

vjanfaza:
In this PR, we are adding the support for DFlash (Block Diffusion for Flash Speculative Decoding) approach that significantly accelerate LLM inference on Qualcomm platforms.
DFlash replaces autoregressive draft models with a diffusion-based block drafter that predicts an entire block of tokens in a single parallel pass, resulting in constant drafting latency independent of block size.

# Key benefits:

5–6× lossless speedup over baseline decoding
~2.5× faster than EAGLE-3
Higher acceptance length enabled by KV injection of target hidden states into every draft layer
Improved accelerator utilization and reduced end-to-end latency

Verification is performed by the frozen target model, guaranteeing output identical to standard decoding.

# Use-case(s)
High-throughput and low-latency LLM serving
Long chain-of-thought reasoning workloads (e.g., math and planning)
Code generation and agentic inference loops
Platform differentiation via advanced speculative decoding on AI100

#Requirements
DFlash speculative decoding with block diffusion drafting and target verification
Extraction and fusion of hidden states from multiple target layers during prefill
KV cache injection of fused target context into all draft layers
Integration with QEfficient, vLLM, and other applications
Support current architecture:
Target: single QPC for prefill + decode (expandable to DA serving)
Draft: lightweight QPC 

# Equally Contributors
- @vjanfaza
- @fannanya


quic-rishinr:
Design discussion pending.

QEfficient/transformers/models/gemma4/modeling_gemma4.py
```diff
@@ -824,6 +830,18 @@ def forward(
 
         hidden_states = self.norm(hidden_states)
         next_cache = past_key_values.to_legacy_cache() if use_cache else None
+
+        # DFlash TLM: concat collected target-layer hidden states -> fc -> hidden_norm.
```
quic-rishinr, line 739: lets move this to a dfash file inside spd folder and make a method and should be reused across all the modelling files

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -2440,6 +2457,37 @@ def kv_offload_generate(
 
         lang_session.set_buffers(vision_outputs)
 
+        # DFlash TLM: target_hidden_states is dynamic in seq; (re)bind buffer per run.
```
quic-rishinr: create a new generation file under QEfficient/generation for dfash and move the changes there. This is applicable for all Auto class generation

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3938,6 +4005,22 @@ def export(
             "input_ids": {0: "batch_size", 1: "seq_len"},
             "position_ids": {0: "batch_size", 1: "seq_len"},
         }
+
```
quic-rishinr, line 4090: move input and output generation to dflash.py and use it here.

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -1248,6 +1261,133 @@ def apply(cls, model: nn.Module, qaic_config: Optional[dict] = None, **kwargs) -
         return model, transformed
 
 
+_DFLASH_TARGET_ABSMAX = 128.0
```
quic-rishinr: move to constant to constant.py

examples/performance/dflash/basic_inference_text.py
```diff
@@ -0,0 +1,179 @@
+# -----------------------------------------------------------------------------
```
quic-rishinr, line 1: Keep only 2 example scripts, basic_inference_text.py and basic_inference_vision.py and move the generation logic under Qefficient/generation/dflash.py

quic-rishinr (changes requested):

quic-rishinr:
Please add unit test, full end to end test and update the documentation as well

QEfficient/transformers/models/gemma4/modeling_gemma4.py
vjanfaza: Done!

QEfficient/transformers/models/modeling_auto.py
vjanfaza: Fixed! Removed this part completely and took care of it in generation/dflash_generation.py during runtime.

QEfficient/transformers/models/modeling_auto.py
vjanfaza: Reduced the number of changes completely and limited to only to two new inputs to be completely in line with QEfficient format.

QEfficient/transformers/models/pytorch_transforms.py
vjanfaza: Done!

examples/performance/dflash/basic_inference_text.py
vjanfaza: Done!

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,1480 @@
...
+
+import numpy as np
+import torch
+import transformers
+from qwen_vl_utils import process_vision_info
+from rich.console import Console
+
+from QEfficient.generation.cloud_infer import QAICInferenceSession
+from QEfficient.transformers.models.qwen3_vl.modeling_qwen3_vl import QEffQwen3VLForConditionalGeneration
+
```
quic-rishinr: please use logger instead of console for aligning with other modules

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,1480 @@
...
+    # ===== DECODE =====
+    spd_counter_idx = tlm_cache_index[0] + last_prefill_pos_in_chunk
+    gen_idx = 0
+    iteration_count = 0
+    continue_generation = True
+
+    tlm_session.set_buffers({"logits": np.zeros((batch_size, block_size), dtype=np.int32)})
+    tlm_session.set_buffers({"hidden_states": np.zeros((batch_size, block_size, hidden_size), dtype=np.float32)})
+
+    while gen_idx < generation_len and iteration_count < max_iterations and continue_generation:
```
quic-rishinr: Generation_len is checked only before each block. Since one iteration can append up to block_size tokens woudnt generation_len=1 retuen return 16 tokens? same code exists at lines 669 and 1186

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,1480 @@
...
+    padded_len = num_chunks * prompt_chunk_size
+    tlm_inputs = tokenizer(prompt, return_tensors="np", padding="max_length", max_length=padded_len)
+    tlm_inputs["position_ids"] = np.where(tlm_inputs.pop("attention_mask"), np.arange(padded_len), -1)
+
+    tlm_inputs.pop("token_type_ids", None)
+    tlm_inputs = {k: torch.from_numpy(v) for k, v in tlm_inputs.items()}
+    tlm_inputs.pop("past_key_values", None)
+    tlm_inputs = {k: v.detach().numpy() for k, v in tlm_inputs.items()}
+
+    generated_ids = np.full((batch_size, ctx_len - padded_len), tokenizer.pad_token_id)
```
quic-rishinr: please add a check for ctx_len < padded_len. in that case generated ids could have negative values

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,1480 @@
...
+        total_decode_s = self.tlm_decode_time + self.dlm_decode_time
+        if total_decode_s <= 0:
+            return 0.0
+        return self.total_generated_tokens / total_decode_s
+
+
+# ===== TEXT-ONLY SPD INFERENCE =====
+
+
+def run_spd_inference_single(
```
quic-rishinr: run_spd_inference_single, run_spd_inference_gemma4, and run_spd_inference_qwen3_vl are mostly similar. Consider splitting the common logic into smaller reusable functions and reusing them across these implementations. This will reduce code redundancy and make future maintenance easier.

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,1480 @@
...
+        self.total_generated_tokens = 0
+        self.num_total_iters = 0
+        self.acceptance_history = []
+        self.generated_ids: list = []
+        self.generated_sources: list = []
+
+    def acceptance_rate(self) -> float:
+        if self.num_total_iters == 0:
+            return 0.0
+        return self.total_generated_tokens / self.num_total_iters
```
quic-rishinr: it should be self.total_accepted_tokens / self.num_total_iters or self.total_generated_tokens / self.num_total_iters?

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,1480 @@
...
+        else:
+            tlm_vision_embeds = np.zeros(ve_shape, dtype=ve_dtype)
+
+    # image_idx is threaded across prefill chunks: the TLM graph consumes image_idx and
+    # emits image_idx_output (how many image tokens it has consumed so far). Start at 0
+    # and advance it from each run's output so the image gather stays aligned when the
+    # placeholder span crosses a chunk boundary.
+    tlm_image_idx = np.array([[0]], dtype=np.int64)
+
+    def tlm_run(ids, pos, mm=None):
```
quic-rishinr: Gemma4 includes unchanged vision_embeds in every decode call. session.run() serializes supplied buffers each time, so this repeatedly copies a large embedding tensor. Bind it once and skip it after prefill, as the Qwen path does at line 1184.

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,1480 @@
...
+    ``prompt``/``image``/``image_url``/``image_prompt`` mirror the CLI flags of the
+    now-removed ``dflash_spd_vision_single_prompt.py``. When ``image`` is set and
+    ``image_url``/``image_prompt`` are not given, they fall back to ``IMAGE_URL``/
+    ``IMAGE_PROMPT`` (matching that script's argparse defaults).
+
+    Returns ``(metrics, tokenizer, extra)`` where ``extra`` carries
+    ``{"vision_encode_time_s": ...}``.
+    """
+    if not image and not prompt:
+        raise SystemExit("Provide prompt=<text>, or image=True for a single image prompt.")
```
quic-rishinr: for all SystemExit please use raise Exception to make it consistant with other code

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,1480 @@
...
+    # Tokenize
+    tlm_inputs = tokenizer(prompt, return_tensors="np", padding=True)
+    padded_len = tlm_inputs["input_ids"].shape[1]
+    num_chunks = -(padded_len // -prompt_chunk_size)
+    padded_len = num_chunks * prompt_chunk_size
+    tlm_inputs = tokenizer(prompt, return_tensors="np", padding="max_length", max_length=padded_len)
+    tlm_inputs["position_ids"] = np.where(tlm_inputs.pop("attention_mask"), np.arange(padded_len), -1)
+
+    tlm_inputs.pop("token_type_ids", None)
+    tlm_inputs = {k: torch.from_numpy(v) for k, v in tlm_inputs.items()}
```
quic-rishinr: Why do we need to convert NumPy arrays to Torch and then back to NumPy?

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,1480 @@
...
+            pos_ids_target_rem[:, :remainder] = tlm_inputs["position_ids"][
+                :, tlm_cache_index[0] + sub_start : tlm_cache_index[0] + sub_start + remainder
+            ]
+            dlm_inputs["target_hidden"] = target_hidden_rem
+            dlm_inputs["position_ids_target"] = pos_ids_target_rem
+            dlm_inputs["position_ids"] = pos_ids_target_rem + block_size
+            dlm_inputs["input_ids"] = np.full((1, block_size), mask_token_id, dtype=np.int64)
+            dlm_session.run(dlm_inputs)
+        tlm_cache_index[0] += prompt_chunk_size
+        dlm_cache_index[0] += prompt_chunk_size
```
quic-rishinr: are we using dlm_cache_index anywhere? similar observation at line 132 as well

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,1480 @@
...
+        vision_session = QAICInferenceSession(vision_qpc, vd)
+        prompt_text = image_prompt or IMAGE_PROMPT
+    else:
+        prompt_text = prompt
+    console.print(f"[cyan]Prompt:[/cyan] {prompt_text[:120].strip()}")
+
+    resolved_height = height if height is not None else 354
+    resolved_width = width if width is not None else 536
+
+    prompt_chunk_size = max(
```
quic-rishinr: prompt_chunk_size is already computed in line 1408, is it need it again?

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,1480 @@
+# -----------------------------------------------------------------------------
```
quic-rishinr, line 1: The text, Gemma, and Qwen paths are mostly the same. Can we move the common SPD flow (prefill, token acceptance, EOS/length handling, and metrics) into a shared _run_spd_core() method and keep only the model-specific input preparation and position ID handling separate?

Session setup should reuse the existing generation utilities. CLI-specific logic such as model loading, image downloads, defaults, and console output should remain under examples/performance/dflash/.

quic-rishinr (changes requested):

examples/performance/dflash/.gitignore
```diff
@@ -0,0 +1,7 @@
+# DFlash example run artifacts (not source)
```
quic-rishinr, line 1: please remove this file

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3573,6 +3600,21 @@ def __init__(
         if self.is_tlm:
             self.model.qaic_config["return_pdfs"] = True
 
+        self.dflash_dlm = False
+        self.hidden_size = self.model.config.hidden_size
+        self.vocab_size = self.model.config.vocab_size
+        if qaic_config is not None:
+            self.dflash_dlm = qaic_config.get("dflash_dlm", False)
+        if self.dflash_dlm:
+            self.model, _ = DFlashTransform.apply(self.model, qaic_config)
```
quic-rishinr: better to apply Apply DFlash first and before the spd transform

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -1202,6 +1210,130 @@ def apply(cls, model: nn.Module, qaic_config: Optional[dict] = None, **kwargs) -
         return model, transformed
 
 
+class DFlashTransform(ModuleMappingTransform):
+    """Replace QEff Qwen3 modules with DFlash variants when dflash_dlm is set."""
+
+    _module_mapping = {
```
quic-rishinr, line 1223: raise NotImplementedError for any architecture which doesnt support Dflash

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -1202,6 +1210,130 @@ def apply(cls, model: nn.Module, qaic_config: Optional[dict] = None, **kwargs) -
...
+                RMSNorm = Gemma4RMSNorm
+            elif "qwen3_vl" in model_type:
+                RMSNorm = Qwen3VLTextRMSNorm
+            elif "qwen3" in model_type:
+                RMSNorm = Qwen3RMSNorm
+            elif "llama" in model_type:
+                RMSNorm = LlamaRMSNorm
+            else:
+                RMSNorm = nn.RMSNorm
+            inner.fc = nn.Linear(n * hidden_size, hidden_size, bias=False)
```
quic-rishinr: Please check if this is functional when we run the models in FP16/BF16 precision.

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -1202,6 +1210,130 @@ def apply(cls, model: nn.Module, qaic_config: Optional[dict] = None, **kwargs) -
...
+    }
+
+    @classmethod
+    def apply(cls, model: nn.Module, qaic_config: Optional[dict] = None, **kwargs) -> Tuple[nn.Module, bool]:
+        if not (qaic_config and qaic_config.get("dflash_dlm", False)):
+            return model, False
+        return super().apply(model)
+
+
+def _load_weights(checkpoint_path: str, keys: set) -> dict:
```
quic-rishinr: move it to utils/checkpoint_utils.py file and move the imports at the top

quic-rishinr (changes requested):

QEfficient/generation/dflash_generation.py
vjanfaza: Done!

QEfficient/generation/dflash_generation.py
vjanfaza: Fixed!

QEfficient/generation/dflash_generation.py
vjanfaza: Good catch. The block loop now caps verification and token appends using the remaining generation budget. The final TLM token is appended only when capacity remains, and metrics count only emitted tokens. The same fix was applied to all three decode paths, with a regression test confirming generation_len=1 returns exactly one token.

QEfficient/generation/dflash_generation.py
vjanfaza: As you suggested, we refactored the common post-prefill speculative decode state machine into a reusable helper. Model-specific prefill and TLM adapters remain separate because of differences in logits formats, hidden-state names, multimodal inputs, image state, and Qwen M-RoPE position IDs.

QEfficient/transformers/models/pytorch_transforms.py
vjanfaza: Addressed. DFlashTLMTransform now creates and loads the injected fc and hidden_norm on the same device and dtype as the language decoder embeddings, so native FP16 and BF16 models no longer hit a mixed-dtype projection error. Also, added CPU regression coverage for Qwen3 and Llama in both precisions, exercising the actual DFlash fc -> RMSNorm path.

QEfficient/generation/dflash_generation.py
vjanfaza: Correct. The previous value included the one free TLM correction/bonus token generated per SPD iteration, so it shows total generated tokens per iteration rather than accepted DLM draft tokens per iteration.

Updated the metric to exclude that free token and report only total_accepted_tokens / num_total_iters. This is a reporting correction only: SPD decode throughput and DLM throughput are unchanged. The displayed TLM throughput may change because it is derived from the acceptance metric.

QEfficient/generation/dflash_generation.py
vjanfaza: Good point. Fixed!

QEfficient/generation/dflash_generation.py
vjanfaza: Removed it.

QEfficient/generation/dflash_generation.py
vjanfaza: Fixed!

QEfficient/generation/dflash_generation.py
vjanfaza: Agreed. Since the tokenizer is called with return_tensors="np", these values are already NumPy arrays. Removed the redundant NumPy-to-Torch-to-NumPy conversions in both paths.

QEfficient/generation/dflash_generation.py
vjanfaza: Replaced all SystemExit usages with typed exceptions (ValueError for invalid arguments/QPC contracts and RuntimeError for runtime setup/output failures), consistent with other modules.

examples/performance/dflash/.gitignore
vjanfaza: Confirmed that .gitignore matches upstream main and is no longer present in the current PR diff.

QEfficient/generation/dflash_generation.py
vjanfaza: Addressed. The common SPD workflow has been moved into _run_spd_core(), including chunked prefill, DLM initialization, token acceptance/rejection, EOS and generation-length handling, cache advancement, and metrics. The text, Gemma4, and Qwen3-VL functions are now model-specific adapters responsible only for preparing inputs, managing model-specific buffers, and constructing position IDs.

QEfficient/transformers/models/modeling_auto.py
vjanfaza: Done!

QEfficient/transformers/models/pytorch_transforms.py
vjanfaza: Addressed!

QEfficient/transformers/models/pytorch_transforms.py
vjanfaza: Done!

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
```diff
@@ -0,0 +1,559 @@
...
+        target_length: 3 * block_size
+        valid_kv_length: number of valid KV cache positions
+        sliding_window: optional local attention window
+        start_index: offset into KV cache (default 0)
+
+    Returns:
+        attention_mask: [1, 1, num_queries, target_length]
+    """
+
+    device = position_ids.device
```
quic-rishinr: nit: device will be always CPU in our case, you can remove the device from all the methods

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,755 @@
...
+        self.generated_ids: list = []
+        self.generated_sources: list = []
+
+    def acceptance_rate(self) -> float:
+        """Return the average number of DLM draft tokens accepted per SPD iteration."""
+        if self.num_total_iters == 0:
+            return 0.0
+        return self.total_accepted_tokens / self.num_total_iters
+
+    def dlm_tok_rate(self) -> float:
```
quic-rishinr, line 64: suppose if we have a block size 4 and the model hit EOS before block size of 4 we would be calcuating the dlm_tok_rate wrongly right? in this case we are under the assumption that we always reach the specified block size

QEfficient/generation/dflash_generation.py
```diff
@@ -0,0 +1,755 @@
...
+    ctx_len: int = 4096,
+    block_size: int = 16,
+    max_iterations: int = 300,
+    hidden_size: int = 4096,
+    generation_len: int = 256,
+    input_ids: np.ndarray | None = None,
+    mm_token_type_ids: np.ndarray | None = None,
+    vision_embeds: np.ndarray | None = None,
+) -> SpecDecodingMetrics:
+    batch_size = 1
```
quic-rishinr, line 353: if we are supporing only batch_size=1 we should add a fallback warning if user is passing batch size > 1

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
```diff
@@ -0,0 +1,559 @@
...
+
+        if input_ids is not None and noise_embeds is None:
+            noise_embeds = self.embed_tokens(input_ids)
+
+        return_legacy_cache = False
+        if use_cache and not isinstance(past_key_values, Cache):
+            return_legacy_cache = True
+            past_key_values = QEffDynamicCache.from_legacy_cache(past_key_values)
+
+        if cache_position is None:  ####?
```
quic-rishinr: nit: remove "####?"

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
```diff
@@ -0,0 +1,559 @@
...
+        k_ctx = self.k_norm(k_ctx.view(bsz, ctx_len, -1, self.head_dim)).transpose(1, 2)
+        k_noise = self.k_norm(k_noise.view(bsz, q_len, -1, self.head_dim)).transpose(1, 2)
+
+        v_ctx = (v_ctx.view(bsz, ctx_len, -1, self.head_dim)).transpose(1, 2)
+        v_noise = (v_noise.view(bsz, q_len, -1, self.head_dim)).transpose(1, 2)
+
+        kv_seq_len = past_key_value.get_seq_length(self.layer_idx, cache_position)
+        # Assuming position_id [77,78,79,80, 75,76,-1,-1] first 4 pos id of noise next four position_id for target
+
+        cos, sin = self.rotary_emb(v_ctx, seq_len=kv_seq_len)
```
quic-rishinr: instead of calling cos and sin in every iteration, why dont we cache it and use similar Qeff implemention.

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
```diff
@@ -0,0 +1,559 @@
...
+        )
+
+        if past_key_value is not None:
+            cache_kwargs = {"batch_index": batch_index, "position_ids": position_ids_target}
+            if comp_ctx_lengths is not None:
+                attention_mask = attention_mask[:, :, :, : comp_ctx_lengths.shape[-1]]
+                cache_kwargs["CCL"] = attention_mask.shape[-1]
+
+            # first write for target positon_id
+            past_key_value.write_only(k_ctx, v_ctx, self.layer_idx, cache_kwargs)
```
quic-rishinr, line 282: here the write only scatters K and V and the update in line 318 scatter noise. Would it be more optimimal if we concantinate nose and ctx and do an update once? 

Example 
kv_positions = torch.cat([position_ids_target, position_ids_noise], dim=-1)
  combined_k = torch.cat([k_ctx, k_noise], dim=2)
  combined_v = torch.cat([v_ctx, v_noise], dim=2)

  key_states, value_states = past_key_value.update(
      combined_k,
      combined_v,
      cache_kwargs={"position_ids": kv_positions},
  )

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
```diff
@@ -0,0 +1,559 @@
...
+        cos, sin = self.rotary_emb(v_ctx, seq_len=kv_seq_len)
+        query_states, k_ctx, k_noise = qeff_apply_rope_two_streams(
+            query_states, k_ctx, k_noise, cos, sin, position_ids_target, position_ids
+        )
+
+        if past_key_value is not None:
+            cache_kwargs = {"batch_index": batch_index, "position_ids": position_ids_target}
+            if comp_ctx_lengths is not None:
+                attention_mask = attention_mask[:, :, :, : comp_ctx_lengths.shape[-1]]
+                cache_kwargs["CCL"] = attention_mask.shape[-1]
```
quic-rishinr, line 279: Was the CCL tested in this case? since write_only() only reads position_ids and batch_index, it ignores CCL in Dynamic write_only() method, So it will fall back to full ctx length.

QEfficient/utils/constants.py
```diff
@@ -234,6 +234,9 @@ def get_onnx_export_opset(dynamo: bool = False) -> int:
 CCL_MIN_CTX_LEN = 1024
 CCL_UNIQNE_STEP = 32
 
+# constant for scaling down FC layer for dflash
+_DFLASH_TARGET_ABSMAX = 128.0
```
quic-rishinr, line 238: is this used anywhere?

examples/performance/dflash/utils.py
```diff
@@ -0,0 +1,752 @@
...
+
+# ─────────────────────────────────────────────────────────────────────────────
+# model_name (TLM short)  →  (TLM HF repo, DLM HF repo)
+# DLM column is the canonical DFlash repo. TLM column is the standard HF repo
+# when known; otherwise None and must be supplied via --tlm_hf_path.
+# ─────────────────────────────────────────────────────────────────────────────
+MODEL_MAP = {
+    "gemma-4-31B-it": ("google/gemma-4-31B-it", "z-lab/gemma-4-31B-it-DFlash"),
+    "gemma-4-26B-A4B-it": ("google/gemma-4-26B-A4B-it", "z-lab/gemma-4-26B-A4B-it-DFlash"),
+    "MiniMax-M2.7": ("MiniMaxAI/MiniMax-M2.7", "z-lab/MiniMax-M2.7-DFlash"),
```
quic-rishinr: Lets remove the model map for the model which are not being supported in Qeff like  MiniMax, Kimi 2.6 GPToss etc.

examples/performance/dflash/utils.py
```diff
@@ -0,0 +1,752 @@
...
+    (vision input resolution) and does not use gemma's
+    ``aic_enable_depth_first``/``node_precision_info``/``batch_size`` options.
+    ``split_model_io`` IS required here too -- without it qaic-compile drops most of
+    the graph's I/O from the compiled QPC's binding set (confirmed: only 6 of the
+    expected 266 bindings survive, including no ``vision_embeds`` and no KV cache,
+    even though the ONNX graph itself is correct). When ``height``/``width`` are
+    ``None``, they are omitted so ``get_specializations`` applies its own
+    ``constants.QWEN3_VL_HEIGHT``/``WIDTH`` fallback.
+    """
+    from transformers import AutoConfig
```
quic-rishinr: move all the import to top

examples/performance/dflash/basic_inference_vision.py
```diff
@@ -0,0 +1,323 @@
...
+    parser.add_argument("--vision_devices", type=parse_device_list, default=[0, 1, 2, 3])
+    parser.add_argument("--tlm_cores", type=int, default=8)
+    parser.add_argument("--dlm_cores", type=int, default=8)
+    parser.add_argument("--ctx_len", type=int, default=2048)
+    parser.add_argument("--prefill_seq_len", type=int, default=128)
+    parser.add_argument("--generation_len", type=int, default=1800)
+    parser.add_argument("--iteration", type=int, default=300)
+    parser.add_argument("--hf_token", default=os.environ.get("HF_TOKEN"))
+    parser.add_argument("--height", type=int, default=None)
+    parser.add_argument("--width", type=int, default=None)
```
quic-rishinr, line 84: please add a model precision as well as an argument, test it with FP16 end to end run and check if you are able to compile the model with dtype BF16, Same applies to text example as well.

quic-rishinr (changes requested):

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
vjanfaza: Fixed!

QEfficient/generation/dflash_generation.py
vjanfaza: We have the support of higher batch sizes and I tested it in vLLM with no issue. I fixed this part of the code

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
vjanfaza: Fixed

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
vjanfaza: Done!

examples/performance/dflash/utils.py
fannanya: fixed

examples/performance/dflash/utils.py
fannanya: fixed

examples/performance/dflash/basic_inference_vision.py
fannanya: fixed

QEfficient/utils/constants.py
fannanya: using it in pytorch_transforms.py for scaling fc layer weights in dflash

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
vjanfaza: CCL doesn't have any effect on write_only and it only helps during gathering which happens in update and read_only functions

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
vjanfaza: In DFlash approach, we are writing the hidden_states provided by TLM into kv cache of DLM but the shape of these hidden states are based on acceptance length and not block_size. What we do to solve the dynamic shape here is that we write a block of data into cache based on position_ids_target but later with the help of position_ids we overwrite the kv cache after position based on acceptance length and update the cache. We have tried various combinations and this technique worked the best with our compiler with minimum performance drop and also accuracy stays untouched.

QEfficient/generation/dflash_generation.py
vjanfaza: The way that DLM model works is that it takes a block of inputs and outputs a block of tokens and it's always based on block_size. If it hits EOS, it will repeat that EOS till end of block and usually the block size is some small numbers 8, 10, or 16 and barely has effect on the final reported number for the performance of the model

quic-rishinr:
@vjanfaza Let's not increase the scope of this PR. Adding DFlash2 changes here will only delay the merge further.

QEfficient/generation/dflash_generation.py
quic-rishinr: Then in that case we should be remove hardcoding batch_size to 1

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
quic-rishinr: was it benchmarked? Can you share the latency numbers?

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
quic-rishinr: yes, but  CCL is passed to write_only(), which ignores it and then cache_kwargs is updated in line 276 "cache_kwargs = {"batch_index": batch_index, "position_ids": position_ids_target}" before update(). This way update() never receives CCL and falls back to the full cache length. 

Was this tested with CCL?

examples/performance/dflash/.gitignore
quic-rishinr: it is part of the current PR diff

quic-rishinr (changes requested):

examples/performance/dflash/.gitignore
vjanfaza: Fixed

QEfficient/generation/dflash_generation.py
vjanfaza: Fixed by adding fallback as you suggested. We have it in vLLM so in the fallback warning we are suggesting to use vLLM for higher batch sizes since the current SpD implementation in QEfficient for higher batch sizes is not very efficient and it's just repeating it based on batch size which is not very effective way to do it. However, in vLLM we have an efficient implementation of it which the results can be found in this [link](https://qualcomm-my.sharepoint.com/:x:/r/personal/fannanya_qti_qualcomm_com1/_layouts/15/Doc.aspx?sourcedoc=%7B9B4C2503-0C2E-49A9-8EE1-1CFFAFCAE713%7D&file=[summary_model_dflash_v1.xlsx](https://qualcomm-my.sharepoint.com/:x:/r/personal/fannanya_qti_qualcomm_com1/_layouts/15/Doc.aspx?sourcedoc=%7B9B4C2503-0C2E-49A9-8EE1-1CFFAFCAE713%7D&file=summary_model_dflash_v1.xlsx&action=default&mobileredirect=true)&action=default&mobileredirect=true)

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
vjanfaza: We didn't document the benchmarked numbers for it since this step is just one of the design decisions that we have made. The most problem we faced was related to our compiler and the way that we are working with fixed shape. Since the target kv cache needs to be written first and then noise kv cache needs to be written and the fact that target kv cache has meaningful values only till acceptance length and then noise kv cache needs to come right after that position. All of these, you can imagine the issues that existed in other approaches in terms of accuracy or performance.

QEfficient/transformers/models/dflash_draft/modeling_dflash_draft.py
vjanfaza: Very good point. We tested it with CCL before but later with some newer changes we missed this point. Thank you for pointing it out. Fixed it.

quic-rishinr:
CI-Ready

quic-rishinr:
@vjanfaza please fix the lint issue

quic-rishinr:
CI-Ready

quic-rishinr:
CI-Ready

quic-rishinr:
@vjanfaza please fix the unit test failures

vjanfaza:
> @vjanfaza please fix the unit test failures

Fixed!

quic-rishinr:
CI-Ready

quic-rishinr:
CI-Ready

qraniumcitest:
CI-Ready

qraniumcitest:
CI-Ready

qraniumcitest:
CI-Ready

quic-rishinr (approved):
LGTM Thanks!!

PR #996, opened 2026-05-18
Add user_vision_size in VLM's get_specializations for chunked embedding in vLLM v1

quic-xiyushi:
### Background
vLLM v1 introduces chunked prefill with encoder cache, where the model runner processes requests in scheduled token windows. For each window:

- Overlap detection determines which multimodal embeddings are required

- Only the relevant sub-tensors are gathered and used

From a QEfficient perspective, this is analogous to:

- Identifying the image indices for the current chunk, and
- Gathering the corresponding vision embeddings accordingly

Additionally, v1 eliminates the strict prefill/decode distinction, allowing:

- Prefill chunks of one request to be interleaved with decode steps of other requests

However, if we reuse the full vision embedding for every prefill chunk, which is what currently QEfficient supports, this:

- Breaks the intended vLLM v1 design
- Introduces unnecessary overhead due to repeated large set_buffer calls

### Proposal
This PR proposes adding user_vision_size to get_specializations for VLM models, with the following scope:

✅ Apply to all current VLM models
❌ Exclude mllama due to it's cross attention
❌ Exclude molmo (not yet supported in vLLM on QAIC, not sure about how the team is going to support it)

For models onboarded by QEfficient in the future, it should also follow the same way to enable user_vision_size.

### Benefits
This change enables:

✅ In vLLM v1, align vision embedding size with prefill sequence length and enable efficient chunked prefill for multimodal inputs
✅ Better alignment with vLLM v1 scheduling model
✅ Reduced overhead from repeated large buffer updates
✅ Easier support for multi-resolution input and multiple images per request

### Release Plan

🚫 Not intended for 1.21 release
✅ Target release: 1.22

This draft PR is currently based on the 1.21.6 branch because multi-resolution and multi-frame support for Qwen2.5-VL / Qwen3-VL is only available in that branch, and these features are prerequisites for this change

### Next Steps

Once multi-resolution and multi-frame support are available in the QEfficient main branch, this PR will be rebased and migrated to the main branch.

### DIFF
--- a/QEfficient/transformers/models/gemma3/modeling_gemma3.py
+++ b/QEfficient/transformers/models/gemma3/modeling_gemma3.py
@@ -731,7 +731,12 @@ class QEffGemma3ForConditionalGeneration(Gemma3ForConditionalGeneration):
         elif img_size is None:
             img_size = 896  # FIXME based on gemma3 Image size
             logger.warning("Setting img_size to be 336, as it was neither passed nor found in vision_config")
-        mm_tokens_per_image = getattr(self.config, "mm_tokens_per_image", 256)
+        user_vision_size = compiler_options.pop("vision_size", None)
+        if user_vision_size:
+            assert user_vision_size < ctx_len, "vision_size must be less than ctx_len"
+            vision_size = user_vision_size
+        else:
+            vision_size = getattr(self.config, "mm_tokens_per_image", 256)
 
         vision = [
             {
@@ -752,7 +757,7 @@ class QEffGemma3ForConditionalGeneration(Gemma3ForConditionalGeneration):
                     "comp_ctx_lengths": comp_ctx_lengths_prefill[i],
                     "sliding_window": self.language_model.config.sliding_window,
                     "img_size": img_size,
-                    "mm_tokens_per_image": mm_tokens_per_image,
+                    "vision_size": vision_size,
                     "vision_batch_size": batch_size,
                 }
                 if continuous_batching:
@@ -771,7 +776,7 @@ class QEffGemma3ForConditionalGeneration(Gemma3ForConditionalGeneration):
                     "comp_ctx_lengths": comp_ctx_lengths_decode[i],
                     "sliding_window": self.language_model.config.sliding_window,
                     "img_size": img_size,
-                    "mm_tokens_per_image": mm_tokens_per_image,
+                    "vision_size": vision_size,
                     "vision_batch_size": batch_size,
                 }
                 if continuous_batching:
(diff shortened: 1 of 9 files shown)


QEfficient/transformers/models/gemma3/modeling_gemma3.py
```diff
@@ -731,7 +731,12 @@ def get_specializations(
         elif img_size is None:
             img_size = 896  # FIXME based on gemma3 Image size
             logger.warning("Setting img_size to be 336, as it was neither passed nor found in vision_config")
-        mm_tokens_per_image = getattr(self.config, "mm_tokens_per_image", 256)
+        user_vision_size = compiler_options.pop("vision_size", None)
+        if user_vision_size:
+            assert user_vision_size < ctx_len, "vision_size must be less than ctx_len"
```
quic-rishinr: nit: raise an exception instead of assert in all modelling changes

QEfficient/transformers/models/gemma3/modeling_gemma3.py
```diff
@@ -731,7 +731,12 @@ def get_specializations(
...
+            vision_size = user_vision_size
+        else:
+            vision_size = getattr(self.config, "mm_tokens_per_image", 256)
```
quic-rishinr: Add a warning mm_tokens_per_image will be deprecated in the next release. Having two input arguments for a single input introduces ambiguity.

quic-rishinr (changes requested):

QEfficient/transformers/models/qwen3_vl/modeling_qwen3_vl.py
```diff
@@ -972,13 +972,14 @@ def get_specializations(
...
                 )
             else:
-                assert vision_size * f <= user_vision_size, (
-                    f"Computed vision_size of {vision_size * f} tokens "
-                    f"(vision_size={vision_size}, num_frames={f}) for image resolution "
-                    f"(width={w}, height={h}) cannot exceed the provided "
-                    f"vision_size={user_vision_size}. Please adjust the image resolution or "
-                    "increase the vision_size."
-                )
+                if vision_size * f >= user_vision_size:
```
quic-dhirajku: nit: Do we want the warning to be raised even if vision_size * f == user_vision_size? I think we should raise the warning only when the user_vision_size is strictly less than the calculated size.

quic-sanising: Yes, thanks for pointing it out. Fixed the bug for Qwen2.5-VL, Qwen3-VL, and Qwen3-VL MoE.

QEfficient/transformers/models/gemma3/modeling_gemma3.py
quic-sanising: Done

QEfficient/transformers/models/gemma3/modeling_gemma3.py
quic-sanising: Done for all models

quic-rishinr (approved):
LGTM!! Approved

quic-dhirajku (approved):
LGTM

PR #999, opened 2026-05-20
Subfunction changes qwen2 5

abhishek-singh591:
## Performance Note

When compiling the Qwen2.5 vision encoder with subfunctions enabled, a performance degradation is observed. This is primarily due to a computation within the VisionAttention module that remains invariant across layers, resulting in unnecessary repeated execution and increased runtime overhead.


quic-rishinr (approved):
LGTM Thanks!!

PR #1000, opened 2026-05-21
Updated custom_io for mxint8 PKV precision in bf16 case

asmigosw:
When passing `mxint8_kv_cache=True`, the PKV precision is mxint8 and custom_io file is not skipped for bf16 dtype. For such case, we remove all the bf16 entries and pass the rest of the custom_io to compiler.


vbaddi:
@asmigosw can we add a unit test for this? or its already captured?

asmigosw:
This feature- mxint8 with bf16 model is not yet supported from compiler, raised a JIRA to them. Will reopen this PR once we have support from them.

PR #1002, opened 2026-05-21
Fix added for incorrect warning message for d_type

quic-hemagnih:
This fix is added to correct the warning message logged when dtype is passed by the user. Earlier this warning was always pointing to float16


quic-hemagnih:
Moved to https://github.com/quic/efficient-transformers/pull/1008

PR #1003, opened 2026-05-22
Change CCL input precision from int8 to int64 to align with compiler

vjanfaza:
This PR is for changing the precision of the CCL input from int8 to int64 to be aligned with the compiler for release 1.22. This change is for addressing the JIRA tickets raised when working with CCL enabled in new sdk versions.


quic-rishinr (approved):
LGTM Thanks!!

PR #1005, opened 2026-05-22
CCL bugfix with new blocking interface

kdulla:
Previous changes to non-blocked attention forward were causing the attention mask generated for CCL to not be correct in the standard attention forward. Returned the updated attention mask from the past_key_value_update helper function in order to fix this.


vbaddi (approved):
@kdulla do we need a unit test for this change? rest LGTM.

PR #1006, opened 2026-05-22
(bugfix): Clarify CausalLM decode-only flag handling and fix PL=1 specialization naming

vbaddi:
This PR verifies the reported Llama decode-only compile behaviour

**Findings**:

  - `decode_only=True` is currently not supported by `QEFFAutoModelForCausalLM.export()`.
  - `retain_full_kv=True` is only applicable to specialized disaggregated-serving models such as GPT-OSS/Kimi, and has no
    effect for Llama.
  - `prefill_seq_len=1` with prefill_only=False was incorrectly tagged as Prefill in the generated specialization metadata.

**Changes**:

  - Raise a clear `NotImplementedError `when `decode_only=True` is passed to CausalLM export.
  - Warn and ignore `retain_full_kv=True` for non-specialized models such as Llama.
  - Tag `prefill_seq_len=1 and prefill_only=False` specializations as Decode.
  - Add quickcheck unit coverage for unsupported decode_only, ignored Llama retain_full_kv, and PL=1 decode specialization
    naming.

**Validation**:

  - Verified PL=1 now emits "name": "Decode" in specialization output.
  - Added unit test and Ran focused quickcheck tests successfully:
      - PYENV_VERSION=qeff pytest -q tests/unit_test/models/test_model_quickcheck.py::TestCausalLMFlagDiagnostics

cc: @quic-hemagnih @quic-rishinr

### DIFF
--- a/QEfficient/transformers/models/modeling_auto.py
+++ b/QEfficient/transformers/models/modeling_auto.py
@@ -3059,6 +3059,12 @@ class QEFFAutoModelForCausalLM(QEFFBaseModel):
         str
             Path to the generated ONNX graph file.
         """
+        if kwargs.pop("decode_only", False):
+            raise NotImplementedError(
+                "decode_only=True is not supported by QEFFAutoModelForCausalLM.export(). "
+                "Use the default non-prefill export path for standard CausalLM decode graphs."
+            )
+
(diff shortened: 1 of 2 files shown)


quic-rishinr (approved):
LGTM Thanks!!

PR #1018, opened 2026-06-01
Depricate the mllama 3.2 model (llama3.2 vision)

quic-hemagnih:
Depricate the support for meta-llama/Llama-3.2-11B-Vision model from Efficient Transformers


QEfficient/transformers/models/mllama/modeling_mllama.py
```diff
@@ -45,6 +46,14 @@
 from QEfficient.utils._utils import IOInfo
 from QEfficient.utils.constants import MIN_MASKED_ATTENTION_VALUE
 
+_MLLAMA_DEPRECATION_MSG = (
```
quic-rishinr, line 49: instead of adding the warning here lets add it under QEffMllamaForConditionalGeneration and QEffMllamaForCausalLM. Then rever the changes in pytorch transforms

quic-rishinr (changes requested):

QEfficient/transformers/models/mllama/modeling_mllama.py
quic-hemagnih: Incorporated the review comments

PR #1025, opened 2026-06-02
ci(0306): speed up QAIC PR tests with safe parallelism

vbaddi:
## Summary
  - Enable xdist parallelism for QAIC CI stages with up to 4 workers.
  - Add an opt-in QAIC device allocator to avoid multiple runtime tests using the same device.
  - Isolate `QEFF_HOME` per xdist worker to prevent cache/output collisions.
  - Replace slow/full CI test model cards with tiny models where coverage does not require full models.
  - Reduce CLI test compile/generation shapes for faster PR turnaround.

  ## Validation
  - `python -m py_compile tests/conftest.py`
  - Focused CLI unit tests: `2 passed`
  - Affected test collection: `54 tests collected`
  - xdist collection smoke with allocator enabled: `11 tests collected`

  ## Notes
  - Full-layer profiles remain serialized to avoid multi-device contention.
  - Jenkins QAIC run is needed for final hardware validation.


vbaddi:
duplicate #1075

PR #1043, opened 2026-06-05
Fix for logits issue in Qwen3 VL MOE

tv-karthikeya:
Fix to support non layer wise export for N layers - updated logits to return lm.head() output instead of hidden_states(for layer wise)


vbaddi:
duplicate #1047

PR #1048, opened 2026-06-05
Rewrite layer-wise ONNX export as an API -> adds CustomLoader and Loop inside export

ochougul:
## Summary

Replace the env-var (`LAYERWISE_EXPORT`) + monkeypatch driven layer-wise export with a generic, first-class API.

## Key changes

- **`layerwise` flag on `from_pretrained`** — builds model on `meta` device; weights streamed per decoder-layer window during export.
- **`CustomLoader`** (`QEfficient/utils/custom_loader.py`) — HF-assisted window loader that restricts sharded checkpoints to the active layer window (handles fused-MoE etc.). Supports one or more layer prefixes.
- **Window loop in `export()`** — load → apply transforms → export → split → add-prefix → merge; `compile()` only forwards `layerwise_window_size` and compiles merged ONNX unchanged.
- **`layerwise_utils.py`** — window tiling, meta-model build, text-model resolution, window-state setattr (no monkeypatching).
- **VLM dual-QPC support** — vision encoder exported once; language decoder windowed and merged.
- **`total_layers` kwarg** on both Auto classes (must be > 1) to override exported decoder-layer count.
- **Auto-force `use_onnx_subfunctions=True`** when `layerwise=True` (with warning).
- **Remove all `LAYERWISE_EXPORT` env-var usage** from the package.
- **Updated example scripts** to new flow: `from_pretrained(layerwise=True)` + `compile(layerwise_window_size=...)`.
- **Tests** added in `tests/utils/test_layerwise_utils.py`; validated on AIC compiler for CausalLM and VLM dual-QPC.


ochougul:
being addressed in new PR

PR #1052, opened 2026-06-08
Reduce whole-model ONNX export memory

anujgupt-github:
## Problem
CPU RAM utilization during ONNX exports peak well above parameter bytes. The extra RAM came from two places: load-time dtype materialization/casts from safetensors, and ONNX initializer payloads living in the in-memory ModelProto during export. Qwen VLM dual-QPC export also kept too much of the parent VLM registered in component wrappers.

## Changes
- Added low-memory ONNX export support that externalizes initializer payloads instead of keeping large raw tensors in the in-memory ModelProto.
- Added opt-in safetensor materialization and a custom streaming loader. The loader constructs the HF model on `meta`, then streams materialized safetensor shards directly into parameters/buffers without `accelerate` or new dependencies.
- Preserved default API behavior. If no precision/materialization option is provided, normal `from_pretrained` behavior is used.
- Added example-level precision toggles so users can set `QEFF_EXAMPLE_LOAD_EXPORT_PRECISION = "float16"` in scripts; `None` keeps existing defaults.
- Fixed Qwen3-VL/Qwen3-VL-MoE dual-QPC wrapper ownership so encoder/decoder wrappers do not register the original full parent VLM.
- Disabled public `layerwise=True` entry points on this branch so validation uses the whole-model path and cannot silently produce partial exports.
- Fixed Qwen3-VL-MoE expert projection layout in the non-layerwise export path. The transformed expert tensors are already in export layout, so the sparse MoE forward should not transpose them again.

## Memory results
- Llama-70B fp16 export:
  - baseline peak RSS: 263.30 GiB
  - materialized fp16 + low-memory ONNX peak RSS: 137.15 GiB
- Qwen3-VL-30B-A3B full VLM export, vision + language, no layerwise:
  - peak RSS: 62.77 GiB
  - ONNX external data total: 58.00 GiB
  - vision external data: 1.00 GiB
  - language external data: 57.00 GiB
  - vision initializers externalized: 351/351
  - language initializers externalized: 533/533

## Validation
- `python -m ruff format --check <PR python files>`: passed
- `python -m ruff check <PR python files>`: passed
- `python -m pytest tests/unit_test/models -q`: 564 passed, 8 skipped
- `python -m pytest tests/unit_test/models/test_vlm_cpu.py tests/base/test_safetensor_materializer.py tests/base/test_export_memory_offload.py::test_low_memory_export_externalizes_initializers -q`: 48 passed

### DIFF
--- a/QEfficient/base/modeling_qeff.py
+++ b/QEfficient/base/modeling_qeff.py
@@ -48,6 +48,15 @@ from QEfficient.utils import (
     to_named_specializations,
 )
 from QEfficient.utils.export_utils import export_wrapper
+from QEfficient.utils.mem_profile import (
+    ExportMemoryProfiler,
+    flatten_torch_tensors,
+    onnx_initializer_summary,
+    profile_torch_onnx_internals,
+)
+from QEfficient.utils.onnx_external_weights import (
+    attach_external_initializers_from_torch,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -307,6 +316,19 @@ class QEFFBaseModel(ABC):
         """
         # TODO: Hack for retain_full_kv, handle this outside
         export_kwargs.pop("retain_full_kv", None)
+        low_memory_external_initializers = export_kwargs.pop("_qeff_low_memory_external_initializers", False)
+        export_params_disabled = export_kwargs.get("export_params") is False or os.environ.get(
+            "QEFF_ONNX_EXPORT_PARAMS", ""
+        ).lower() in {
+            "0",
+            "false",
+            "no",
+            "off",
+        }
+        low_memory_external_initializers = low_memory_external_initializers or export_params_disabled
+        if low_memory_external_initializers:
+            export_kwargs.setdefault("export_params", False)
+            export_kwargs.setdefault("do_constant_folding", False)
         onnx_path = export_dir / f"{self.model_name}.onnx"
 
         # Return early if ONNX already exists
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -119,6 +119,14 @@ pip install dist/qefficient-0.0.1.dev0-py3-none-any.whl
 
 For more details about using ``QEfficient`` via Cloud AI 100 Apps SDK, visit [Linux Installation Guide](https://quic.github.io/efficient-transformers/source/installation.html)
 
+## Low-memory export options
```
quic-amitraj: Better to add this under the export section of documentations.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -2233,8 +2353,12 @@ def from_pretrained(
         config = AutoConfig.from_pretrained(pretrained_model_name_or_path, trust_remote_code=True)
```
quic-amitraj, line 2634: There is no need of line #2351 now. please update.

QEfficient/utils/safetensor_materializer.py
```diff
@@ -0,0 +1,429 @@
...
+        cache_dir=cache_dir,
+        revision=revision,
+        token=token,
+        local_files_only=local_files_only,
+        force=env_flag(QEFF_MATERIALIZE_FORCE_ENV) if force is None else force,
+        hub_cache_dir=hub_cache_dir,
+    )
+
+
+def load_safetensor_checkpoint_into_model(
```
quic-amitraj, line 322: Please add docstring for all the new function and class with this PR . 
Also, please add comments to explain the code logic.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -123,6 +133,108 @@ def _resolve_torch_dtype(kwargs: dict) -> None:
...
+    For example, torch_dtype=torch.float16 materializes floating safetensor
+    weights as fp16 first, then streams them into a meta-initialized model so
+    load/export does not hold both the source checkpoint dtype and fp16 copy.
+    If torch_dtype is omitted, this returns the original source and preserves
+    the existing Hugging Face load path.
+    """
+    requested_torch_dtype = kwargs.get("torch_dtype") if "torch_dtype" in kwargs else kwargs.get("dtype")
+    _resolve_torch_dtype(kwargs)
+    if requested_torch_dtype is None:
+        if kwargs.get("low_cpu_mem_usage", None):
```
quic-amitraj: Is there need of this condition again as we already have at line 3256?

QEfficient/base/modeling_qeff.py
```diff
@@ -306,6 +315,19 @@ def _export(
...
+        ).lower() in {
+            "0",
+            "false",
+            "no",
+            "off",
+        }
+        low_memory_external_initializers = low_memory_external_initializers or export_params_disabled
+        if low_memory_external_initializers:
+            export_kwargs.setdefault("export_params", False)
+            export_kwargs.setdefault("do_constant_folding", False)
```
ochougul: we should make this True, we don't know how much graph changes because of this. Our compiler might be relying on some constant folding done by ONNX exporter

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -4588,8 +4716,10 @@ def from_pretrained(cls, pretrained_model_name_or_path, pooling=None, *args, **k
 
         kwargs.update({"attn_implementation": "eager", "low_cpu_mem_usage": False})
 
-        _resolve_torch_dtype(kwargs)
-        model = cls._hf_auto_class.from_pretrained(pretrained_model_name_or_path, *args, **kwargs)
+        pretrained_model_name_or_path = _prepare_weight_materialized_from_pretrained_source(
+            pretrained_model_name_or_path, kwargs
+        )
```
ochougul: we should be able to directly stream without making a new copy of weights ideally.

quic-amitraj (changes requested):

QEfficient/utils/safetensor_materializer.py
```diff
@@ -0,0 +1,429 @@
...
+    missing_buffer_names: List[str]
+    unexpected_tensor_names: List[str]
+
+
+def env_flag(name: str) -> bool:
+    value = os.environ.get(name, "")
+    return value.strip().lower() in {"1", "true", "yes", "on"}
+
+
+def normalize_torch_dtype(dtype: Any) -> Optional[torch.dtype]:
```
quic-amitraj: Not sure if we need this method, as we have already restricitng user from the example script to pass the choices=["float16", "bfloat16", "float32"]. I think we can avoid this one.

QEfficient/utils/safetensor_materializer.py
```diff
@@ -0,0 +1,429 @@
...
+    return MaterializedCheckpoint(
+        path=target_path,
+        source_path=source_path,
+        target_dtype=target_dtype,
+        materialized=True,
+        dtype_aligned=True,
+    )
+
+
+def maybe_materialize_safetensor_checkpoint(
```
quic-amitraj: Conditions can be shifted to materialize_safetensor_checkpoint and this method can be removed.

quic-amitraj (changes requested):

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -123,6 +133,108 @@ def _resolve_torch_dtype(kwargs: dict) -> None:
         kwargs["torch_dtype"] = torch.float32
 
 
+def _prepare_weight_materialized_from_pretrained_source(pretrained_model_name_or_path: str, kwargs: dict) -> str:
```
quic-amitraj: Better we add all these method to utills, goal is keep modeling_auto clean and it suits well in utills.

quic-amitraj (changes requested):

QEfficient/base/modeling_qeff.py
anujgupt-github: @ochougul : Agree, will enable constant folding. 
This condition starts becoming more plausible now during low memory ONNX export. Though the mapping I have is named mapping, constant folding may change some parameter-dependent subgraphs.

`folded/exported graph input names != torch named_parameters/named_buffers`

Also adding a narrow guardrail for the low-memory external-initializer path: after export, we validate that parameters/buffers promoted to graph inputs were reattached as external initializers and removed from graph inputs. If that reconstruction fails and the caller did not explicitly set do_constant_folding, we retry once with do_constant_folding=False. If the caller explicitly set do_constant_folding=True, we fail with a clear error instead of silently changing their requested export behavior.

README.md
anujgupt-github: Done. Moving the low-memory export details out of the root README and into the export documentation

QEfficient/transformers/models/modeling_auto.py
anujgupt-github: Done. Removed the local AutoConfig import and reused the existing top-level import

QEfficient/utils/safetensor_materializer.py
anujgupt-github: This helper is private and kept local to the safetensor utility. I think we still need it because this path is used by the library API, not only by the example script. Callers can pass torch.float16, "float16", or "torch.float16" directly to from_pretrained/export flows, so the utility should normalize and validate those inputs consistently.

QEfficient/transformers/models/modeling_auto.py
anujgupt-github: Agree, and if the materialization takes longer, we can look into adding thread-pool.
As of now, OK to move in memory stream tensors directly into a meta-initialized model

QEfficient/utils/safetensor_materializer.py
anujgupt-github: Adding

QEfficient/transformers/models/modeling_auto.py
anujgupt-github: Agree, not needed

QEfficient/utils/safetensor_materializer.py
anujgupt-github: Agree. Removing maybe_materialize wrapper from the from_pretrained path. The normal explicit-dtype path now uses a direct streaming (in-memory), and materialize_safetensor_checkpoint remains as the explicit utility for callers/tests that need a converted checkpoint copy.

QEfficient/transformers/models/modeling_auto.py
anujgupt-github: Thanks @quic-amitraj - Agree.
moved the safetensor streaming preparation/loading helpers into QEfficient/utils/safetensor_materializer.py

PR #1065, opened 2026-06-11
NPI dependency removal for vision component

qcdipankar:
Fix for NPI dependency for Gemma4vision component


qcdipankar:
rebase issue

PR #1089, opened 2026-06-16
Added subfunction fix for qwen3.5

abhishek-singh591:


ochougul:
fixed in #1084 except the mkdir line. will be fixed as bugfix later

PR #1090, opened 2026-06-16
fix(qwen3moe): make TorchScript export+compile work with transformers…

Cs23m011:
… 5.x

transformers 5.x changed Qwen3 MoE internals in two ways that broke the TorchScript (legacy torch.onnx.export) path:

1. Gate API change: Qwen3MoeTopKRouter.forward() now returns a 3-tuple (router_logits, scores, indices) instead of a plain logits tensor. Forward methods that assumed a single tensor return now crash.

2. Expert storage change: Qwen3MoeExperts uses stacked weight tensors (gate_up_proj, down_proj) instead of nn.ModuleList, so self.experts[e] is no longer subscriptable.

3. ReduceSum in ONNX subfunctions: ONNX opset-17 encodes ReduceSum axes as a separate input tensor.  The QAIC compiler rejects this when compiling a subfunction in isolation ("Non-constant axes tensor not supported").  Qwen3MoeTopKRouter.forward() contains such a ReduceSum via norm_topk_prob.  Fix by subclassing as QEffQwen3MoeTopKRouter and replacing .sum() with matmul@ones which generates a MatMul ONNX node.

4. MDP partition JSON written before its parent directory exists, causing compile to fail when num_devices > 1.  Fix by calling mkdir before create_json.

Also adds examples/text_generation/torchscript_export_compile.py as a ready-to-run script for TorchScript export and compile.

### DIFF
--- a/QEfficient/base/modeling_qeff.py
+++ b/QEfficient/base/modeling_qeff.py
@@ -915,6 +915,8 @@ class QEFFBaseModel(ABC):
                 mdp_ts_num_devices, compiler_options.get("aic_num_cores", constants.DEFAULT_AIC_NUM_CORES)
             )
             mdp_ts_json_path = compile_dir / f"mdp_ts_{mdp_ts_num_devices}.json"
+            # TODO: remove below line, fallback when user explicitly sends compile_dir in compile API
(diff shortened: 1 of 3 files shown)


ochougul:
fixed in #1084

PR #1093, opened 2026-06-17
Fix Gemma4 continuous batching: batch_index propagation and mm_token_type_ids shape

quic-amitraj:
## Summary

This PR fixes two bugs in the Gemma4 continuous-batching (CB) export and inference path, adds a working CB example script, and adds unit tests for all three Gemma4 inference variants.

---

## Bug 1 — `batch_index` not propagated through Gemma4 decoder (root cause)

**Root cause:** `batch_index` was not being passed through the decoder attention and KV cache update logic in CB mode. The KV cache was allocated at `full_batch_size`, but without `batch_index` the gather/scatter behaved like non-CB, causing a batch broadcast inside attention that made `attn_output` width 8192 instead of the expected hidden size (2048) fed into `o_proj`:

```
mat1 and mat2 shapes cannot be multiplied (288x8192 and 2048x1536)
```

A secondary issue was a hardcoded `compile(...)` block in the local example that overrode `compile_kwargs`, masking the real bug.

**Fix:** `batch_index` is now correctly threaded through the Gemma4 decoder and attention cache update path in CB mode; example script uses consistent `compile_kwargs`.

---

## Bug 2 — `mm_token_type_ids` decode shape wrong in CB mode

**Root cause:** `VisionLanguageGeneration._execute_chunked_prefill` allocated `_decode_mm_token_type_ids` using the per-slot prefill batch size (1), not `full_batch_size` (4). Every CB decode step failed with:

```
No matching specialization found.
Buffername: mm_token_type_ids; Supplied: <1 x 1>; Expected: <4 x 1> or <1 x 256>
```

A leftover `import ipdb; ipdb.set_trace()` debug breakpoint was also present in the CB decode loop.

**Fix:** Use `self.full_batch_size` for `_decode_mm_token_type_ids` shape in CB mode; removed debug breakpoint.

---

## Changes

| File | Change |
|------|--------|
| `QEfficient/generation/vlm_generation.py` | Use `full_batch_size` for `_decode_mm_token_type_ids` in CB mode |
| `QEfficient/generation/text_generation_inference.py` | Remove `import ipdb; ipdb.set_trace()` from CB decode loop |
| `examples/image_text_to_text/models/gemma_vision/gemma4/continuous_batching.py` | Working end-to-end CB example |
| `examples/image_text_to_text/models/gemma_vision/gemma4/gemma4_cb.py` | CB example (alternate entry) |
| `tests/unit_test/utils/test_generation.py` | 19 unit tests for all 3 Gemma4 variants |

---

## Unit tests added

**`TestGemma4CBVariant`** — `gemma4_cb.py` / `continuous_batching.py` (7 tests):
- No debug breakpoints in CB decode loop
- `full_batch_size` stored, `batch_index` shape `(full_batch_size, 1)` and values `0…N-1`
- `prepare_decode_inputs` includes `batch_index`
- `_decode_mm_token_type_ids` uses `full_batch_size` not per-slot size (regression guard)
- `compile()` accepts `full_batch_size`

**`TestGemma4RegularVariant`** — `gemma4_example.py` (5 tests):
- `full_batch_size` is `None`, no `batch_index` in decode inputs
- Decode `input_ids` shape `(batch_size, 1)`, `mm_token_type_ids` absent when not set
- `compile()` accepts `skip_vision`

**`TestGemma4DissVariant`** — `gemma4_diss.py` (7 tests):
- `compile()` accepts `prefill_only`, `enable_chunking`, `skip_lang`
- Ceiling-division chunking, padded length is multiple of `PREFILL_SEQ_LEN`
- `mm_token_type_ids` sliced per chunk during prefill; absent from decode inputs

---

## Test commands run

```bash
python examples/image_text_to_text/models/gemma_vision/gemma4/continuous_batching.py  # exit 0, 4 responses generated
pytest tests/unit_test/utils/test_generation.py::TestGemma4CBVariant \
       tests/unit_test/utils/test_generation.py::TestGemma4RegularVariant \
       tests/unit_test/utils/test_generation.py::TestGemma4DissVariant -v   # 19 passed
```

> AI-assisted: all changes reviewed and validated by the author.

### DIFF
--- a/examples/image_text_to_text/models/gemma_vision/gemma4/gemma4_example.py
+++ b/examples/image_text_to_text/models/gemma_vision/gemma4/gemma4_example.py
@@ -93,7 +93,7 @@ def main():
     )
     remove_fp16clip_transform_if_disabled(qeff_model, True)
 
-    if not SKIP_VISION:
+    if SKIP_VISION:
         messages = build_messages(SYSTEM_PROMPT, TEXT_PROMPT, use_image=False)
         text_inputs = processor.apply_chat_template(
             messages,
--- a/examples/image_text_to_text/models/gemma_vision/gemma4/gemma4_example.py
+++ b/examples/image_text_to_text/models/gemma_vision/gemma4/gemma4_example.py
@@ -24,6 +24,7 @@ IMAGE_PROMPT = "Can you Describe this image in detail?"
 IMAGE_URL = "https://wallup.net/wp-content/uploads/2017/03/28/351036-San_Francisco-USA-bridge-sunset-Golden_Gate_Bridge-lights.jpg"
 SKIP_VISION = False
 BS = 1
+FULL_BATCH_SIZE = 2
 PREFILL_SEQ_LEN = 128
 CTX_LEN = 2048
 GENERATION_LEN = 1920
@@ -42,7 +43,7 @@ NODE_PRECISION_INFO = True
 
 compiler_kwargs = {
     "NUM_CORES": 16,
-    "NUM_DEVICES": 4,
+    "NUM_DEVICES": 2,
     "MXFP6_MATMUL": True,
     "MXINT8_KV_CACHE": True,
     "AIC_ENABLE_DEPTH_FIRST": True,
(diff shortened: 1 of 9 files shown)


examples/image_text_to_text/models/gemma_vision/gemma4/gemma4_example.py
```diff
@@ -93,7 +93,7 @@ def main():
     )
     remove_fp16clip_transform_if_disabled(qeff_model, True)
 
-    if SKIP_VISION:
+    if not SKIP_VISION:
```
quic-hemagnih: please remove not from here

quic-hemagnih (changes requested):

ochougul (approved):

ochougul (changes requested):
still reviewing

quic-hemagnih (approved):

PR #1105, opened 2026-06-22
cleaned up layerwise API and added CustomLoader

ochougul:
- Moves layerwise state from global/class-level flags into per-instance LayerwiseContext, preventing stale state from leaking
    across models or exports.
  - Adds CustomLoader to selectively load only the active layer window while keeping Hugging Face’s normal checkpoint loading
    flow.
  - Updates CausalLM and ImageTextToText wrappers so layerwise mode is configured at from_pretrained(..., layerwise=True,
    layerwise_window_size=...); export() / compile() layerwise flags are now deprecated.
  - Refreshes Qwen3 MoE, Qwen3.5 MoE, and Qwen3 VL MoE layerwise hooks to resolve windows/cache indices from the attached context
    instead of shared class attributes.
  - Improves layerwise export/compile behavior with merged ONNX cache reuse, intermediate cleanup, final compile context
    suspension, and safer ONNX export patch gating.
  - Expands quickcheck coverage for context scoping, window validation, selective loader filtering/meta APIs, export patch
    toggling, cache reuse, and final compile behavior.

### DIFF
--- a/QEfficient/base/modeling_qeff.py
+++ b/QEfficient/base/modeling_qeff.py
@@ -30,6 +30,12 @@ from QEfficient.base.pytorch_transforms import PytorchTransform
 from QEfficient.blocking.blocking_configurator import build_transformer_blocking_config_for_transform
 from QEfficient.compile.qnn_compiler import compile as qnn_compile
 from QEfficient.generation.cloud_infer import QAICInferenceSession
+from QEfficient.transformers.models._layerwise import (
+    get_layerwise_end,
+    get_layerwise_start,
+    get_layerwise_total_layers,
+    is_layerwise_active,
+)
 from QEfficient.transformers.models.pytorch_transforms import (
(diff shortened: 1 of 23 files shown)


vbaddi:
w/layerwise path is completely segregated from the w/o layerwise right? any intersection we see?

vbaddi:
We will throw out error or warning if w/layerwise is being used by other models except qwen series?

ochougul:
> We will throw out error or warning if w/layerwise is being used by other models except qwen series?

We error out if user passed `layerwise=True` in `from_pretrained`, and the model does not support that.

ochougul:
> w/layerwise path is completely segregated from the w/o layerwise right? any intersection we see?

No intersection. The paths are separate. There are more methods that needed and can be cleaned up but leaving that for later as of now. Critical Reviews from our release review are addressed.

PR #1107, opened 2026-06-22
fix(mixtral): fix MXFP6 quantization and ONNX export issues for Mixtral_moe 

Cs23m011:
- Add QEffMixtralExperts with __qeff_init__ to split gate_up_proj into gate_proj/up_proj and transpose weights to (H, I) layout so expert weights appear as the second BMM operand (input @ weight), enabling QAIC -mxfp6-matmul to quantize them. Without this, _batched_linear used weight-first BMM which QAIC does not quantize, causing ~144 GB DRAM instead of ~42 GB.

- Rewrite QEffMixtralSparseMoeBlock.forward() to mirror QEffQwen3MoeSparseMoeBlock: dynamic gather + BMM with weight as second operand, einsum accumulation.

- Add QEffMixtralTopKRouter to replace router .sum(dim=-1, keepdim=True) with torch.einsum consistent with QEffQwen3MoeTopKRouter, avoiding ReduceSum with non-constant axes in ONNX opset 17 / PyTorch 2.7.

- Set QEffMixtralForCausalLM.ONNX_EXPORT_EXAMPLE_SEQ_LEN = 2 to reduce the expert weight gather from ~15 GB to ~1 GB during ONNX tracing; seq_len is a dynamic axis so the compiled QPC uses real prefill_seq_len.

- Register MixtralExperts, MixtralTopKRouter transforms in pytorch_transforms.py.

- Add --num-layers arg to continuous_batching.py for quick layer-reduced testing.

### DIFF
--- a/QEfficient/transformers/models/modeling_auto.py
+++ b/QEfficient/transformers/models/modeling_auto.py
@@ -3761,7 +3761,7 @@ class QEFFAutoModelForCausalLM(QEFFBaseModel):
             )
 
         bs: int = constants.ONNX_EXPORT_EXAMPLE_BATCH_SIZE
-        seq_len: int = constants.ONNX_EXPORT_EXAMPLE_SEQ_LEN
+        seq_len: int = getattr(self.model, "ONNX_EXPORT_EXAMPLE_SEQ_LEN", constants.ONNX_EXPORT_EXAMPLE_SEQ_LEN)
 
         # increase seq_len if using a larger number of blocks
         if self.hash_params.get("blocking_kwargs", None):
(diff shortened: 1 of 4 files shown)


QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3761,7 +3761,7 @@ def export(
             )
 
         bs: int = constants.ONNX_EXPORT_EXAMPLE_BATCH_SIZE
-        seq_len: int = constants.ONNX_EXPORT_EXAMPLE_SEQ_LEN
+        seq_len: int = getattr(self.model, "ONNX_EXPORT_EXAMPLE_SEQ_LEN", constants.ONNX_EXPORT_EXAMPLE_SEQ_LEN)
```
vbaddi, line 3764: nit: why is this change required?

vbaddi (changes requested):

quic-rishinr:
Generalized changes will be added part of #1090 closing this PR

PR #1114, opened 2026-06-24
RepeatKV Transform changes with comments addressed

quic-dhirajku:
• Refactor RepeatKV end-to-end for LLM/VLM language paths and align runtime/test plumbing.

  - Reworked ReplicateKVHeadTransform into a mutator-style flow with strict config validation, idempotent re-apply handling, and encoder-wrapper skip (EncoderWrapper class-name guard).
  - Replaced KV projection duplication branching with internal factory dispatch (repeat_kv_projection_dispatch.py) and routed all duplication through shared repeat_kv_utils.
  - Added shared config utilities/constants to resolve head/hidden keys across model families and compute/apply num_replicate_kv_heads.
  - Plumbed num_replicate_kv_heads through auto-model wrappers/export paths for CausalLM and ImageTextToText.
  - Updated tests/helpers for RepeatKV coverage (causal + image-text paths, plus fast unit checks) and renamed example/ docs QAIC knob from num_kv_heads_repeat to num_replicate_kv_heads.

### DIFF
--- a/QEfficient/base/modeling_qeff.py
+++ b/QEfficient/base/modeling_qeff.py
@@ -52,6 +52,7 @@ from QEfficient.utils import (
     require_value,
     to_named_specializations,
 )
+from QEfficient.utils.config_utils import calculate_num_replicate_kv_heads
 from QEfficient.utils.export_utils import export_wrapper
 from QEfficient.utils.torch_patches import layerwise_safe_onnx_export_patches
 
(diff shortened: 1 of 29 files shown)


QEfficient/base/modeling_qeff.py
```diff
@@ -790,23 +791,41 @@ def transform(
...
+                    # RepeatKV is intentionally skipped for encoder wrappers, but we still
+                    # want the requested value reflected in encoder hash params.
+                    if "EncoderWrapper" in self.model.__class__.__name__:
+                        effective_num_replicate_kv_heads = num_replicate_kv_heads
+                if replicate_kv_transformed:
+                    if hasattr(model_config, "to_diff_dict"):
+                        self.hash_params["config"] = model_config.to_diff_dict()
+                    else:
+                        self.hash_params["config"] = model_config
+                    effective_num_replicate_kv_heads = num_replicate_kv_heads
```
ochougul: can you make a new method called `maybe_apply_replicate_kv_transform` and move the new code there?
Right a clean method and call it here.

QEfficient/transformers/models/molmo/modeling_molmo.py
```diff
@@ -565,6 +565,7 @@ class QEffMolmoEncoderWrapper(nn.Module):
     def __init__(self, model):
         super().__init__()
         self.model = model
+        self.config = self.model.config
```
ochougul: why is this needed. Its okay for encoderWrapper to not have the config right?  as we can have a early check that checks if EncoderWrapper is in the `self.__clas__.__name__` then don't even check config?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1535,6 +1535,7 @@ def from_pretrained(cls, pretrained_model_name_or_path: str, qaic_config: Option
         )
 
         _resolve_torch_dtype(kwargs)
+        num_replicate_kv_heads = kwargs.pop("num_replicate_kv_heads", 1)
```
ochougul: This should be part of `qaic_config` not `kwargs`

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -979,73 +992,144 @@ class RevertPrefillOnlyTransform(ModuleMappingTransform):
...
-        # Check for CausalLMs
-        if hasattr(model, "model"):
-            return model.model
-
-        raise AttributeError("No suitable text model found in the provided model.")
+        cfg.orig_kv_heads = orig_kv_heads
+        set_kv_head_aliases(cfg, new_kv_heads)
+
+        logger.warning(f"Original KV heads: {orig_kv_heads}")
+        logger.warning(f"Modified KV heads: {new_kv_heads}")
```
ochougul, line 1117: this should be at end of transform after the actual mutate is done. i.e. just before return in this method.

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -979,73 +992,144 @@ class RevertPrefillOnlyTransform(ModuleMappingTransform):
...
-                attn = getattr(block, "cross_attn", getattr(block, "self_attn", None))
-                attn.num_key_value_heads = new_kv_heads
-                head_dim = attn.kv_lora_rank + attn.qk_rope_head_dim
-
-                cls._duplicate_weights_for_linear_layer(
-                    attn.kv_a_proj_with_mqa, orig_kv_heads, num_kv_heads_repeat, head_dim, hidden_size
+        if n_repeat is not None and n_repeat > 1:
+            # RepeatKV is decode-language only; encoder wrappers are explicitly excluded.
+            if "EncoderWrapper" in model.__class__.__name__:
+                return model, transformed
```
ochougul: I think this condition should be outside. i.e. if EncoderWrapper never call this Transform

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -979,73 +992,144 @@ class RevertPrefillOnlyTransform(ModuleMappingTransform):
...
+                    f"{model.__class__.__name__}; expected an nn.Module."
+                )
+            if mutated_model is not model:
+                raise RuntimeError(
+                    f"{cls.__name__}.mutate returned a different module instance for {model.__class__.__name__}. "
+                    "RepeatKV mutate is expected to mutate and return the same instance."
                 )
+            after_kv_heads = resolve_kv_heads(get_text_model(mutated_model).config)
+            transformed = after_kv_heads != before_kv_heads
+            return mutated_model, transformed
```
ochougul: this code feels like a test, it should be in test, we don't really need to change apply method.

ochougul (changes requested):

ochougul (approved):
LGTM

PR #1116, opened 2026-06-24
fix(QRANIUMSW-62219): Gemma4 pop vision_size from compiler_options in…

quic-gthiruko:
… get_specializations

Gemma4 does not support the vision_size and uses a fixed vision_tokens count via mm_tokens_per_image=256), causing QPC compilation to fail with when skip_vision=True and vision_size is supplied:

  Invalid option -vision-size=xxx

All other VLMs (gemma3, llava, internvl, qwen2_5_vl, etc.) already pop vision_size from compiler_options in get_specializations() before it reaches the QAIC compiler. Streamlining Gemma4 to be consistent with other supported VLMs.


quic-gthiruko:
Unit Tests:
 
Tested the changes with vLLM flow for the following models and is working as expected. Logs from the run attached in the JIRA.
 
- google/gemma-4-E2B-it & google/gemma-4-E4B-it --> 2 QPC (Dense)
- google/gemma-4-26B-A4B-it --> 3 QPC (MoE)

quic-rishinr:
Closing this PR as @quic-sanising will be raising a new PR

PR #1117, opened 2026-06-24
Make num_cores (NSP) configurable via NUM_CORES env for nightly pipeline

quic-vishali:
**Summary**:
Enabled dynamic configuration of num_cores (NSP) in the nightly pipeline using an environment variable.

**Background**:
Nightly tests were failing on setups with 14 NSP cores because the default hardcoded value of num_cores was set to 16. This mismatch caused runtime errors due to insufficient resources.

**Change**:
- Added support for overriding num_cores using the NUM_CORES environment variable
- If NUM_CORES is set -> its value is used
- If not set -> pipeline config value is used (default remains unchanged)

**Usage**:
Before running the nightly tests, export the NUM_CORES environment variable if needed.

**Example**:
export NUM_CORES=14


quic-rishinr:
Enabled for testing the CI.

PR #1118, opened 2026-06-24
Cherry-picking Gemma4 with subfunction 

tchawada:
Cherry picked the latest commit from PR-1109.


quic-hemagnih:
@tchawada we need to only cherry pick the fixes which are made in PR 1109, we should not take any other changes from main branch.

tchawada:
fixed in PR 1125

PR #1121, opened 2026-06-24
Reduce MoE export RAM by aliasing fused expert weights as views

quic-rishinr:
## Summary

Reduces peak host RAM during export/compile of MoE models by avoiding a duplicate copy of the expert weights.

The QEff MoE experts wrappers build transposed aliases (`gate_proj`, `up_proj`, `down_proj_t`) from the fused HF parameters (`gate_up_proj`, `down_proj`). They did this with `.detach().clone().transpose(...)`, where the `.clone()` allocates a **second full copy** of the expert weights. Since experts are the overwhelming majority of an MoE model's parameters, this roughly **doubled resident PyTorch RAM**.

On `Qwen/Qwen3-VL-30B-A3B-Instruct` this manifested as ~60 GB → ~120 GB resident, and the disaggregated prefill + decode flow (which intentionally keeps PyTorch weights resident across the prefill export so decode can re-export) stacked `torch.onnx.export` buffers on top of the already-doubled baseline, pushing peak RAM beyond 120 GB.

## Change

Drop the `.clone()` so the transposed aliases are **views** that share storage with the fused `gate_up_proj`/`down_proj` instead of allocating a second copy. `requires_grad` is carried over from the source parameters.

Applied to the three affected MoE families:
- `QEffQwen3MoeExperts` (`qwen3_moe`)
- `QEffQwen3_5MoeExperts` (`qwen3_5_moe`)
- `QEffQwen3VLMoeTextExperts` (`qwen3_vl_moe`)

## What changes from existing architecture

No behavioral change. The aliases hold the same values, just as views over the fused storage rather than independent copies. The fused parameters remain registered (so checkpoint/state_dict and the FP8 dequant path are unaffected); only the duplicate allocation is removed.

## Validation

- **Memory:** verified on `tiny-random/qwen3-vl-moe` that `gate_proj` now shares storage with `gate_up_proj`, and unique storage bytes per experts module halved (98304 → 49152).
- **Accuracy:** MoE block forward output is **bit-identical** before/after (`max abs diff = 0.0`).
- **Offload:** `_offload_model_weights(True)` still moves all params to `meta` with the shared storage.

### Tests run (CPU, `pytest -n0`)
- `tests/transformers/models/test_moe_prefill_blocked.py` — **13 passed** (qwen3moe, glm4_moe, gptoss forward-parity + prefill/decode export)
- `tests/unit_test/models/test_model_quickcheck.py -k "qwen3_5_moe or qwen3_vl_moe"` — **9 passed**

`on_qaic` hardware runs and the full 30B end-to-end RAM/accuracy check were not run in this environment and should be confirmed on a QAIC host.

## AI assistance disclosure
AI assistance (Claude Code) was used to investigate the root cause and draft the change; all edits were reviewed and the tests above were run.


QEfficient/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py
```diff
@@ -2112,9 +2112,15 @@ def __qeff_init__(self):
         # HF 5.x keeps fused gate_up projections. Keep aliases expected by
         # QEff MoE execution paths without changing checkpoint behavior.
         self.expert_dim = getattr(self, "intermediate_size", self.gate_up_proj.shape[-2] // 2)
-        self.gate_proj = nn.Parameter(self.gate_up_proj[:, : self.expert_dim, :].detach().clone().transpose(1, 2))
-        self.up_proj = nn.Parameter(self.gate_up_proj[:, self.expert_dim :, :].detach().clone().transpose(1, 2))
-        self.down_proj_t = nn.Parameter(self.down_proj.detach().clone().transpose(1, 2))
+        gate_up_proj = self.gate_up_proj.detach()
+        down_proj = self.down_proj.detach()
+        self.gate_proj = nn.Parameter(
+            gate_up_proj[:, : self.expert_dim, :].transpose(1, 2), requires_grad=self.gate_up_proj.requires_grad
```
ochougul: Let's hardcode `requires_grad=False`

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1606,10 +1611,11 @@ def export(
...
 
-        if prefill_only and prefill_seq_len > 1:
-            offload_pt_weights = False  # to keep weight for decode onnx
-        else:
-            offload_pt_weights = kwargs.get("offload_pt_weights", True)
+        if offload_pt_weights is None:
+            if prefill_only and prefill_seq_len > 1:
+                offload_pt_weights = False  # keep weights resident for the decode export
+            else:
+                offload_pt_weights = True
```
ochougul, line 1619: can you add a `TODO: remove this once CustomLoader is in place`

ochougul (changes requested):

ochougul (approved):
This will still fail whenever user changes the order of export ie decode first and then prefill
But approving for release

QEfficient/transformers/models/modeling_auto.py
quic-rishinr: done

PR #1141, opened 2026-06-30
Added multi resolution support for qwen models on 1.22.0_tf_4.57.3 branch

quic-rishinr:
get_specializations now accepts list-valued height/width/num_frames, building one vision specialization per resolution. This fixes the vLLM path,
  which passes dimensions as single-element lists (e.g. height=[512], width=[910]) and previously crashed with TypeError: unsupported operand type(s) 
  for /: 'list' and 'list' inside smart_resize.

  Changes:
  - Replace the inlined smart_resize helper with from qwen_vl_utils import smart_resize across the three model files.
  - Hoist IMAGE_FACTOR / min-max pixel constants into QEfficient/utils/constants.py.
  - Support vision_size and mm_processor_kwargs overrides via compiler_options.
  - Declare qwen-vl-utils==0.0.8 in pyproject.toml.
  - Add a regression unit test covering single- and multi-resolution inputs.


pyproject.toml
```diff
@@ -37,6 +37,7 @@ dependencies = [
     "tensorboard",
     "fire",
     "py7zr",
+    "qwen-vl-utils==0.0.8",
```
ochougul: what numpy version does this install? do we have approval?

quic-rishinr: We have approval for "qwen-vl-utils==0.0.8". For numpy we will be following the same process on 1.22 release

QEfficient/generation/cloud_infer.py
```diff
@@ -104,9 +104,14 @@ def __init__(
...
-            prog_properties.devMapping = ":".join(map(str, device_ids))
-        self.program = qaicrt.Program(self.context, None, qpc, prog_properties)
+        dev_id_non_mq = None
+        if device_ids:
+            if len(device_ids) == 1:
+                dev_id_non_mq = device_ids[0]
+            elif len(device_ids) > 1:
+                prog_properties.devMapping = ":".join(map(str, device_ids))
+        self.program = qaicrt.Program(self.context, dev_id_non_mq, qpc, prog_properties)
+
```
ochougul, line 114: If autodevicepicker is not working as expected, this is a platform bug right?

ochougul (changes requested):

QEfficient/generation/cloud_infer.py
quic-rishinr: Currently we are not handling the case when len(device_ids) ==1. The condition is only handled for device_ids and len(device_ids) > 1.

ochougul: It should be simple
I think we need to pass None, in `qaicrt.Program` for autodevicepicker to kick in. Let's keep way of calling this API as old way. This should be only single line change
```
if device_ids:
        prog_properties.devMapping = ":".join(map(str, device_ids))
        self.program = qaicrt.Program(self.context, None, qpc, prog_properties)
```

quic-rishinr: you should be passing the qaicrt.Program(self.context, **dev_id_non_mq**, qpc, prog_properties) for non mq cases

ochougul (approved):

QEfficient/generation/cloud_infer.py
quic-akuruvil: Yes, this case helps with specifying which device id(single device case) to pick incase user gives one explicitly. Otherwise, will fall back to None(autopick).

quic-hemagnih (approved):

PR #1142, opened 2026-06-30
Make dataPathTimeoutMs configurable, default 120s

quic-rishinr:
On long-context, long-generation runs the 60s data-path timeout is too tight. With Llama-3.3-70B at ctx-len 32768 / generation-len 16000 using hqkv blocking, a single decode iteration can take longer than 60s, and the host aborts mid-generation across all devices with "failed to receive response from NW: Connection timed out"

The timeout was hardcoded at 60s with no way to change it short of editing the source. This exposes `data_path_timeout_ms` on `QAICInferenceSession` and bumps the default to 120s. Existing callers are unaffected since it's a keyword arg with a default, and anyone who needs more can pass a larger value per session.

Tested locally with the blocked_attention_inference example on the reported config; the run now completes instead of timing out at 60s.

### DIFF
--- a/QEfficient/generation/cloud_infer.py
+++ b/QEfficient/generation/cloud_infer.py
@@ -67,7 +67,7 @@ class QAICInferenceSession:
         device_ids: Optional[List[int]] = None,
         activate: bool = True,
         enable_debug_logs: bool = False,
-        data_path_timeout_ms: int = 120_000,
+        data_path_timeout_ms: int = 60_000,
     ):
         """
         Initialise for QAIC inference Session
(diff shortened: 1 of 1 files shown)


QEfficient/generation/cloud_infer.py
```diff
@@ -67,6 +67,7 @@ def __init__(
         device_ids: Optional[List[int]] = None,
         activate: bool = True,
         enable_debug_logs: bool = False,
+        data_path_timeout_ms: int = 60_000,
```
quic-hemagnih, line 70: default is 120s or 60s?

quic-rishinr: we are setting default as 60 in our case.

quic-hemagnih: Then should we update the doc string - to 60sec?
data_path_timeout_ms: int. Host wait timeout (in ms) for a data-path response from the device. Default=120000 (120s).

quic-mamta (approved):
LGTM

quic-hemagnih (approved):

PR #1143, opened 2026-07-02
Replaced qwen3.5 models with dummy models and added support for zero weigths

quic-rishinr:
Replaces the real Qwen3.5 models in the dummy_layers VLM subfunction tests with tiny-random dummy models


quic-rishinr:
PR not relevant anymore. Closing this PR

PR #1147, opened 2026-07-03
Updated Default AIC hardware detection to NSP IMAGE_VARIANT

asmigosw:
## Summary

  This PR updates AI200 default hardware detection and adjusts bf16 handling in compile-time custom_io validation.

  ## Changes

  1. Updated default AIC hardware detection in QEfficient/utils/constants.py:

  - Switched regex match from FW IMAGE_VARIANT to NSP IMAGE_VARIANT when parsing qaic-util -q output.
  - Keeps existing fallback behavior to "ai100" when detection fails.

  2. Updated bf16 input checks in QEfficient/base/modeling_qeff.py:

  - Expanded bf16 detection logic for custom_io keys from only past_* to both:
      - past_*
      - pixel_values
  - This ensures bf16 usage in vision/multimodal-style inputs is correctly detected during compile flow.

  3. Removed stale warning in QEFFBaseModel init path:

  - Dropped warning that said bf16 is unsupported and would be converted to fp16.
  - Aligns logging with current expected behavior for AI200-related paths.

  ## Why

  - NSP IMAGE_VARIANT is the correct field for newer hardware/version reporting in current qaic-util -q output.
  - bf16 may now be intentionally provided for pixel_values; detection logic should reflect this.
  - Old warning was misleading for current bf16 support assumptions.


QEfficient/utils/constants.py
```diff
@@ -137,7 +137,7 @@ def get_default_aic_hw_version() -> str:
     except Exception:
         return "ai100"
 
-    match = re.search(r"FW IMAGE_VARIANT\s*:\s*(\S+)", output)
+    match = re.search(r"NSP IMAGE_VARIANT\s*:\s*(\S+)", output)
```
quic-rishinr: This assumes the platform ask is installed on the system if it's not installed it might throw an error. Add a fallback method using "lspci | grep Qualcomm"

quic-rishinr (changes requested):

QEfficient/utils/constants.py
```diff
@@ -114,13 +114,55 @@ def get_onnx_export_opset(dynamo: bool = False) -> int:
...
+            text=True,
+            timeout=10,
+        )
+        output = result.stdout
+    except Exception:
+        return "ai100"
+
+    for line in output.splitlines():
+        # lspci -n lines look like: "01:00.0 0200: 17cb:a0dc (rev 01)"
+        match = re.search(rf"{_QUALCOMM_PCI_VENDOR_ID}:([0-9a-fA-F]+)", line, re.IGNORECASE)
```
quic-rishinr: why do we need a vendor ID? can we just search for Qualcomm?

QEfficient/utils/constants.py
```diff
@@ -114,13 +114,55 @@ def get_onnx_export_opset(dynamo: bool = False) -> int:
 COMPILER = ["/opt/qti-aic/exec/qaic-compile", "-aic-hw"]
 
 
+# Qualcomm PCI vendor ID and known Cloud AI device IDs.
+_QUALCOMM_PCI_VENDOR_ID = "17cb"
+_AIC200_PCI_DEVICE_IDS = {"a110"}
```
quic-rishinr: nit: these would change with the hardware revision add a note

quic-rishinr (changes requested):

QEfficient/utils/constants.py
asmigosw: Without using `-n` the output format varies by pci.ids database version, which is maintained separately independently of lspci itself, which is why the code uses -n for robustness. Although, we can use `qualcomm`. I'll update the code.

quic-rishinr (approved):
Nit: Better to keep the new file name as hw_utils.py
other changes look good to me.

QEfficient/base/modeling_qeff.py
```diff
@@ -1200,7 +1197,7 @@ def _compile(
         # Write custom_io.yaml file
         model_in_bfloat16 = hasattr(self, "config") and (self.config.torch_dtype == torch.bfloat16)
         pkv_in_bfloat16 = (custom_io is not None) and any(
-            "past_" in key and "bfloat16" in value for key, value in custom_io.items()
+            ("past_" in key or "pixel_values" in key) and "bfloat16" in value for key, value in custom_io.items()
```
ochougul, line 1200: why are you changing this?

ochougul (changes requested):
We should also change default value of num_cores based on this detcted hardware version

Also, let us call this method again and again instead of setting it's value in `constants.py` That's more explicit

QEfficient/base/modeling_qeff.py
quic-rishinr: this change was added for supporting BF16 and got reverted in one PR. Adding back this change.

quic-rishinr:
> We should also change default value of num_cores based on this detcted hardware version
> 
> Also, let us call this method again and again instead of setting it's value in `constants.py` That's more explicit
its a good point. @asmigosw lets add a singleton class with a method to return the number of cores. Keep the default as 16, and if the target is AI200, return 4 as the default num_cores and update the DEFAULT_AIC_NUM_CORES variable to use it. constants.py

ochougul (approved):

ochougul:
CI-Ready

quic-rishinr (approved):
LGTM!! Thanks

ochougul:
CI-Ready

ochougul:
CI-Ready

ochougul:
CI-Ready

PR #1151, opened 2026-07-06
Enabling model Qwen3-Embedding-0.6B and Qwen3-Embedding-8B

quic-amitraj:
## Problem
AI100 support for `Qwen/Qwen3-Embedding-0.6B` and `Qwen/Qwen3-Embedding-8B` was failing in the `QEFFAutoModel` embedding flow.

Observed failures:
- ONNX export/runtime path stayed on HF Qwen3 modules in the embedding (`AutoModel`) path.
- No-cache path hit `UnboundLocalError` in `past_key_value_update()`.
- AI100 runtime assumed fixed QPC binding layout and always-required `attention_mask`.
- Script-level compile/inference batch mismatch caused buffer-size errors.

## Root Cause
These checkpoints are embedding variants, but their config is still Qwen3 decoder-family (`model_type=qwen3`, `architectures=Qwen3ForCausalLM`).
In this repo, embedding flow uses `QEFFAutoModel`, so decoder-family Qwen3 still needs QEff module remapping and robust AI100 I/O handling in that path.

## Fixes
1. Enabled Qwen3 remapping in `QEFFAutoModel` by adding `KVCacheTransform`.
2. Initialized `cache_kwargs` in `past_key_value_update()` for no-cache execution paths.
3. Fixed Qwen3 blocked-attention kwarg mismatch (`comp_ctx_lengths`).
4. Updated AI100 embedding runtime to:
   - resolve output binding by declared output name (no hardcoded binding index),
   - pass `attention_mask` only when the compiled QPC declares it.
5. Updated `testing_qwen.py` to compile with actual batch/seq and consume AI100 output tensor correctly.
6. Added focused unit tests for:
   - Qwen3 transform application in `QEFFAutoModel`,
   - no-cache Qwen3 forward path,
   - single-input-QPC runtime handling.

## Compatibility / Risk
- No behavior change for encoder-only embedding models; `KVCacheTransform` remains a no-op when no matching decoder modules exist.
- AI100 embedding runtime now supports both QPC signatures:
  - `input_ids + attention_mask`, and
  - `input_ids`-only.

### DIFF
--- a/QEfficient/transformers/models/modeling_auto.py
+++ b/QEfficient/transformers/models/modeling_auto.py
@@ -469,7 +469,7 @@ class QEFFAutoModel(QEFFTransformersBase):
     """
 
     _hf_auto_class = AutoModel
-    _pytorch_transforms = [CustomOpsTransform, AwqToMatmulNbitsTransform, GPTQToMatmulNbitsTransform]
+    _pytorch_transforms = [CustomOpsTransform, KVCacheTransform, AwqToMatmulNbitsTransform, GPTQToMatmulNbitsTransform]
     # FP16Clip inlines external weights; without Split the saved protobuf exceeds 2GB for large embedders.
     _onnx_transforms = [FP16ClipTransform, SplitTensorsTransform]
 
@@ -777,7 +777,7 @@ class QEFFAutoModel(QEFFTransformersBase):
             self.qpc_session = QAICInferenceSession(str(self.qpc_path), device_ids)
             self.batch_size = self.qpc_session.bindings[0].dims[0]
 
-        # Dynamic switching to closest seq_Len based on input_ids_len
+        # Dynamic switching to closest seq_len based on input_ids length.
         input_ids_len = inputs["input_ids"].shape[1]
 
         for allowed_shape in self.qpc_session.allowed_shapes:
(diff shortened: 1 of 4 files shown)


QEfficient/transformers/models/modeling_auto.py
```diff
@@ -726,7 +777,7 @@ def cloud_ai_100_feature_generate(
             self.qpc_session = QAICInferenceSession(str(self.qpc_path), device_ids)
             self.batch_size = self.qpc_session.bindings[0].dims[0]
 
-        # Dynamic switching to closest seq_Len based on input_ids_len
+        # Dynamic switching to closest seq_len based on input_ids length.
         input_ids_len = inputs["input_ids"].shape[1]
 
         for allowed_shape in self.qpc_session.allowed_shapes:
```
quic-hemagnih, line 783: Is the list self.qpc_session.allowed_shapes is sorted list or not? If its a sorted list then we can apply the binary seach here rather than linear search. 
It will cut the time complexity from O(N) to O(logN)

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -742,34 +793,39 @@ def cloud_ai_100_feature_generate(
...
+                torch.nn.functional.pad(
+                    inputs["attention_mask"], (0, self.seq_len - inputs["attention_mask"].size(1)), "constant", 0
+                )
             )
-        )
+            session_inputs["attention_mask"] = attention_mask
 
-        inputs = dict(input_ids=input_ids, attention_mask=attention_mask)
+        output_name = self.qpc_session.output_names[0]
+        output_binding = next((binding for binding in self.qpc_session.bindings if binding.name == output_name), None)
```
quic-hemagnih, line 806: Can we cache the output_bindings. Currently we are scanning the bindings every inference runs. If we can cache it then it will be fast

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -742,34 +793,39 @@ def cloud_ai_100_feature_generate(
...
+                output_name: np.random.randn(*list(output_binding.dims)).astype(TORCH_TO_NUMPY_DTYPE_MAP[dtype]),
             }
             self.qpc_session.set_buffers(outputs)
-            outputs = self.qpc_session.run(inputs)
+            outputs = self.qpc_session.run(session_inputs)
         except Exception:
+            feature_dim = output_binding.dims[-1] if len(output_binding.dims) > 0 else 1
             outputs = {
-                "output": np.random.randn(self.batch_size, self.seq_len, self.qpc_session.bindings[2].dims[1]).astype(
+                output_name: np.random.randn(self.batch_size, self.seq_len, feature_dim).astype(
```
quic-hemagnih, line 820: Can we use np.empty(shape, dtype=dtype), where we are using these randomly generated numbers

tests/unit_test/models/test_modeling_auto_cpu.py
```diff
@@ -670,6 +687,64 @@ def test_init_sets_use_cache_true(self):
         qeff_decoder = QEFFAutoModel(decoder_model)
         assert qeff_decoder.model.base_model.config.use_cache is True
 
+    def test_init_applies_qwen3_kv_cache_transform(self):
+        """Qwen3 models loaded via AutoModel path are transformed to QEff wrappers."""
```
quic-hemagnih, line 691: Does this test covers sub-functions?

quic-hemagnih (changes requested):

PR #1159, opened 2026-07-07
Gemma4 E2B vision and decoder fix

nsivalin:
1. Fixed Gemma4 vision token count handling. Removed hardcoded 256-token fallback and now derive the image token count from the HF config (mm_tokens_per_image / vision_config.default_output_length). This correctly uses 280 vision tokens for Gemma4 E2B.

2. Fixed decoder bidirectional vision-mask gating. QEff was applying the bidirectional vision mask whenever mm_token_type_ids were present. The mask is now applied only when use_bidirectional_attention == "vision", matching Hugging Face behavior and avoiding incorrect masking for E2B/E4B.

3. Fixed Gemma4 vision padding attention mask handling. QEff was passing a 0/1 mask directly into vision attention, while Gemma4 attention expects an additive mask. Padding patch positions are now masked with a large negative value, matching Hugging Face behavior.

### DIFF
--- a/QEfficient/transformers/models/gemma4/modeling_gemma4.py
+++ b/QEfficient/transformers/models/gemma4/modeling_gemma4.py
@@ -469,7 +469,12 @@ class QEffGemma4TextAttention(Gemma4TextAttention):
                 if token_key_states is not None and token_value_states is not None:
                     past_key_values.shared_layers_token[self.layer_idx] = token_key_states, token_value_states
 
-        if mm_token_type_ids is not None and hidden_states.shape[1] != 1:
+        if (
+            mm_token_type_ids is not None
+            and hidden_states.shape[1] != 1
+            and getattr(self.config, "use_bidirectional_attention", None) == "vision"
+            and self.sliding_window is not None
+        ):
             attention_mask = _build_bidirectional_vision_attention_mask(
                 position_ids=position_ids,
                 mm_token_type_ids=mm_token_type_ids,
@@ -1072,7 +1077,11 @@ class QEffGemma4EncoderWrapper(nn.Module):
         super().__init__()
         self.model = model
         self.model.vision_model = self.model.model.vision_tower
-        self.mm_tokens_per_image = getattr(self.model.config, "mm_tokens_per_image", 256)
+        self.mm_tokens_per_image = getattr(
+            self.model.config,
+            "mm_tokens_per_image",
+            getattr(self.model.config.vision_config, "default_output_length", 280),
+        )
 
     def get_submodules_for_export(self) -> Type[nn.Module]:
         return {self.model.model.vision_tower.encoder.layers[0].__class__}
@@ -1083,7 +1092,13 @@ class QEffGemma4EncoderWrapper(nn.Module):
         inputs_embeds = vision_tower.patch_embedder(pixel_values, image_position_ids, padding_positions)
 
         valid_tokens = ~padding_positions
-        vision_attention_mask = (~valid_tokens).unsqueeze(1).unsqueeze(2).to(dtype=inputs_embeds.dtype)
+
+        vision_attention_mask = torch.where(
+            valid_tokens.unsqueeze(1).unsqueeze(2),
+            torch.zeros((), dtype=inputs_embeds.dtype, device=inputs_embeds.device),
+            _attention_mask_min(inputs_embeds.dtype, inputs_embeds.device),
+        )
+
         vision_attention_mask = vision_attention_mask.expand(-1, 1, inputs_embeds.shape[1], -1)
 
         hidden_states = inputs_embeds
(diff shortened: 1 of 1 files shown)


quic-hemagnih:
Fix the LINT errors

QEfficient/transformers/models/gemma4/modeling_gemma4.py
```diff
@@ -469,7 +469,12 @@ def forward(
                 if token_key_states is not None and token_value_states is not None:
                     past_key_values.shared_layers_token[self.layer_idx] = token_key_states, token_value_states
 
-        if mm_token_type_ids is not None and hidden_states.shape[1] != 1:
+        if (
+            mm_token_type_ids is not None
+            and hidden_states.shape[1] != 1
+            and getattr(self.config, "use_bidirectional_attention", None) == "vision"
+            and self.sliding_window is not None
```
qcdipankar: have we activated sliding window in this PR?

QEfficient/transformers/models/gemma4/modeling_gemma4.py
```diff
@@ -1083,7 +1092,13 @@ def forward(self, pixel_values, image_position_ids):
         inputs_embeds = vision_tower.patch_embedder(pixel_values, image_position_ids, padding_positions)
 
         valid_tokens = ~padding_positions
-        vision_attention_mask = (~valid_tokens).unsqueeze(1).unsqueeze(2).to(dtype=inputs_embeds.dtype)
+
+        vision_attention_mask = torch.where(
```
qcdipankar, line 1095: How is it different from using valid_tokens?

nsivalin: In the Hugging Face implementation, the valid token mask is first converted by create_bidirectional_mask into an additive mask (0 for valid tokens and a large negative value for padded tokens) before being passed to eager_attention_forward. In QEFF, the 0/1 mask was passed directly to eager_attention_forward, bypassing this conversion. This issue was resolved by assigning large negative values to padded tokens, aligning the behavior with the Hugging Face implementation.

QEfficient/transformers/models/gemma4/modeling_gemma4.py
nsivalin: The sliding_window parameter was there already and we used that in this condition

QEfficient/transformers/models/gemma4/modeling_gemma4.py
```diff
@@ -469,7 +469,12 @@ def forward(
                 if token_key_states is not None and token_value_states is not None:
                     past_key_values.shared_layers_token[self.layer_idx] = token_key_states, token_value_states
 
-        if mm_token_type_ids is not None and hidden_states.shape[1] != 1:
+        if (
```
quic-hemagnih, line 472: At line#621, when use_bidirectional_attention == "vision" and mm_token_type_ids is present, layer_attention_mask is set to None for all layers.
     
Now Here at line#477, the PR now rebuilds mask only if self.sliding_window is not None (sliding layers only).

Will it cause any issue in computing mask for full_attention layers?

quic-hemagnih:
Suggested fix direction - 
  1. Keep bidirectional mask path only for sliding layers.
  2. But for full_attention layers in bidirectional models, still build/apply normal causal additive mask (do not leave attention_mask=None).

QEfficient/transformers/models/gemma4/modeling_gemma4.py
quic-hemagnih: This is for Gemma4-26B MOE and Gemma4-31B model

QEfficient/transformers/models/gemma4/modeling_gemma4.py
```diff
@@ -469,7 +469,12 @@ def forward(
...
+            mm_token_type_ids is not None
+            and hidden_states.shape[1] != 1
+            and getattr(self.config, "use_bidirectional_attention", None) == "vision"
```
qcdipankar: this line will not work for 31b .Can you check if this also holds?

QEfficient/transformers/models/gemma4/modeling_gemma4.py
qcdipankar: is this necessary as all the three variants have sliding_window set to not none values also how is it in hf.Is it similar?

nsivalin: the sliding window condition is not mandatory as we have sliding window for current gemma4 models.

qcdipankar: ok in case the implementation is at per with hf you can keep it else revert to what it was if it is not required.

QEfficient/transformers/models/gemma4/modeling_gemma4.py
nsivalin: sliding_window condition is not mandatory as all gemma4 models have sliding window. I'll remove it

vjanfaza:
@nsivalin 
The changes don't support CCL and there are JIRA tickets reporting compiler errors. Please follow the same steps as other VLM models to add CCL support.

quic-hemagnih (approved):

PR #1168, opened 2026-07-08
Gpt_oss prefill: avoid packed expert-axis slicing in chunked MoE to restore MXFP6 constant compression for release1.22

abhishek-singh591:
#### Before (problematic pattern)
  In `QEffPrefillOnlyChunkedGptOssMLP.forward`, the prefill path did:

  1. Repacked expert tensors into grouped views:
     - `W_g/W_u/W_d` and biases were reshaped to `[num_nsp, local_experts, ...]`.
  2. Sliced that grouped axis inside the `local_slot` loop:
     - `W_g[:, local_slot]`, `W_u[:, local_slot]`, etc.

  This matches the compiler pattern we observed:
  - input like `float16<16 x 2 x 2880 x 2880>`
  - then two slices on axis-1 (index `0` and `1`)

  So even with logically split gate/up params, this path reintroduced a packed-then-sliced representation during subfunction export.

  #### After (fixed pattern)
  The updated code now:

  1. Keeps routing weights in flat per-expert layout:
     - `routing_weights_by_expert = routing_weights.transpose(0, 1)` → `[num_experts, T]`
  2. Computes explicit contiguous expert ranges per `local_slot`:
     - `slot_start = local_slot * num_nsp`
     - `slot_end = slot_start + num_nsp`
  3. Passes per-slot expert blocks directly:
     - `self.experts.gate_proj[slot_start:slot_end]`
     - `self.experts.up_proj[slot_start:slot_end]`
     - `self.experts.down_proj[slot_start:slot_end]`
     - same pattern for biases and routing weights

  So we no longer materialize `[num_nsp, local_experts, ...]` temporaries and no longer slice that axis inside the subfunction body.

### DIFF
--- a/QEfficient/generation/vlm_generation.py
+++ b/QEfficient/generation/vlm_generation.py
@@ -737,6 +737,7 @@ class VisionLanguageGeneration(QEffTextGenerationBase):
         generation_len: int = None,
         stream: List[str] = None,
     ):
+
         exec_batch_size = self.batch_size
         max_gen_length = self._ctx_len if not generation_len else max(self._ctx_len, generation_len)
         self.initialize_decode_inputs(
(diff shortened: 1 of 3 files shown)


quic-rishinr:
Changes should be cherrypicked once the PR is merged on mainline. Closing the PR

PR #1173, opened 2026-07-10
Add reported reproducer config test suite

anujgupt-github:
## Summary
- Add a pre-merge reported reproducer config suite under `tests/reproducer_configs/`.
- Capture reported import/download/export/compile/inference cases as explicit `RegressionScenario` entries.
- Run the real QEfficient stage for active scenarios instead of placeholder option checks.
- Use `QEFF_REPRODUCER_SCENARIO` for targeted local runs.
- Add `tests/reproducer_configs/README.md` with the common run commands.

## Notes
- Compile-stage scenarios either run as reduced tests or stay gated behind official-model/full-model settings.
- MDP scenarios generate local compiler dump/intersection artifacts instead of relying on reporter-machine paths.
- Compile kwargs are checked before long compile jobs start.
- The runtime report records the Hugging Face cache selected for the run.
- The reduced run still emits an existing non-fatal qconfig warning: `Object of type PosixPath is not JSON serializable`; QPC `programqpc.bin` artifacts were still produced and validated.

## Validation
- Ran lint/format checks for the reproducer test file.
- Checked full test collection for the reproducer suite.
- Checked targeted collection with `QEFF_REPRODUCER_SCENARIO=hf-transfer-import-order`.
- Ran catalog guard tests for scenario count, uniqueness, and active scenario coverage.
- Earlier full reproducer-suite run passed with reduced scenarios: `18 passed, 34 skipped`.
- Earlier model quickcheck gate passed: `185 passed, 3 skipped`.


quic-hemagnih (changes requested):
Overall the framework looks very nice and informatime, few suggestions -
Can we add following in the scenario details -

- JIRA# - would be good for future reference
- PR# - would be good for furture reference

Currently we are categorizing  issues based on stages but there are many issues which are reported let say due to 

- Accuracy degradation, 
- Memory issues (RAM increase, QPC bloat up) etc.
 
how are we planning to categorize these issue?

anujgupt-github:
@quic-hemagnih thanks!.

Regarding internal JIRA#, I intentionally didn't want to add those. We can have a JIRA to PR mapping in internal database.

Issues related to accuracy, ONNX Export RAM bloat will require full model execution. By default, those tests are disabled but can be added.
Check the test scenario here
`name="qwen3vl-moe-memory-after-prefill`

We can enable couple of envvars to avoid skipping those models. There is a skip_reason which can be used to set which models require full model and which don't
See here: https://github.com/quic/efficient-transformers/pull/1173/changes#diff-ec050fc650d48b6106c7c5c19d99f2c4f5622bc51d0e531c9110c60ac728919bR961

quic-hemagnih (approved):
LGTM

quic-rishinr:
CI-Ready

quic-rishinr:
CI-Ready

quic-rishinr (approved):

PR #1176, opened 2026-07-12
Removed sliding window condition

nsivalin:
- Removed the sliding window condition from gemma4


quic-hemagnih:
We have a workaround for this issue in the release 1.22.
use the below command to get better accuracy for Gemma4-26B and 31B models before Qeff installation. (or pip install again after the command)
sed -i '476d' Qefficient/transformers/models/gemma4/modeling_gemma4.py

PR #1178, opened 2026-07-13
feat-dynamo(0713): Dynamo based ONNX export with subfunction support for Causal-LM models

vbaddi:
## Summary

This PR adds `dynamo=True` export support for CausalLM models, enabling `torch.export`-based ONNX export with optional repeated-subgraph subfunction extraction (`use_onnx_subfunctions=True`). It also includes a suite of robustness fixes to the ONNX transform pipeline discovered during integration.

  ### What changed

  **New infrastructure**
  - `QEfficient/utils/custom_op_utils.py`: `select_interface` helper that switches between the `torch.autograd.Function` path (TorchScript/eager) and `torch.ops.qefficient.*` path (dynamo) at call time, enabling the same model code to work under both exporters without code duplication.
  - `QEfficient/utils/export_utils.py`: `export_wrapper` refactored into four clearly separated export paths (plain TorchScript, TorchScript+subfunctions, flat dynamo, dynamo+subfunctions). `use_dynamo` is now an explicit parameter throughout rather than being re-read from `kwargs`. TorchScript monkey-patches are guarded behind `if not use_dynamo:` so they are never applied on the dynamo path. Added PyTorch version guard that raises a `RuntimeError` with install instructions when `use_dynamo=True` and torch < 2.13.
  - `QEfficient/utils/torch_patches.py`: Added `temporarily_disable_nested_compile_regions` context manager (Path 3) that unwraps static `@torch.compiler.nested_compile_region` decorators for flat-dynamo export, and `temporarily_enable_nested_compile_regions` reserved for future use.
  - `QEfficient/base/modeling_qeff.py`: 
    - `_prune_unused_fake_initializers` — removes `FakeTensor`-valued initializers that dynamo injects but are not referenced in the graph, preventing downstream ONNX load failures.
    - `_export` extended with `dynamic_shapes` parameter and full dynamo export path: sets `dynamo=True`, injects `custom_translation_table` mapping all `torch.ops.qefficient.*` ops to their ONNX counterparts, and reorders `example_inputs` / `dynamic_shapes` to match the model's `forward` signature before passing to `torch.onnx.export`.
  - `QEfficient/base/onnx_transforms.py`: Two new transforms for the dynamo+subfunctions path:
    - `PreserveNestedCacheRetainedStateTransform` — exposes CtxScatter outputs from repeated-subgraph function bodies as explicit ONNX outputs and renames KV cache graph inputs from `past_key.N_RetainedState` → `past_key.N` so the QAIC compiler can pair inputs and outputs correctly.
    - `RenameRepeatedSubgraphTransform` — renames `repeated_subgraphN` ONNX Local Functions to the actual decoder layer class names, with multi-pattern matching across known PyTorch naming conventions and `logger.warning` on mismatch or count mismatch.
  - `QEfficient/transformers/models/modeling_auto.py`: `convert_dynamic_axes_to_dynamic_shapes` converts ONNX `dynamic_axes` dicts to `torch.export` `Dim`-keyed `dynamic_shapes` dicts, handling nested `past_key_values`, compressed KV (DeepseekV3), and CB pool dimensions.
  - `QEfficient/utils/constants.py`: ONNX export opset bumped from **17 → 18**, required by the dynamo ONNX exporter.
  - `pyproject.toml`: Updated to torch **2.13.0** stable CPU wheels (minimum required for dynamo export); explicit per-python per-arch wheel URLs for x86_64 and aarch64. `onnxscript` updated to `0.6.2`, `compressed-tensors` updated to `0.17.0`.

  **CausalLM model wrappers** (`llama`, `mistral`, `qwen2`, `qwen3`, `gemma`, `gemma2`, `gpt2`, `gpt_bigcode`, `gptj`, `falcon`, `mpt`, `phi`, `phi3`, `codegen`, `starcoder2`, `granite`, `granitemoe`, `mixtral_moe`, `qwen3_moe`, `olmo2`, `glm4_moe`, `grok_1`, `deepseek_v3`, `gpt_oss`, `llama_swiftkv`):
  - `QEffDecoderLayer.forward` decorated with `@torch.compiler.nested_compile_region` so dynamo emits repeated-subgraph ONNX Local Functions instead of inlining each layer individually.
  - `torch.tensor(MIN_MASKED_ATTENTION_VALUE)` replaced with `torch.full_like(attn_weights, ...)` throughout — removes a scalar constant that dynamo cannot trace as a dynamic value and that caused shape inference failures during export.
  - `select_interface` wired into all custom op call sites (`CtxScatter`, `CtxGather`, `CustomRMSNorm`, and CB/3D variants) so both the TochTrace and dynamo paths dispatch to the correct backend op without branching in model code.


vbaddi:
This review needs to be completed by end of this week to have this in mainline.
@quic-rishinr @ochougul @quic-amitraj can you pls help review this asap and provide feedback. thanks

Re: Backward compatibility, we can revert the torch packages from _.toml_ and have it specific to 2.7.0 and dynamo specific users can enable required packages (torch, onnxscript, compressed-tensors) offline for now. @smedhe

QEfficient/base/modeling_qeff.py
```diff
@@ -325,6 +356,8 @@ def _export(
         export_dir: Optional[str] = None,
         offload_pt_weights: bool = True,
         prefill_only: Optional[bool] = False,
+        use_dynamo: bool = False,
+        dynamic_shapes: Optional[Dict[str, Dict[int, any]]] = None,
```
quic-amitraj: `any` here is Python's builtin function, not a type alias. Should be `Any` from `typing`:

```python
dynamic_shapes: Optional[Dict[str, Dict[int, Any]]] = None
```

`Any` needs to be added to the `from typing import ...` line as well.

QEfficient/base/modeling_qeff.py
```diff
@@ -433,19 +466,98 @@ def _resolve_pkv_names(layer_idx, layer_state):
...
-                    (),
-                    str(onnx_path),
-                    kwargs=example_inputs,
-                    input_names=input_names,
-                    output_names=output_names,
-                    dynamic_axes=dynamic_axes,
-                    opset_version=constants.ONNX_EXPORT_OPSET,
-                    **export_kwargs,
-                )
+            export_kwargs = {} if export_kwargs is None else export_kwargs
```
quic-amitraj: `export_kwargs` comes from `**export_kwargs` in the function signature — it is always a `dict`, never `None`. This guard is dead code and can be removed.

quic-amitraj: `export_kwargs` comes from `**export_kwargs` in the function signature — it is always a `dict`, never `None`. This guard is dead code and can be removed.

QEfficient/base/modeling_qeff.py
```diff
@@ -433,19 +466,98 @@ def _resolve_pkv_names(layer_idx, layer_state):
...
+
+            if use_dynamo:
+                dynamic_axes = None
```
quic-amitraj: Setting `dynamic_axes = None` here and then referencing it again in `transform_kwargs` further down means the transform pipeline silently receives `None` for every dynamo export. This is easy to miss later. I'd make it explicit at the point of use:

```python
transform_kwargs = {
    "onnx_base_dir": str(export_dir) if needs_external_tensor_data else None,
    "model_name": self.model_name,
    "dynamic_axes": None if use_dynamo else dynamic_axes,  # dynamo uses dynamic_shapes, not axes
}
```

QEfficient/base/modeling_qeff.py
```diff
@@ -433,19 +466,98 @@ def _resolve_pkv_names(layer_idx, layer_state):
...
+                export_kwargs = dict(export_kwargs)
+                export_kwargs.setdefault("report", False)
+                export_kwargs.setdefault("optimize", False)
+                export_kwargs["dynamo"] = True
+                export_kwargs["custom_translation_table"] = {
```
quic-amitraj: This 14-entry dict is constructed on every call to `_export()`. None of the entries are dynamic - `torch.ops.*` attributes resolve at import time. Move it to a module-level constant so it is built once and is easy to audit:

```python
_DYNAMO_CUSTOM_OP_TABLE: Dict = {
    torch.ops.qefficient.rms_norm.default:                   CustomRMSNorm,
    torch.ops.qefficient.ctx_scatter.default:                CtxScatter,
    # ...
}

# in _export():
export_kwargs["custom_translation_table"] = {
    **(export_kwargs.pop("custom_translation_table", None) or {}),
    **_DYNAMO_CUSTOM_OP_TABLE,
}
```

QEfficient/base/modeling_qeff.py
```diff
@@ -433,19 +466,98 @@ def _resolve_pkv_names(layer_idx, layer_state):
...
+                        input_names=input_names,
+                        output_names=output_names,
+                        dynamic_axes=dynamic_axes,
+                        dynamic_shapes=dynamic_shapes,
+                        opset_version=constants.ONNX_EXPORT_OPSET,
+                        **export_kwargs,
+                    )
+                    if onnx_program is None:
+                        raise RuntimeError("torch.onnx.export returned None for dynamo export")
+                    _prune_unused_fake_initializers(onnx_program)
```
quic-amitraj: All other ONNX graph mutations in this codebase are encapsulated as `BaseOnnxTransform` subclasses and run through `OnnxTransformPipeline`. Calling this inline bypasses the pipeline, making it invisible to logging and impossible to selectively disable. Consider promoting it to a `PruneFakeInitializersTransform(BaseOnnxTransform)` in `onnx_transforms.py` and injecting it into the dynamo transform list from `_setup_onnx_subfunctions()`.

QEfficient/base/modeling_qeff.py
```diff
@@ -433,19 +466,98 @@ def _resolve_pkv_names(layer_idx, layer_state):
...
+                with layerwise_safe_onnx_export_patches():
+                    torch.onnx.export(
+                        self.model,
+                        (),
+                        str(onnx_path),
+                        kwargs=example_inputs,
+                        input_names=input_names,
+                        output_names=output_names,
+                        dynamic_axes=dynamic_axes,
+                        dynamo=False,
```
quic-amitraj: `dynamo=False` is already the default for `torch.onnx.export` — passing it explicitly here adds noise without clarity. Can be removed.

QEfficient/base/modeling_qeff.py
```diff
@@ -433,19 +466,98 @@ def _resolve_pkv_names(layer_idx, layer_state):
...
+                        opset_version=constants.ONNX_EXPORT_OPSET,
+                        **export_kwargs,
+                    )
             logger.info("PyTorch export successful")
             _ = self._offload_model_weights(offload_pt_weights)
```
quic-amitraj: The return value of `_offload_model_weights` is not used here. The `_ =` assignment serves no purpose — just call the method directly:

```python
self._offload_model_weights(offload_pt_weights)
```

QEfficient/base/modeling_qeff.py
```diff
@@ -55,6 +73,19 @@
 from QEfficient.utils.export_utils import export_wrapper
 from QEfficient.utils.torch_patches import layerwise_safe_onnx_export_patches
 
+
+def _prune_unused_fake_initializers(onnx_program) -> None:
```
quic-amitraj: nit: All other ONNX graph mutations in this codebase are encapsulated as `BaseOnnxTransform` subclasses and run through `OnnxTransformPipeline`. Calling this inline bypasses the pipeline, making it invisible to logging and impossible to selectively disable. Consider promoting it to a `PruneFakeInitializersTransform(BaseOnnxTransform)` in `onnx_transforms.py` and injecting it into the dynamo transform list from `_setup_onnx_subfunctions()`.

QEfficient/base/modeling_qeff.py
```diff
@@ -433,19 +466,98 @@ def _resolve_pkv_names(layer_idx, layer_state):
...
-                    kwargs=example_inputs,
-                    input_names=input_names,
-                    output_names=output_names,
-                    dynamic_axes=dynamic_axes,
-                    opset_version=constants.ONNX_EXPORT_OPSET,
-                    **export_kwargs,
-                )
+            export_kwargs = {} if export_kwargs is None else export_kwargs
+
+            if use_dynamo:
```
quic-amitraj: `_export()` is already a long method and this PR adds ~50 lines of dynamo-specific logic inline. The dynamo and TorchScript branches are independent enough that each deserves its own private method. This makes `_export()` a clean orchestrator and makes each backend easier to read, test, and eventually remove.

Suggested split:
```python
def _export(self, ..., use_dynamo=False, ...):
    # shared: build input/output names
    ...
    if use_dynamo:
        self._export_via_dynamo(onnx_path, example_inputs, input_names,
                                output_names, dynamic_shapes, export_kwargs)
    else:
        self._export_via_torchscript(onnx_path, example_inputs, input_names,
                                     output_names, dynamic_axes, export_kwargs)
    # shared: load ONNX, apply transforms
    ...

def _export_via_dynamo(self, onnx_path, example_inputs, input_names,
                       output_names, dynamic_shapes, export_kwargs):
    """Export via torch.export (dynamo=True) with custom op translation."""
    ...

def _export_via_torchscript(self, onnx_path, example_inputs, input_names,
                             output_names, dynamic_axes, export_kwargs):
    """Export via TorchScript symbolic tracing (dynamo=False)."""
    ...
```
This also makes Phase 3 (TorchScript removal) a clean delete of `_export_via_torchscript` and the `else` branch — no surgery needed on the rest of `_export()`.

quic-amitraj (changes requested):

QEfficient/customop/ctx_scatter_gather.py
```diff
@@ -217,7 +217,7 @@ def symbolic(g: torch.Graph, data: torch.Value, ctx_indices: torch.Value) -> tor
 
 @onnxscript.script(onnxscript.values.Opset("com.qualcomm.cloud", 1))
 def CtxGather(
-    data: onnxscript.FLOAT, ctx_indices: onnxscript.INT32, comp_ctx_len: onnxscript.INT32
+    data: onnxscript.FLOAT, ctx_indices: onnxscript.INT32, comp_ctx_len: onnxscript.INT64
```
quic-amitraj, line 221: Why its needed?

QEfficient/customop/rms_norm.py
```diff
@@ -51,7 +57,8 @@ def __init__(self, hidden_size, eps=1e-05):
         self.weight = torch.nn.Parameter(torch.ones(hidden_size))
 
     def forward(self, hidden_states):
```
quic-amitraj, line 55: `CustomRMSNormAIC` gets this right — `select_interface()` lives inside the module's `forward()`, so every caller is completely unaware of the dispatch. This is exactly the pattern we should follow.

However, the same principle was **not applied** to the `torch.autograd.Function` subclasses (`CtxScatterFunc`, `CtxGatherFunc`, `CtxScatterFunc3D`, and all their variants in `ctx_scatter_gather.py`). Instead, the PR places `select_interface()` call-sites directly in 25+ model files:

```python
# modeling_glm4_moe.py, modeling_qwen3_moe.py, ... (25+ files)
matched_idx = select_interface(CtxScatterFunc3DInt.apply, torch.ops.qefficient.ctx_scatter_3d_int)(
    matched_idx.unsqueeze(-1), scatter_pos, token_idx.unsqueeze(-1),
).squeeze(-1)
```

Model files should contain only PyTorch forward logic — they should not know which export backend is active. The fix is to move the dispatch into each op's `forward()`, following the pattern established here:

```python
# customop/ctx_scatter_gather.py
class CtxScatterFunc3DInt(torch.autograd.Function):
    @staticmethod
    def _eager(data, position_ids, updates):
        # original forward() body moved here
        ...

    @staticmethod
    def forward(data, position_ids, updates):
        return select_interface(
            CtxScatterFunc3DInt._eager,
            torch.ops.qefficient.ctx_scatter_3d_int,
        )(data, position_ids, updates)

    @staticmethod
    def symbolic(g, data, position_ids, updates):
        ...  # TorchScript path — unchanged

# modeling_glm4_moe.py — restored to pre-PR, zero changes
matched_idx = CtxScatterFunc3DInt.apply(
    matched_idx.unsqueeze(-1), scatter_pos, token_idx.unsqueeze(-1),
).squeeze(-1)
```

With this, all 25+ model files revert to their original state and the dispatch decision is co-located with the op definition — exactly as done here for `CustomRMSNormAIC`.

QEfficient/utils/custom_op_utils.py
```diff
@@ -0,0 +1,367 @@
...
+#
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# -----------------------------------------------------------------------------
+
+import torch
+
+
+def select_interface(eager_impl, custom_op_impl):
```
quic-amitraj: This file belongs in `QEfficient/customop/`, not `QEfficient/utils/`.

`utils/` is the home for generic infrastructure — export hashing, path helpers, runtime utilities. This file contains two distinct things:

1. **Custom op definitions** (`torch.library.custom_op` + `register_fake`) — these are op implementations and belong alongside the existing `torch.autograd.Function` ops in `customop/`.
2. **`select_interface()`** — a dispatch helper for those ops, which also belongs in `customop/`.

Having the dynamo-path ops in `utils/` and the TorchScript-path ops in `customop/` means there is no single place to find all custom op implementations. The proposed split:

```
QEfficient/customop/
    ctx_scatter_gather.py      # torch.autograd.Function (TorchScript path) — unchanged
    rms_norm.py                # torch.autograd.Function — unchanged
    dynamo_ops.py              # this file's torch.library.custom_op definitions
    utils.py                   # select_interface() helper
    __init__.py                # export everything
```

This makes `customop/` the single authoritative registry for all custom op implementations regardless of which export backend they serve.

QEfficient/utils/custom_op_utils.py
```diff
@@ -0,0 +1,367 @@
...
+
+
+@torch.library.custom_op("qefficient::ctx_gather", mutates_args=())
+def ctx_gather_op(data: torch.Tensor, ctx_indices: torch.Tensor, comp_ctx_len: int) -> torch.Tensor:
+    batch_indices = torch.arange(data.shape[0]).view(-1, 1, 1)
+    head_indices = torch.arange(data.shape[1]).view(1, -1, 1)
+    return data[batch_indices, head_indices, ctx_indices]
+
+
+# @ctx_gather_op.register_fake
```
quic-amitraj: Commented-out code should not be left in a merged PR. This is an alternative `register_fake` implementation for `ctx_gather_op` that was superseded by the implementation that follows (lines 100–126). Either delete it entirely, or if the reasoning behind the alternative is worth preserving, move it to a code comment explaining why the current approach was chosen over this one.

QEfficient/base/onnx_transforms.py
```diff
@@ -135,9 +136,24 @@ def apply(cls, model: ModelProto) -> bool:
             if proto.name not in existing:
                 model.functions.append(proto)
                 op_applied = True
-
+                cls._ensure_opset_imports(model, proto.domain, 1)
+        cls._propagate_function_opset_imports(model)
         return op_applied
 
+    @staticmethod
+    def _ensure_opset_imports(container, domain: str, version: int) -> None:
```
quic-amitraj: `container` is untyped but accepts either a `ModelProto` or an `onnx.FunctionProto`. `Union` is already imported in this file:

```python
@staticmethod
def _ensure_opset_imports(container: Union[ModelProto, onnx.FunctionProto], domain: str, version: int) -> None:
```

QEfficient/base/onnx_transforms.py
```diff
@@ -255,6 +271,236 @@ def apply(cls, model: ModelProto, layer_idx=0) -> bool:
...
+            # Collect scatter nodes that write back the KV cache.
+            # Sort by first-input name so the key scatter reliably precedes the
+            # value scatter: dynamo names function-body args generically (arg7_1
+            # etc.), so we sort by the scatter output name instead — dynamo
+            # preserves "key"/"value" in output tensor names even when input
+            # argument names are opaque.
+            scatter_nodes = [
+                fn_node for fn_node in fn.node if fn_node.op_type in cls._SCATTER_OP_TYPES and fn_node.output
+            ]
+            if len(scatter_nodes) != 2:
```
quic-amitraj, line 338: Silent skip with no diagnostic. When this fires on a model with a different KV layout (compressed KV, sliding window, etc.) the transform silently produces no output and there is nothing in the logs to explain why. At minimum a debug-level log would make failures traceable:

```python
if len(scatter_nodes) != 2:
    logger.debug(
        "PreserveNestedCacheRetainedStateTransform: node '%s' has %d scatter node(s), expected 2 — skipping.",
        node.name, len(scatter_nodes),
    )
    continue
```

QEfficient/base/onnx_transforms.py
```diff
@@ -255,6 +271,236 @@ def apply(cls, model: ModelProto, layer_idx=0) -> bool:
...
+                continue
+
+            def _scatter_sort_key(n):
```
quic-amitraj: `_scatter_sort_key` does not capture any loop variable — it is a purely stateless function. Redefining it on every iteration of `for node in graph.node:` is wasteful and misleading. Move it to a `@staticmethod` on the class:

```python
@staticmethod
def _scatter_sort_key(n) -> int:
    output_name = n.output[0] if n.output else ""
    if "key" in output_name:   return 0
    if "value" in output_name: return 1
    return 2
```

QEfficient/base/onnx_transforms.py
```diff
@@ -255,6 +271,236 @@ def apply(cls, model: ModelProto, layer_idx=0) -> bool:
...
+        for node in nodes:
+            yield node
+            for attr in node.attribute:
+                if attr.HasField("g"):
+                    yield from cls._iter_all_nodes(attr.g.node)
+
+    @classmethod
+    def apply(cls, model: ModelProto, target_classnames: Optional[List[str]] = None, **kwargs) -> bool:
+        target_classnames = [name for name in (target_classnames or []) if name]
+        if not target_classnames:
```
quic-amitraj, line 455: `target_classnames` is the entire purpose of this transform. An empty list means either the caller omitted it from `kwargs` or `get_submodules_for_export()` returned nothing — both cases are likely misconfiguration, not intentional behaviour. Silently returning `False` makes this hard to diagnose. Add a warning:

```python
if not target_classnames:
    logger.warning(
        "RenameRepeatedSubgraphTransform: target_classnames is empty — transform is a no-op. "
        "Check that get_submodules_for_export() returns the decoder layer classes."
    )
    return False
```

QEfficient/base/onnx_transforms.py
```diff
@@ -255,6 +271,236 @@ def apply(cls, model: ModelProto, layer_idx=0) -> bool:
...
+            old_to_new[fn.name] = candidate
+
+        if not old_to_new:
+            return False
+
+        for fn in model.functions:
+            if fn.name in old_to_new:
+                fn.name = old_to_new[fn.name]
+
+        def _rename_op_types(nodes):
```
quic-amitraj: `_rename_op_types` only closes over `old_to_new`. It can be a `@staticmethod` taking it as an explicit parameter, which is consistent with `_iter_all_nodes` (already a classmethod) and makes it independently testable:

```python
@staticmethod
def _rename_op_types(nodes, old_to_new: Dict[str, str]) -> None:
    for node in RenameRepeatedSubgraphTransform._iter_all_nodes(nodes):
        if node.op_type in old_to_new:
            node.op_type = old_to_new[node.op_type]
```

QEfficient/base/onnx_transforms.py
```diff
@@ -255,6 +271,236 @@ def apply(cls, model: ModelProto, layer_idx=0) -> bool:
...
+            "CtxScatter",
+            "CtxScatterCB",
+            "CtxScatter3D",
+            "CtxScatter3DInt",
+            "CtxScatterCB3D",
+        }
+    )
+
+    @classmethod
+    def apply(cls, model: ModelProto) -> bool:
```
quic-amitraj, line 311: `BaseOnnxTransform.apply()` declares `-> Tuple[ModelProto, bool]` but both new transforms return only `bool`. This is a pre-existing inconsistency in the file (`CustomOpTransform` also returns `bool`), but this PR adds two more deviating implementations.

The base class signature should be updated to reflect actual usage across the file:

```python
@classmethod
def apply(cls, model: ModelProto, **kwargs) -> bool:
    raise NotImplementedError("Use subclasses for ONNX transform")
```

This is a small change but removes a misleading contract that every new transform has to silently ignore.

QEfficient/utils/export_utils.py
```diff
@@ -172,54 +363,85 @@ def _setup_onnx_subfunctions(qeff_model, args, kwargs):
         "The subfunction feature is experimental. Please note that using compile "
         "consecutively with and without subfunction may produce inconsistent results."
     )
-
-    # Apply torch patches for subfunction support
-    apply_torch_patches()
+    qeff_model._use_onnx_subfunctions = True
```
quic-amitraj, line 422: `_use_onnx_subfunctions` and `hash_params` are mutated here but never restored by `_cleanup_onnx_subfunctions`. If the export fails after this point, a subsequent export on the same model object will use a stale hash and may hit a wrong cache entry. Add the rollback to the cleanup path.

QEfficient/utils/export_utils.py
```diff
@@ -9,15 +9,140 @@
...
+        first_cls = modules[0].__class__
+        if all(isinstance(item, first_cls) for item in modules):
+            return first_cls
+    return None
+
+
+def _discover_submodule_classes_for_export(model):
+    discovered = set()
+
+    attr_paths = [
```
quic-amitraj: `attr_paths` and `repeated_suffixes` (line 97) are static — they never change at runtime and are reconstructed on every call. Move both to module-level constants:

```python
_KNOWN_DECODER_LAYER_ATTR_PATHS = ("layers", "h", "model.layers", ...)
_KNOWN_DECODER_LAYER_SUFFIXES = (".layers", ".layer", ".h", ...)
```

This also makes them easy to find and extend when onboarding a new model family.

quic-amitraj:
Please remove all the commented code from all the files.

QEfficient/utils/generate_inputs.py
```diff
@@ -138,7 +138,6 @@ def prepare_pytorch_inputs(self):
             pkv = (past_key, past_value)
             past_key_values.append(pkv)
         inputs["past_key_values"] = tuple(past_key_values)
-
```
quic-amitraj, line 141: Why needed?, revert it.

examples/image_text_to_text/models/gemma_vision/gemma3/gemma3_example.py
```diff
@@ -53,6 +56,7 @@
         skip_vision=True,
```
quic-amitraj, line 53: Not passing dynamo flag?

examples/image_text_to_text/models/qwen2_5_vl/basic_inference.py
```diff
@@ -88,6 +89,7 @@
         mxint8_kv_cache=True,
         aic_enable_depth_first=True,
         mos=1,
```
quic-amitraj, line 90: Update example script.

examples/text_generation/test_subfunction_dynamo.py
```diff
@@ -0,0 +1,50 @@
...
+    prompt_len=32,
+    ctx_len=128,
+)
+
+# PyTorch (KV) output
+hf_model = AutoModelForCausalLM.from_pretrained(model_name, config=config)
+hf_tokens = runner.run_hf_model_on_pytorch(hf_model)
+print(hf_tokens)
+
+qeff_model = QEFFAutoModelForCausalLM.from_pretrained(model_name, config=config)
```
quic-amitraj: Do we need this file?

quic-amitraj (changes requested):

QEfficient/base/modeling_qeff.py
smedhe: Addressed the issue, similar issue was there in onnx_transform_kwargs, changed that too

QEfficient/base/modeling_qeff.py
smedhe: use_dynamo is False here, not dynamo.  I agree that in torch 2.7 by default, dynamo is False in `torch.onnx.export`. But by default in torch 2.13, it uses `dynamo=True`. Putting `dynamo=False` guarantees that in the legacy path dynamo tracing is not done.

examples/image_text_to_text/models/gemma_vision/gemma3/gemma3_example.py
smedhe: dynamo will be enabled for VLMs in next sprint

examples/image_text_to_text/models/qwen2_5_vl/basic_inference.py
smedhe: dynamo will be enabled for VLMs in next sprint

examples/text_generation/test_subfunction_dynamo.py
smedhe: We can directly add `use_dynamo` flag in the example script for causal lm and remove this file

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3730,6 +3847,7 @@ def export(
         layerwise: bool = False,
         layerwise_window_size: int = 1,
         kv_cache_prefix: Optional[str] = None,
+        use_dynamo: bool = False,
```
ochougul: make this parameter name `dynamo` instead of `use_dynamo` let's be consistent with torch.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3782,10 +3902,22 @@ def export(
             block_size = -(-seq_len // max_blocks)
             seq_len = block_size * max_blocks
         fbs: int = constants.ONNX_EXPORT_EXAMPLE_FBS
-
+        if use_dynamo:
+            seq_len = max(2, seq_len)
+            fbs = max(2, fbs)
+            if getattr(self.model.config, "model_type", None) == "gpt_oss" and not self.continuous_batching:
+                bs = 1
```
ochougul: why is this?

QEfficient/utils/custom_op_utils.py
ochougul: I agree.

QEfficient/utils/constants.py
```diff
@@ -98,7 +98,7 @@ def get_models_dir():
 ONNX_EXPORT_EXAMPLE_MAX_TOP_K_IDS = 512
 ONNX_EXPORT_EXAMPLE_TOP_PS = 0.80
 ONNX_EXPORT_EXAMPLE_MIN_PS = 0.99
-ONNX_EXPORT_OPSET = 17
+ONNX_EXPORT_OPSET = 18
```
ochougul: Is this tested for the Auto classes for which we are not enabling dynamo?

examples/text_generation/test_subfunction_dynamo.py
ochougul, line 1: examples should not have a test file

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -4029,12 +4161,18 @@ def _legacyify_cache(obj):
                 _layerwise_cache_probe=kwargs.get("_layerwise_cache_probe", False),
             )
         else:
+            dynamic_shapes = None
+            if use_dynamo:
+                dynamic_shapes = self.convert_dynamic_axes_to_dynamic_shapes(dynamic_axes)
+
```
ochougul: this call should be done inside the `_export` method, with same condition, keep interface simple. just pass `dynamo=True/False` here, and if True inside do the conversion of dynamic shapes.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3782,10 +3902,22 @@ def export(
...
             seq_len = block_size * max_blocks
         fbs: int = constants.ONNX_EXPORT_EXAMPLE_FBS
-
+        if use_dynamo:
+            seq_len = max(2, seq_len)
+            fbs = max(2, fbs)
+            if getattr(self.model.config, "model_type", None) == "gpt_oss" and not self.continuous_batching:
+                bs = 1
+            else:
+                bs = max(2, bs)
```
ochougul: if you set `seq_len = constants.ONNX_EXPORT_EXAMPLE_BATCH_SIZE` if not dynamo else 2, we may not need these many lines of code.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3843,7 +3975,7 @@ def export(
             "position_ids": {0: "batch_size", 1: "seq_len"},
         }
         if self.ccl_enabled:
-            example_inputs["comp_ctx_lengths"] = torch.randint(0, 127, (512,), dtype=torch.int64)
+            example_inputs["comp_ctx_lengths"] = torch.randint(0, 127, (seq_len,), dtype=torch.int64)
```
ochougul, line 3858: Have we tested CCL with dynamo? can you point me to the test case?

QEfficient/customop/rms_norm.py
```diff
@@ -30,6 +31,11 @@ def forward(hidden_states: torch.Tensor, weight: torch.Tensor, epsilon: float):
         hidden_states = hidden_states * torch.rsqrt(variance + epsilon)
         return weight * hidden_states
 
+    @staticmethod
+    def backward(ctx, grad_output):
+        # Not needed for inference/export
+        raise NotImplementedError("backward not supported for export")
+
```
ochougul: bloat: rmove

QEfficient/customop/rms_norm.py
ochougul: I agree

pyproject.toml
ochougul, line 1: add a `.[dynamo]` install option similar to `.[test]` 
When user does `pip install .[dynamo]` -> this should install the 2.13 torch otherwise by default keep old behaviour

ochougul:
Can we make dynamo=True as default for CausalLM ? So, our CI tests this feature thoroughly and keep it that way?

pyproject.toml
```diff
@@ -45,18 +45,25 @@ dependencies = [
...
-    "torchvision@https://download.pytorch.org/whl/cpu/torchvision-0.22.0%2Bcpu-cp39-cp39-manylinux_2_28_x86_64.whl ; python_version=='3.9' and platform_machine=='x86_64'",
-    "torchvision@https://download.pytorch.org/whl/cpu/torchvision-0.22.0%2Bcpu-cp310-cp310-manylinux_2_28_x86_64.whl ; python_version=='3.10' and platform_machine=='x86_64'",
-    "torchvision@https://download.pytorch.org/whl/cpu/torchvision-0.22.0%2Bcpu-cp311-cp311-manylinux_2_28_x86_64.whl ; python_version=='3.11' and platform_machine=='x86_64'",
-    "torchvision@https://download.pytorch.org/whl/cpu/torchvision-0.22.0%2Bcpu-cp312-cp312-manylinux_2_28_x86_64.whl ; python_version=='3.12' and platform_machine=='x86_64'",
+    "compressed-tensors==0.17.0",
+    # Specifying torch 2.13.0 stable cpu package URL per python version (torch >= 2.13 required for use_dynamo=True).
+    # aarch64 (Linux ARM)
+    "torch@https://download.pytorch.org/whl/cpu/torch-2.13.0%2Bcpu-cp310-cp310-manylinux_2_28_aarch64.whl ; python_version=='3.10' and platform_machine=='aarch64'",
```
quic-rishinr: Does torch 2.13 support python 3.10? please check the dependencies. Also the idea was install torch 2.13 when user switches to dynamo right? why we install it by default?

pyproject.toml
```diff
@@ -31,9 +31,9 @@ dependencies = [
     "onnxruntime==1.22",
     "numpy==1.26.4",
     "protobuf==6.31.0",
-    "onnxscript==0.2.5",
     # for layerwise infra
     "onnx-ir==0.2.1",  
+    "onnxscript==0.6.2",
```
quic-rishinr: Do we have approval for this?

tests/dynamo/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -----------------------------------------------------------------------------
```
quic-rishinr, line 1: is this file needed?

pyproject.toml
```diff
@@ -80,6 +87,7 @@ namespaces = false
 attr = "QEfficient.__version__"
 
 [tool.ruff]
+ignore = ["E731"]
```
quic-rishinr: why we want to add an exection?

quic-rishinr:
@smedhe @vbaddi please run a memory profiling for couple of models and also capture the export, compiler time for the models with and without the changes.

QEfficient/customop/rms_norm.py
smedhe: required for dynamo true and subfunction true use case

QEfficient/base/modeling_qeff.py
smedhe: resolved

pyproject.toml
smedhe: there are conflicts while installing the dynamo extras, giving the user the direct command to download the dynamo extras. Once I have verified that updating the dependencies do not break the other supported features, I will make this suggested change in toml

QEfficient/base/modeling_qeff.py
smedhe: resolved

QEfficient/base/modeling_qeff.py
smedhe: resolved

QEfficient/base/modeling_qeff.py
smedhe: resolved

QEfficient/base/modeling_qeff.py
smedhe: resolved

QEfficient/base/modeling_qeff.py
smedhe: resolved

QEfficient/customop/ctx_scatter_gather.py
smedhe: there was a dtype mismatch while compiling when exported via dynamo. This needs to be tested in the newer sdk versions.

QEfficient/customop/rms_norm.py
smedhe: This is done, as an effect to this, I had to add dummy backward functions in the scatter gather autograd functions

QEfficient/utils/custom_op_utils.py
smedhe: made these new files in custom op

QEfficient/base/onnx_transforms.py
smedhe: resolved

tests/dynamo/__init__.py
smedhe: deleted

QEfficient/transformers/models/modeling_auto.py
vbaddi: Agree.

QEfficient/customop/ctx_scatter_gather.py
quic-amitraj: Test with the latest sdk and check if its needed or not.

pyproject.toml
smedhe: legals have been sent for approval

QEfficient/base/modeling_qeff.py
```diff
@@ -55,6 +73,19 @@
 from QEfficient.utils.export_utils import export_wrapper
 from QEfficient.utils.torch_patches import layerwise_safe_onnx_export_patches
 
+
+def _prune_unused_fake_initializers(onnx_program) -> None:
+    initializers = onnx_program.model.graph.initializers
+    used_names = {name for node in onnx_program.model.graph for name in node.inputs}
```
quic-hemagnih: Should it be onnx_program.model.graph or onnx_program.model.graph.**node**

QEfficient/transformers/cache_utils.py
quic-hemagnih: Shouldn't we use torch.zeros_like instead of torch.tensor(0.0,...) here

QEfficient/transformers/cache_utils.py
```diff
@@ -441,41 +500,41 @@ def update_ckv(self, compressed_kv, cache_kwargs):
...
 
         ctx_indices = torch.where(invalid_mask, invalid_idx_value, ctx_indices)
 
         ctx_indices = ctx_indices.expand(batch, num_kv_heads, ctx_indices.shape[-1])
-        ckv_out = CtxGatherFuncBlockedKV.apply(ckv_out, ctx_indices)
+        ckv_out = select_interface(CtxGatherFuncBlockedKV.apply, torch.ops.qefficient.ctx_gather_blocked_kv)(
+            ckv_out, ctx_indices
+        )
 
         ckv_out = torch.where(invalid_mask.unsqueeze(-1), torch.tensor(0.0, dtype=torch.float32), ckv_out)
```
quic-hemagnih: same as above for torch.tensor(0.0)

QEfficient/transformers/cache_utils.py
```diff
@@ -485,33 +544,36 @@ def read_only_blocked_k_pe(self, start_index, end_index, cache_kwargs):
...
 
         ctx_indices = torch.where(invalid_mask, invalid_idx_value, ctx_indices)
 
         ctx_indices = ctx_indices.expand(batch, num_kv_heads, ctx_indices.shape[-1])
-        k_pe_out = CtxGatherFuncBlockedKV.apply(k_pe_out, ctx_indices)
+        k_pe_out = select_interface(CtxGatherFuncBlockedKV.apply, torch.ops.qefficient.ctx_gather_blocked_kv)(
+            k_pe_out, ctx_indices
+        )
 
         k_pe_out = torch.where(invalid_mask.unsqueeze(-1), torch.tensor(0.0, dtype=torch.float32), k_pe_out)
```
quic-hemagnih: same here also

QEfficient/base/modeling_qeff.py
```diff
@@ -1080,6 +1198,7 @@ def _compile(
                 command.append(f"-custom-IO-list-file={custom_io_yaml}")
 
         command.append(f"-aic-binary-dir={qpc_path}")
+        print(command)
```
quic-hemagnih: Do we need this?

quic-hemagnih (changes requested):
Can you add a unit test for PreserveNestedCacheRetainedStateTransform

QEfficient/base/onnx_transforms.py
```diff
@@ -255,6 +271,236 @@ def apply(cls, model: ModelProto, layer_idx=0) -> bool:
...
+
+            # Append the retained-state output names to this call node.
+            missing_outputs = [name for name in desired_outputs if name not in node.output]
+            if missing_outputs:
+                node.output.extend(missing_outputs)
+                changed = True
+
+        return changed
+
+    @staticmethod
```
quic-hemagnih, line 398: The complexity

  For a 32-layer model:

  - 32 layers × 2 (key + value) = 64 separate calls to _rename_graph_input
  - Each call does a full scan of every node in the graph
  - A 32-layer model graph has hundreds of nodes

  So the total work is: 64 calls × N nodes = 64N node scans, where N can be in the hundreds. That's O(layers × nodes) — quadratic in practice.

  ———

  Fix: Collect All Renames First, Apply in One Pass

  The insight is: all 64 renames are known before any node is touched. So instead of calling _rename_graph_input 64 times (each scanning all nodes),
  collect the full rename map first, then walk the graph exactly once.

  Collect the rename map across all layers

  In apply(), instead of calling _rename_graph_input immediately inside the per-layer loop, accumulate into a dict:

  # Before the per-node loop:
  kv_rename_map: Dict[str, str] = {}

  # Inside the per-node loop, replace the immediate _rename_graph_input call:
  for kind in ("key", "value"):
      retained_input = kv_inputs[kind]           # "past_key.3_RetainedState"
      plain_input = f"past_{kind}.{layer_idx}"   # "past_key.3"
      if retained_input.endswith("_RetainedState"):
          kv_rename_map[retained_input] = plain_input   # just record it

  After the loop over all nodes is done, kv_rename_map looks like:

  {
      "past_key.0_RetainedState":   "past_key.0",
      "past_value.0_RetainedState": "past_value.0",
      "past_key.1_RetainedState":   "past_key.1",
      "past_value.1_RetainedState": "past_value.1",
      # ... up to layer 31
  }

Apply all renames in a single pass

  Replace _rename_graph_input with a new _rename_graph_inputs_bulk that takes the whole map:

  @staticmethod
  def _rename_graph_inputs_bulk(graph: onnx.GraphProto, rename_map: Dict[str, str]) -> bool:
      if not rename_map:
          return False
      changed = False

      # 1. Rename graph-level input declarations
      for value in graph.input:
          if value.name in rename_map:
              value.name = rename_map[value.name]
              changed = True

      # 2. Rename value_info entries
      for value in graph.value_info:
          if value.name in rename_map:
              value.name = rename_map[value.name]
              changed = True

      # 3. Single pass over all nodes — apply all renames at once
      for node in graph.node:
          node.input[:] = [rename_map.get(n, n) for n in node.input]
          if any(n in rename_map for n in node.input):
              changed = True

      return changed

  Then at the end of apply(), after the per-layer loop:

  if kv_rename_map:
      changed |= cls._rename_graph_inputs_bulk(graph, kv_rename_map)

QEfficient/base/modeling_qeff.py
```diff
@@ -433,19 +466,98 @@ def _resolve_pkv_names(layer_idx, layer_state):
             dynamic_axes = {rename_map.get(k, k): v for k, v in dynamic_axes.items()}
             input_names = aligned_input_names
 
+        # Rearrange example_inputs and dynamic_shapes to follow model.forward signature
+        sig = inspect.signature(self.model.forward)
+
+        ordered_example_inputs = OrderedDict()
```
quic-hemagnih: The individual dict lookups (name in example_inputs, name in ordered_example_inputs) are each O(1). So why is the overall complexity called O(N²)?

  The issue is not the dict lookups themselves. It is the combination of the two passes and what they touch:

 Pass 1: iterates sig.parameters — size P (number of forward args)

  Pass 2: iterates example_inputs — size E (number of inputs, including all KV cache entries)

  For a 32-layer model, example_inputs contains:

  - input_ids, position_ids, attention_mask, batch_index → ~4 regular inputs
  - past_key.0 … past_key.31 → 32 key cache inputs
  - past_value.0 … past_value.31 → 32 value cache inputs

  That's ~68 entries in example_inputs. For a 70B model with 80 layers it's ~164 entries.

  But sig.parameters for forward() has only ~5–8 entries (the actual named parameters: input_ids, position_ids, past_key_values, etc.) — not the
  flattened KV cache names, because those are packed into past_key_values in the signature.

  So Pass 1 finds only the ~5 signature params. Pass 2 then iterates all 68+ example_inputs entries and for each one does name not in
  ordered_example_inputs — a membership check on a dict that grows as the loop runs.

  The real inefficiency is that Pass 2 is doing redundant work: it re-visits every key in example_inputs to find the ones that Pass 1 missed. For 100+ KV
  cache inputs, Pass 2 iterates 100+ times, and for each iteration it does a membership check against ordered_example_inputs. This is O(E) work done
  twice — once in Pass 1 and once in Pass 2 — when a single pass could do it.

  Additionally, OrderedDict is used when a plain dict suffices (Python 3.7+ dicts preserve insertion order).

  ———

 The Fix: Two Dict Comprehensions, One Pass Each

  The entire block can be replaced with two comprehensions that each touch every key exactly once:

  sig = inspect.signature(self.model.forward)
  sig_keys = sig.parameters.keys()   # a KeysView — O(1) membership checks

  # Single pass: signature-ordered keys first, then extras
  example_inputs = {
      **{k: example_inputs[k] for k in sig_keys if k in example_inputs},
      **{k: v for k, v in example_inputs.items() if k not in sig_keys},
  }

  if dynamic_shapes is not None:
      dynamic_shapes = {
          **{k: dynamic_shapes[k] for k in sig_keys if k in dynamic_shapes},
          **{k: v for k, v in dynamic_shapes.items() if k not in sig_keys},
      }

  That's it. No OrderedDict, no intermediate variables, no three separate loops.

quic-hemagnih (changes requested):

QEfficient/base/modeling_qeff.py
smedhe: No , this function operates on torch.onnx.ONNXProgram (dynamo export), which uses the onnx.ir API, not the protobuf ModelProto API used elsewhere

QEfficient/base/modeling_qeff.py
```diff
@@ -97,6 +116,24 @@ def _restore_output_names_exact(model: onnx.ModelProto, output_names: List[str])
         _rename_graph_value(model.graph, current_name, expected_name)
 
 
+_DYNAMO_CUSTOM_OP_TABLE: Dict = {
```
quic-amitraj: Please move this to constants.

QEfficient/customop/ctx_scatter_gather.py
```diff
@@ -80,16 +92,27 @@ def CtxScatter3D(data: onnxscript.FLOAT, position_ids: onnxscript.INT32, updates
...
     @staticmethod
-    def forward(data: torch.Tensor, position_ids: torch.Tensor, updates: torch.Tensor):
+    def _eager(data: torch.Tensor, position_ids: torch.Tensor, updates: torch.Tensor):
         data = data.clone()
         batch_idx = torch.arange(data.shape[0]).view(-1, 1)
         ctx_idx = torch.where(position_ids == torch.iinfo(torch.int32).max, data.shape[1] - 1, position_ids)
         data[batch_idx, ctx_idx] = updates
         return data
 
+    @staticmethod
```
quic-amitraj: These changes are needed now as we are reverting those changes?

QEfficient/transformers/models/deepseek_v3/modeling_deepseek.py
```diff
@@ -75,7 +76,7 @@ def forward(self, hidden_states):
         Returns:
             torch.Tensor: Normalized tensor.
         """
-        return CustomRMSNormFunc.apply(
+        return select_interface(CustomRMSNormFunc.apply, torch.ops.qefficient.rms_norm)(
```
quic-amitraj, line 79: Why its sill here, changes doesn't seems uniform?

QEfficient/base/modeling_qeff.py
smedhe: resolved this in the newer commits

pyproject.toml
```diff
@@ -101,4 +101,4 @@ markers = [
     "cli: marks CLI tests",
     "finetune: marks finetune tests",
     "vllm: marks vLLM tests"
-]
```
quic-amitraj, line 104: Remove this.

QEfficient/transformers/models/deepseek_v3/modeling_deepseek.py
smedhe: Needed here because we are seeing issue when dynamo tries to create subfunctions with autograd function. Deciding between the fake dynamo op and autograd func earlier, by passes the error

QEfficient/customop/ctx_scatter_gather.py
smedhe: reverted back now

QEfficient/base/modeling_qeff.py
smedhe: Moving this to constants.py file causes circular import between constants.py and ctx_scatter_gather.py file

QEfficient/base/onnx_transforms.py
smedhe: resolved this in newer comments, thanks for bringing this up. We were testing with small models, so time complexity bloating was not observable.

QEfficient/transformers/cache_utils.py
smedhe: Maintained the continuity here as well

QEfficient/transformers/cache_utils.py
smedhe: added the changes

QEfficient/transformers/cache_utils.py
smedhe: added in the newer commits

pyproject.toml
smedhe: removed this

pyproject.toml
smedhe: We are now installing torch 2.7.0 to maintain backward compatibility, user will have to install the torch 2.13. this is only in the transient phase.

smedhe:
> Can we make dynamo=True as default for CausalLM ? So, our CI tests this feature thoroughly and keep it that way?

Added separate test suite for dynamo

QEfficient/transformers/models/modeling_auto.py
smedhe: We have tested locally, we are focusing on getting Basic inference and CB working e2e for this merge.

QEfficient/transformers/models/modeling_auto.py
smedhe: added this change

QEfficient/transformers/models/modeling_auto.py
smedhe: Done, added in export utils.

examples/text_generation/test_subfunction_dynamo.py
smedhe: removed.

QEfficient/base/modeling_qeff.py
ochougul: move it to `customop/__init__.py`

examples/image_text_to_text/models/qwen2_5_vl/basic_inference.py
```diff
@@ -14,7 +14,7 @@
 from QEfficient import QEFFAutoModelForImageTextToText
 
 ## For AWQ model update pytorch version to 2.8.*
-model_id = "Qwen/Qwen2.5-VL-32B-Instruct"
+model_id = "Qwen/Qwen2.5-VL-3B-Instruct"
```
quic-amitraj: Please update the changes here.

scripts/Jenkinsfile
```diff
@@ -209,6 +212,84 @@ pipeline {
...
+                    export QEFF_HOME=\\$PWD/dynamo_qeff_qaic &&
+                    pytest tests/dynamo -m '(dynamo_on_qaic) and ${TEST_FILTER}' --ignore tests/dynamo/nightly --dist loadgroup -n auto --junitxml=tests/tests_log_dynamo_qaic.xml --durations=10 &&
+                    junitparser merge tests/tests_log_dynamo_qaic.xml tests/tests_log.xml &&
+                    deactivate"
+                    '''
+                }
+            }
+        }
+
+        stage('Dynamo Nightly Tests') {
```
abukhoy: In the regular CI, Nightly stage need to be removed.

QEfficient/customop/dynamo_ops.py
```diff
@@ -0,0 +1,388 @@
...
+)
+from QEfficient.customop.ctx_scatter_gather_cb import (  # noqa: E402
+    CtxGatherBlockedKVCB,
+    CtxGatherCB,
+    CtxGatherCB3D,
+    CtxScatterCB,
+    CtxScatterCB3D,
+)
+from QEfficient.customop.rms_norm import CustomRMSNorm  # noqa: E402
+
```
ochougul, line 374: move this to top of file

scripts/Jenkinsfile
```diff
@@ -209,6 +210,46 @@ pipeline {
...
+                    cd /efficient-transformers &&
+                    . preflight_qeff/bin/activate &&
+                    py=\\$(python3 -c \\"import sys; print('cp'+str(sys.version_info.major)+str(sys.version_info.minor))\\") &&
+                    arch=\\$(uname -m) &&
+                    pip install \\
+                        \\"torch@https://download.pytorch.org/whl/cpu/torch-2.13.0%2Bcpu-\\${py}-\\${py}-manylinux_2_28_\\${arch}.whl\\" \\
+                        \\"torchvision@https://download.pytorch.org/whl/cpu/torchvision-0.28.0%2Bcpu-\\${py}-\\${py}-manylinux_2_28_\\${arch}.whl\\" \\
+                        'compressed-tensors==0.17.0' \\
+                        'onnxscript==0.6.2' &&
+                    deactivate"
```
quic-amitraj: Lets move these stages post finetune and lets make a single stage for dynamo tests where we install and run tests for it as finetuning tests.

.github/workflows/unit-tests.yml
```diff
@@ -23,3 +23,38 @@ jobs:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
           HUGGINGFACE_HUB_TOKEN: ${{ secrets.HF_TOKEN }}
         run: pytest tests/unit_test/ -n auto -v
+
+  dynamo-tests:
```
quic-rishinr, line 27: how much time would this unit test would take?

QEfficient/base/modeling_qeff.py
```diff
@@ -344,21 +347,106 @@ def compile(self, *args, **kwargs) -> Path:
...
+        output_names: List[str],
+        dynamic_shapes: Optional[Dict],
+        export_kwargs: Dict,
+    ) -> None:
+        """Export via torch.export (dynamo=True) with custom op translation."""
+        # Reorder example_inputs and dynamic_shapes to match model.forward signature order,
+        # which torch.export requires for dynamic_shapes to bind correctly.
+        sig = inspect.signature(self.model.forward)
+        sig_keys = sig.parameters.keys()
+        example_inputs = {
```
quic-rishinr: Can we have a single loop over sig_keys and iterate it once? now its looping 4 times

QEfficient/base/modeling_qeff.py
```diff
@@ -344,21 +347,106 @@ def compile(self, *args, **kwargs) -> Path:
...
         example_inputs: Dict[str, torch.Tensor],
         output_names: List[str],
         dynamic_axes: Dict[str, Dict[int, str]],
-        onnx_transform_kwargs: Optional[Dict[str, any]] = None,
+        onnx_transform_kwargs: Optional[Dict[str, Any]] = None,
         export_dir: Optional[str] = None,
         offload_pt_weights: bool = True,
         prefill_only: Optional[bool] = False,
+        dynamo: bool = False,
+        dynamic_shapes: Optional[Dict[str, Dict[int, Any]]] = None,
```
quic-rishinr, line 443: nit: update the doc sting with the new variable functionality

QEfficient/transformers/models/llama/modeling_llama.py
```diff
@@ -134,8 +133,11 @@ def forward(
         key_states = self.k_proj(hidden_states, **kwargs).view(hidden_shape).transpose(1, 2)
         value_states = self.v_proj(hidden_states, **kwargs).view(hidden_shape).transpose(1, 2)
 
+        # Keep rotary cache length static inside nested compile regions to avoid
+        # Python cache-object calls (e.g., get_seq_length) that break invoke_subgraph capture.
+        past_seen_tokens = position_ids.max() + 1 if position_ids is not None else 0
```
quic-rishinr: any reason why we are moving from our previous approach of 
`kv_seq_len = past_key_value.get_seq_length(self.layer_idx, cache_position)
past_seen_tokens = past_key_values.get_seq_length(self.layer_idx) if past_key_values is not None else 0` to this?

QEfficient/transformers/models/glm4_moe/modeling_glm4_moe.py
```diff
@@ -349,7 +349,7 @@ def forward(
 
         if sin_cached is not None and cos_cached is not None:
             sin, cos = sin_cached, cos_cached
-            rotary_dim = int(self.rotary_emb.cos_cached.shape[-1])
+            rotary_dim = int(self.rotary_emb.cos_cached.detach().clone().shape[-1])
```
quic-rishinr, line 353: why do we need to detach and clone self.rotary_emb.cos_cached for fetching the rotary_dim?

QEfficient/transformers/models/mixtral_moe/modeling_mixtral.py
```diff
@@ -231,10 +287,10 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
                 experts_dtype = param.dtype
                 break
             hidden_states_for_experts = hidden_states.to(experts_dtype) if experts_dtype else hidden_states
-            if torch.onnx.is_in_onnx_export():
+            if torch.onnx.is_in_onnx_export() or torch._dynamo.is_compiling():
```
quic-rishinr, line 290: torch._dynamo.is_compiling() seems to be depricated. why do we need this check here?

QEfficient/transformers/models/granitemoe/modeling_granitemoe.py
```diff
@@ -445,7 +446,7 @@ def _update_causal_mask(
             causal_mask *= torch.arange(target_length, device=device) > cache_position.reshape(-1, 1)
             causal_mask = causal_mask[None, None, :, :].expand(input_tensor.shape[0], 1, -1, -1)
             if attention_mask is not None:
-                causal_mask = causal_mask.clone()  # copy to contiguous memory for in-place edit
+                causal_mask = causal_mask.detach().clone()  # copy to contiguous memory for in-place edit
```
quic-rishinr, line 449: why do we need to have detach here? can we remove it?

QEfficient/transformers/models/qwen3_moe/modeling_qwen3_moe.py
```diff
@@ -196,9 +193,11 @@ def __qeff_init__(self):
         self.expert_dim = getattr(self, "intermediate_size", self.gate_up_proj.shape[-2] // 2)
         gate_up_proj = self.gate_up_proj.detach()
         down_proj = self.down_proj.detach()
-        self.gate_proj = nn.Parameter(gate_up_proj[:, : self.expert_dim, :].transpose(1, 2), requires_grad=False)
-        self.up_proj = nn.Parameter(gate_up_proj[:, self.expert_dim :, :].transpose(1, 2), requires_grad=False)
-        self.down_proj_t = nn.Parameter(down_proj.transpose(1, 2), requires_grad=False)
+        self.gate_proj = nn.Parameter(
```
quic-rishinr, line 196: wouldnt this create an additional copy of all expert during forward? why do we need this change as prt of this PR?

QEfficient/transformers/cache_utils.py
```diff
@@ -55,7 +55,7 @@ def _get_invalid_idx_value(cls):
             int: Invalid index value (0 for ONNX functions, INT32_MAX otherwise)
         """
         if torch.onnx.is_in_onnx_export():
-            if cls.SUBFUNC_ENABLED:
+            if cls.SUBFUNC_ENABLED or torch._dynamo.is_compiling():
```
quic-rishinr, line 58: same as above. torch._dynamo.is_compiling() why do we need this check here?

QEfficient/transformers/cache_utils.py
```diff
@@ -1148,8 +1158,12 @@ def from_legacy_cache(
         cache = cls(
             config,
             batch_size=past_key_values[0][0].shape[0],
-            max_cache_len=past_key_values[1][0].shape[2],
-            sliding_window_len=past_key_values[0][0].shape[2],
+            max_cache_len=int(past_key_values[1][0].shape[2])
+            if torch._dynamo.is_compiling()
```
quic-rishinr, line 1162: torch._dynamo.is_compiling() is depricated either use the alternative or remove the chack.

QEfficient/utils/export_utils.py
```diff
@@ -9,15 +9,194 @@
...
+        When None, safe defaults are used for all bounds and model-type-specific
+        passes are skipped.
+
+    Returns
+    -------
+    Dict[str, Any]
+        torch.export dynamic_shapes dict with Dim objects, suitable for
+        torch.onnx.export(dynamic_shapes=...).
+    """
+    max_seq_len = getattr(model_config, "max_position_embeddings", 1024)
```
quic-rishinr, line 71: nit: can we reuse the max seq_len varible from constants.py?

QEfficient/utils/export_utils.py
```diff
@@ -9,15 +9,194 @@
...
+    model_type = getattr(model_config, "model_type", None)
+    batch_min = 1 if model_type == "gpt_oss" else 2
+
+    dim_registry: Dict[str, Any] = {}
+
+    def resolve_dim(dim_name: str):
+        if dim_name not in dim_registry:
+            if dim_name == "batch_size":
+                dim_registry[dim_name] = Dim("batch_size", min=batch_min, max=512)
```
quic-rishinr: nit: move constant to constant.py, its applicable for all magic numbers

examples/image_text_to_text/models/gemma_vision/gemma3/gemma3_example.py
```diff
@@ -32,7 +32,10 @@
 
 # For single QPC: kv_offload=False, For dual QPC: kv_offload=True
 qeff_model = QEFFAutoModelForImageTextToText.from_pretrained(
-    model_id, config=config, attn_implementation="eager", kv_offload=True
+    model_id,
+    config=config,
+    attn_implementation="eager",
+    kv_offload=False,
```
quic-rishinr: why its moved to false? We only support dual QPC approach for all VLMs now. Please revert

tests/transformers/models/image_text_to_text/test_continuous_batching.py
```diff
@@ -134,7 +139,7 @@ def check_image_text_to_text_pytorch_vs_kv_vs_ort_vs_ai100_CB(
         image_height = 448
         image_width = 448
         for img_url in image_urls:
-            img = requests.get(img_url, stream=True)
+            img = _session.get(img_url, stream=True)
```
quic-rishinr, line 142: lets use local path for the images.2 images are maintained inside docs/image and lets use the image for all testing

tests/transformers/subfunction/test_single_subfunction.py
```diff
@@ -98,7 +98,7 @@ def test_tinyllama_exports_single_decoder_subfunction(tmp_path):
 
     try:
         qeff_model = QEFFAutoModelForCausalLM(
-            AutoModelForCausalLM.from_pretrained(model_id, **model_kwargs),
+            AutoModelForCausalLM.from_pretrained(model_id, **model_kwargs, dtype=torch.float32),
```
quic-rishinr: by default dytpe is set to float32 we can remove it

examples/dynamo/causal_lm/requirements.txt
```diff
@@ -0,0 +1,20 @@
+# PyTorch >= 2.13 is required for dynamo export (dynamo=True).
```
quic-rishinr, line 1: move it under example folder for now. Add a readme file and the requirement.txt file on how to enable and run it. We should not keep it under root folder

quic-rishinr (changes requested):

.github/workflows/unit-tests.yml
smedhe: these take 7-8 mins to run

examples/dynamo/causal_lm/requirements.txt
quic-amitraj: cc: @ochougul

QEfficient/transformers/models/glm4_moe/modeling_glm4_moe.py
smedhe: To solve the dynamo related error, there was a mismatch in arg count in the subfunction boundary, which was solved

QEfficient/transformers/models/llama/modeling_llama.py
smedhe: reverted to the original code, check now dynamo is working with the old code and cache utils, select interface changes.

QEfficient/transformers/models/mixtral_moe/modeling_mixtral.py
smedhe: torch.onnx.is_in_onnx_export() is not set during dynamo-path export (dynamo=True), so
  torch._dynamo.is_compiling() is required to gate this path; it's intentional and will be hardened
  in a follow-up once the dynamo export stabilizes.

QEfficient/transformers/cache_utils.py
smedhe: torch.onnx.is_in_onnx_export() is not set during dynamo-path export (dynamo=True), so
  torch._dynamo.is_compiling() is required to gate this path

QEfficient/transformers/cache_utils.py
smedhe: this check is needed for dynamo export. Currently the api is working, once we have dynamo stable we'll switch

QEfficient/base/modeling_qeff.py
smedhe: done added

QEfficient/base/modeling_qeff.py
smedhe: added doc string

examples/image_text_to_text/models/gemma_vision/gemma3/gemma3_example.py
smedhe: reverted

quic-amitraj (approved):

quic-amitraj:
### `gpt_oss` disabled in `test_dynamo_cb_generate`

Temporarily skipped `gpt_oss` in `tests/dynamo/test_on_qaic.py::test_dynamo_cb_generate` to unblock CI.

**Failure:** The QAIC compiler rejects `GptOssForCausalLM.onnx` when `-sub-functions` is active:

PR #1188, opened 2026-07-17
Add CCL support to Gemma4, Qwen3.5 (moe), and qwn3_vl models

vjanfaza:
Enable the compute-context-length (CCL) feature for the Gemma4 and Qwen3.5 image-text-to-text model families and fix CCL specialization gating so it also works for disaggregated (separate prefill/decode) serving.

Gemma4 (modeling_gemma4.py):
- Thread comp_ctx_lengths through the text model, decoder layer, and attention so the exported ONNX actually consumes it: slice the attention mask to the CCL width and pass "CCL" into cache_kwargs (mirrors Gemma3).
- Fix the dummy comp_ctx_lengths input dtype (int8 -> int64) so the traced ONNX input matches the int64 runtime buffer.
- Split the "prefill and decode" specialization gate into an OR so CCL is honored when only one of the prefill/decode lists is provided.

Qwen3.5 (modeling_qwen3_5.py, modeling_qwen3_5_moe.py):
- Fix the dummy comp_ctx_lengths input dtype (int8 -> int64).
- Add the missing comp_ctx_lengths pass-through on the MoE full-model forwards (QEffQwen3_5MoeModel, QEffQwen3_5MoeForConditionalGeneration) to match the non-MoE variant.
- Allow decode-only CCL specialization (gate -> OR, guard prefill loop).

Qwen3.5-VL (modeling_qwen3_vl.py, modeling_qwen3_vl_moe.py):
- Allow decode-only CCL specialization and guard both prefill/decode loops with `or []` to avoid a TypeError when one list is None (disaggregated serving).

Examples (gemma4_example.py, gemma4_utils.py):
- Carry comp_ctx_lengths_prefill/decode through build_compile_kwargs and document CCL activation in the example.

Important points to consider:
- In this support, for gemma4 and qwen3_5 models the use_onnx_subfunctions should be False to have CCL feature actually work for these models otherwise it will be compiled but CCL won't be effective. Related ongoing work is in progress to solve the source of this issue.
- I've already raised the following Jira ticket for this issue. Please track the changes there:
https://jira-dc.qualcomm.com/jira/browse/QRANIUMSW-63142

### DIFF
--- a/QEfficient/transformers/models/gemma4/modeling_gemma4.py
+++ b/QEfficient/transformers/models/gemma4/modeling_gemma4.py
@@ -415,6 +415,7 @@ class QEffGemma4TextAttention(Gemma4TextAttention):
         position_ids: Optional[torch.LongTensor] = None,
         mm_token_type_ids: Optional[torch.Tensor] = None,
         batch_index: Optional[torch.LongTensor] = None,
+        comp_ctx_lengths: Optional[torch.Tensor] = None,
         **kwargs,
     ) -> tuple[torch.Tensor, torch.Tensor | None]:
         input_shape = hidden_states.shape[:-1]
@@ -451,6 +452,9 @@ class QEffGemma4TextAttention(Gemma4TextAttention):
             token_key_states, token_value_states = key_states, value_states
 
         if past_key_values is not None:
+            if comp_ctx_lengths is not None and attention_mask is not None:
+                attention_mask = attention_mask[:, :, :, : comp_ctx_lengths.shape[-1]]
+                cache_kwargs["CCL"] = attention_mask.shape[-1]
             if self.is_kv_shared_layer:
                 if token_key_states is not None and token_value_states is not None:
                     key_states, value_states = past_key_values.update(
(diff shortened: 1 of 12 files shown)


quic-hemagnih:
Can you please update the example script for one of the QWEN3 or QWEN3.5 model for CCL. Also do post the results in the comments after execution via QEfficient.

quic-hemagnih (changes requested):
Can you post the execution results via QEFF, and update one of the QWEN3/3.5 example script

vjanfaza:
> Can you post the execution results via QEFF, and update one of the QWEN3/3.5 example script

I added CCL related changes but they are commented by default

vjanfaza:
> Can you post the execution results via QEFF, and update one of the QWEN3/3.5 example script

I have created the following confluence page:
https://qualcomm-confluence.atlassian.net/wiki/spaces/QRAN/pages/3378664571/Benchmarking-Gemma4-Qwen3.5+moe+-Qwen3-vl+moe

quic-rishinr (approved):
LGTM!! thanks

PR #1190, opened 2026-07-17
Introducing OptimizedMoeTransform

ochougul:
This PR standardizes QEff MoE model wrappers around a single explicit
  QEffMoEBlockMixin contract.

  Key changes:

  - Enforces every QEffMoEBlockMixin subclass to explicitly declare
    supported_moe_flavours.

  - Collapses DeepSeek V3 MoE into a single QEffDeepseekV3MoE block and
    removes the separate prefill-only class.

  - Adds explicit MoE flavour declarations for DeepSeek V3 and Grok1.
  - Introduces a single Gemma4 text MoE block backed by QEffMoEBlockMixin.
  - Updates external MoE mapper wiring to source support metadata from the
    QEff block classes.

  - Removes Gemma4’s prefill-only expert mapping in favor of the shared MoE
    flavour path.

  ## Impact For New MoE Models

  When adding a new MoE model, the model should expose one QEff MoE block that
  inherits QEffMoEBlockMixin and explicitly declares:

  supported_moe_flavours = (...)

  The block should implement the common variation points:

  - route(...)
  - build_moe_weights(...) / get_moe_weights(...)
  - moe_profile
  - optional shared expert handling

  This keeps flavour selection, export config, and future optimizations
  centralized instead of spreading model-specific prefill/decode paths across
  multiple classes.

  ## Future Scope

  The longer-term direction is to support all MoE flavours across all MoE
  models where the model architecture allows it, especially:

  - decode_bmm
  - simple_loop
  - expert_parallel

  This PR makes that easier by making supported flavour declarations explicit
  and discoverable.

  ## Tests

  Added/updated coverage for:

  - explicit supported_moe_flavours enforcement
  - DeepSeek V3 single-block/external mapper behavior
  - Grok1 explicit flavour metadata
  - Gemma4 MoE block parity for decode_bmm and simple_loop
  - DeepSeek legacy prefill fallback behavior

  Validation run:

  - pytest tests/transformers/models/test_moe_prefill_blocked.py -k 'moe_block
    or deepseek or grok or gemma4'

  - pytest tests/unit_test/transforms/test_transform_accuracy.py -k
    'OptimizedMoE or deepseek or external_mapper'

  - pytest tests/unit_test/models/test_model_quickcheck.py -k 'moe_prefill'


QEfficient/transformers/models/glm4_moe/modeling_glm4_moe.py
```diff
@@ -678,103 +653,8 @@ def orig_moe(self, hidden_states: torch.Tensor, topk_indices: torch.Tensor, topk
...
-            topk_indices, topk_weights = router_output
-        else:
-            topk_indices, topk_weights = self.route_tokens_to_experts(router_output)
-        hidden_states = hidden_states.view(-1, hidden_states.shape[-1])
-
-        hidden_states = self._forward_expert_blocked(hidden_states, topk_indices, topk_weights).view(*orig_shape)
-
-        hidden_states = hidden_states + self.shared_experts(residuals)
-        return hidden_states
+QEffPrefillChunkedGlm4MoeMoE = QEffGlm4MoeMoE
```
ochougul, line 659: not needed.

QEfficient/transformers/models/glm4_moe/modeling_glm4_moe.py
```diff
@@ -619,34 +560,68 @@ def forward(self, hidden_states):
...
+            self.moe_weights = build_canonical_expert_weights(
+                gate_up=self.experts.gate_up_proj,
+                down=self.experts.down_proj,
+                fused=True,
+                fused_split_dim=1,
+                transpose_gate_up=True,
+                transpose_down=True,
+            )
+            delete_module_attrs(self.experts, "gate_up_proj", "down_proj")
+        else:
```
ochougul, line 602: this if-else is not needed, as we always know what weights model is going to have, let's fix it instead of handling un-necessary cases.

QEfficient/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py
```diff
@@ -438,8 +442,8 @@ def qeff_torch_causal_conv1d_update(
     pos_ids = position_ids[0]
     zeros = torch.zeros((pos_ids.shape[0], state_len), dtype=pos_ids.dtype, device=pos_ids.device)
     out = torch.cat([zeros, pos_ids], dim=1)
-    order = torch.argsort(out, dim=1)  # sorted positions per batch row
-    last_positions = order[:, -state_len:]  # (B, state_len)
+    order = torch.argsort(out, dim=1)
+    last_positions = order[:, -state_len:]
```
ochougul, line 453: revert

QEfficient/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py
```diff
@@ -560,10 +564,10 @@ def __qeff_init__(self):
         # Precompute all constant masks — no triu/tril with diagonal args at runtime
         # mask_causal: upper triangular including diagonal (diagonal=0)
         # = triu(ones, diagonal=0)
-        mask_causal = torch.zeros(chunk_size, chunk_size, dtype=torch.bool)
+        mask_causal = torch.ones(chunk_size, chunk_size, dtype=torch.bool)
         for i in range(chunk_size):
-            for j in range(i, chunk_size):
-                mask_causal[i, j] = True
+            for j in range(i + 1):
+                mask_causal[i, j] = False
```
ochougul, line 580: revert

QEfficient/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py
```diff
@@ -656,8 +660,7 @@ def torch_chunk_gated_delta_rule_qeff(
             x.reshape(x.shape[0], x.shape[1], -1, chunk_size, x.shape[-1]) for x in (query, key, value, k_beta, v_beta)
         ]
         g = g.reshape(g.shape[0], g.shape[1], -1, chunk_size)
-        # mask = torch.triu(torch.ones(chunk_size, chunk_size, dtype=torch.bool, device=query.device), diagonal=0)
-        mask = mask_causal
+        mask = mask_causal.to(device=query.device)
```
ochougul, line 771: revert

ochougul:
TODO: allow cores_per_expert to be fraction
TODO: what happens when we do prefill -> decode compile back to back on same object?

QEfficient/base/modeling_qeff.py
```diff
@@ -540,24 +550,21 @@ def get_onnx_path(
             kwargs["_layerwise_cache_probe"] = True
         if kv_cache_prefix:
             kwargs["kv_cache_prefix"] = kv_cache_prefix
-
+        num_devices = int(compiler_options.pop("num_devices", 1))
```
quic-rishinr, line 694: nit: any specific reason why we are popping the num_devices? similar to how we take the num_cores we can use get right?

QEfficient/base/modeling_qeff.py
```diff
@@ -571,7 +578,12 @@ def get_onnx_path(
             ctx_len=ctx_len,
             seq_len=seq_len,
             bs=bs,
+            num_devices=num_devices,
             qaic_config=qaic_config,
+            prefill_only=prefill_only,
+            enable_chunking=enable_chunking,
+            num_cores=kwargs.get("num_cores", compiler_options.get("aic_num_cores", constants.DEFAULT_AIC_NUM_CORES)),
```
quic-rishinr, line 726: variables are not added to transfrom method, in the transform method is still fetching the num_cores from compiler args.

QEfficient/base/modeling_qeff.py
```diff
@@ -848,6 +862,21 @@ def transform(
             self.hash_params["qaic_config"] = qaic_config
         self.hash_params["num_replicate_kv_heads"] = effective_num_replicate_kv_heads
 
+        num_cores = compiler_options.get("num_cores", compiler_options.get("aic_num_cores"))
```
quic-rishinr, line 1007: num_cores, prefill_seq, prefill_only length is passed as arguments. Need not extrat it again.

quic-rishinr (changes requested):

abhishek-singh591:
## Dynamo MoE Export Fix Summary

  ### Problem 1: `supported_moe_flavours` Error During Dynamo Export

  MoE Dynamo export was failing for models such as:

  - `glm4_moe`
  - `gpt_oss`
  - `mixtral`
  - `qwen3_moe`

  The failure looked like:

  ```text
  TypeError: QEffGptOssMLP must explicitly define supported_moe_flavours
  TypeError: QEffGlm4MoeMoE must explicitly define supported_moe_flavours
  TypeError: QEffMixtralSparseMoeBlock must explicitly define supported_moe_flavours
  TypeError: QEffQwen3MoeSparseMoeBlock must explicitly define supported_moe_flavours

  The real QEff MoE classes already define:

  supported_moe_flavours = (...)

  The issue was caused by PyTorch Dynamo / FX internals. During torch.export, PyTorch creates temporary proxy subclasses while tracing the model. These temporary classes do not directly define supported_moe_flavours, so the strict QEffMoEBlockMixin.__init_subclass__ check rejected
  them.

  ### Fix 1

  Updated:

  QEfficient/transformers/moe/block.py

  Now:

  - Real QEff MoE classes still must explicitly define supported_moe_flavours.
  - PyTorch FX/Dynamo temporary proxy classes are allowed to inherit the already-validated value from the real QEff class.

  This keeps the explicit MoE contract intact while making it compatible with Dynamo tracing.

  ———

  ### Problem 2: Dynamo Aliasing Error For Fused MoE Weights

  granitemoe failed with a Dynamo aliasing error during subfunction export.

  Some MoE models store gate and up weights inside a single fused tensor. QEff splits that fused tensor into separate gate and up tensors.

  Previously, those split tensors could be views into the same original storage. Dynamo does not allow this kind of input aliasing during higher-order/subfunction export.

  The failure looked like:

  UncapturedHigherOrderOpError: Encountered aliasing during higher order op tracing
  Input-to-input aliasing detected at ... moe_weights ... gate ... up ...

  ### Fix 2

  Updated fused MoE weight canonicalization call sites to use:

  clone=True

  This makes gate, up, and down independent tensors instead of shared-storage views.

  That removes the Dynamo aliasing issue.

  ———

  ### Problem 3: Duplicate num_cores Keyword

  There was also a syntax error in modeling_auto.py:

  SyntaxError: keyword argument repeated: num_cores

  The same export call passed num_cores twice.

  ### Fix 3

  Updated:

  QEfficient/transformers/models/modeling_auto.py

  Removed the duplicate num_cores argument and kept the explicit num_cores=num_cores value.

 ———

abhishek-singh591:
Gemma4 test: The test simulates a valid mapper behavior: QEff classes can be attached to existing HF
    modules while preserving weights. The code assumed __qeff_init__() always ran, but a QEff class method
    should be robust when its required derived parameters are missing. Fixing lazy init keeps the wrapper
    safe without weakening parity coverage.
  - GPT-OSS prefill-only test: The test expected a clean ValueError for invalid API usage. The code instead
    crashed with an accidental TypeError from None % int. Changing the test would accept a worse user-facing
    error and hide broken validation.
  - GPT-OSS disagg tests: The tests intentionally compile a tiny GPT-OSS dummy model with num_cores=16,
    which is a normal pattern in this repo: small configs should exercise the same export/compile path as
    real models. The code incorrectly required 4 experts * 1 core_per_expert to be divisible by 16 requested
    cores. For over-provisioned cores, the layout should use only the available expert lanes, not fail.

vbaddi:
fix the failing unit tests too. @abhishek-singh591

abhishek-singh591:
Commit note:

  - Adds a DeepSeek-V3 MoE path for quantized expert projections.
  - Detects quantized experts using qweight, qzeros, scales, g_idx, bits, and group_size.
  - Stacks packed expert weights, scales, and zeros during weight transform.
  - Runs quantized experts through unpack, dequantize, and batched matmul.
  - Keeps the existing generic MoE forward path for non-quantized experts.

QEfficient/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -36,228 +37,59 @@
...
-
-            # Down projection
-            down_out = (intermediate @ W_d) + b_d  # [T, H]
-
-            # Apply routing weights and accumulate
-            expert_out += down_out * routing_weight
-
-        # original shape [B, S, H]
-        return expert_out.view(B, S, H), router_logits
+        return super().forward(hidden)
```
abhishek-singh591: Can we move this too in MOE section.

QEfficient/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -461,57 +329,18 @@ def forward_weights_as_activation(self, hidden_states):
...
-
-        # Apply routing weights AFTER expert computation
-        experts_out = experts_out * router_top_value.unsqueeze(-1)
-        experts_out_sum = torch.einsum("bnd->bd", experts_out)
-        return experts_out_sum, router_logits
+        if os.environ.get("NUM_FFN_BLOCKS", None) is not None:
+            return self.blocked_ffn_forward(hidden_states)
+        return QEffMoEBlockMixin.forward(self, hidden_states)
 
     def optimized_moe_forward(self, hidden_states: torch.Tensor):
```
abhishek-singh591: This as well.

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -1453,14 +1453,18 @@ def _resolve_expert_parallel_layout(
...
+    total_avl_cores = num_devices * num_cores
+    if required_expert_cores < total_avl_cores and total_avl_cores % required_expert_cores == 0:
+        num_pipeline_stages = 1
+    elif required_expert_cores % total_avl_cores == 0:
+        num_pipeline_stages = required_expert_cores // total_avl_cores
+    else:
         raise ValueError(
             "num_experts * cores_per_expert "
-            f"({num_experts * cores_per_expert}) must be divisible by total_avl_cores ({total_avl_cores})"
+            f"({required_expert_cores}) must be divisible by total_avl_cores ({total_avl_cores})"
```
abhishek-singh591: total_avl_cores needs to be verbose.

tests/transformers/models/test_moe_prefill_blocked.py
```diff
@@ -928,27 +928,19 @@ def test_gemma4_text_moe_block_weight_transform_splits_fused_experts():
...
 
-    qeff_block = QEffGemma4TextMoeBlock(
-        Gemma4TextRouter(config).eval(),
-        copy.deepcopy(original_experts).eval(),
-        Gemma4RMSNorm(config.hidden_size, eps=config.rms_norm_eps).eval(),
-        Gemma4RMSNorm(config.hidden_size, eps=config.rms_norm_eps).eval(),
-    ).eval()
+    qeff_experts = copy.deepcopy(original_experts).eval()
+    qeff_experts.__class__ = QEffGemma4TextExperts
+    qeff_experts.__qeff_init__()
```
abhishek-singh591: Remove qeff_init from here and change the test.

QEfficient/transformers/models/gemma4/modeling_gemma4.py
abhishek-singh591: We should remove this.

abhishek-singh591:
Latest Fix :-
  - Dual-QPC compile now passes prefill_only and prefill_seq_len into transform(), so optimized MoE sees prefill mode at QEfficient/transformers/models/modeling_auto.py:2019.
  - Single-QPC early self.transform() is removed, so normal compile→export uses the inner get_onnx_path() transform once at _compile() time.
  - Removed QEffGemma4TextExperts and its optimized MoE mapping, since Gemma4 expert execution now uses the shared MoE block/mixin with canonical moe_weights.
  Fixes I Added During Review
  - Added missing import math because current helper uses math.gcd: QEfficient/transformers/models/modeling_auto.py:8.
  - Changed dual-QPC transform call from batch_size=batch_size to bs=batch_size, matching the transform() signature: QEfficient/transformers/models/modeling_auto.py:2022.
  - Removed unused DYNAMIC_PREFILL_SEQ_LEN_SUPPORTED_MODEL_ARCH import after deleting the inline export-time MoE seq-len logic.

QEfficient/transformers/models/gemma4/modeling_gemma4.py
abhishek-singh591: done.

tests/transformers/models/test_moe_prefill_blocked.py
abhishek-singh591: done.

QEfficient/transformers/models/pytorch_transforms.py
abhishek-singh591: Done.

QEfficient/transformers/models/gpt_oss/modeling_gpt_oss.py
abhishek-singh591: Not needed as of now.

QEfficient/transformers/models/gpt_oss/modeling_gpt_oss.py
abhishek-singh591: Not needed as of now.

QEfficient/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py
abhishek-singh591: done.

QEfficient/transformers/models/qwen3_5_moe/modeling_qwen3_5_moe.py
abhishek-singh591: done.

ochougul:
CI-Ready

ochougul:
CI-Ready

QEfficient/utils/export_utils.py
```diff
@@ -246,6 +246,7 @@ def wrapper(self, *args, **kwargs):
 
         # Cache probe flag (used for layerwise inspection runs)
         cache_probe = kwargs.pop("_layerwise_cache_probe", False)
+        kwargs.pop("_skip_pre_export_pytorch_transforms", False)
```
quic-mamta: _skip_pre_export_pytorch_transforms is popped but never used, and the new _apply_pre_export_pytorch_transforms() is not used anywhere. This means direct .export() does not actually reconfigure MoE, only .compile() does.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -4139,7 +4176,6 @@ def _legacyify_cache(obj):
                 use_onnx_subfunctions=kwargs.get("use_onnx_subfunctions", False),
                 dynamo=dynamo,
                 offload_pt_weights=kwargs.get("offload_pt_weights", True),
-                prefill_only=prefill_only,
```
quic-mamta, line 4145: Non-layerwise causal lm _export() path still doesn’t run OptimizedMoETransform for
direct export(prefill_only=True, qaic_config)
 
The same issue in the VLM non layerwise export paths, Lines 1328 and 2794, they parse MoE related kwargs but do not pass them into _export(). So direct prefill exports for VLM MoE models can export the decode_bmm path instead of prefill/expert-parallel path.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1929,6 +1919,7 @@ def compile(
         """
         if skip_lang and skip_vision:
             raise ValueError("Expected at least one of 'skip_lang' or 'skip_vision' to be False")
+        reject_legacy_moe_prefill_packed_chunk_size(compiler_options)
```
quic-mamta, line 1926: moe_prefill_packed_chunk_size is still an explicit compile parameter, so reject_legacy_moe_prefill_packed_chunk_size(compiler_options) cannot catch it. Since the PR says this option is no longer supported, remove it from the signature or explicitly reject non-default usage.

QEfficient/base/modeling_qeff.py
```diff
@@ -350,6 +361,38 @@ def compile(self, *args, **kwargs) -> Path:
...
+        """Apply mode-dependent PyTorch transforms before export hashing and ONNX tracing."""
+        from QEfficient.transformers.models.pytorch_transforms import OptimizedMoETransform
+
+        if num_cores is None:
+            num_cores = constants.DEFAULT_AIC_NUM_CORES
+        if moe_prefill_packed_chunk_size is None:
+            moe_prefill_packed_chunk_size = constants.MOE_PREFILL_PACKED_CHUNK_SIZE
+        qaic_config = qaic_config if qaic_config is not None else getattr(self.model, "qaic_config", None)
+
+        self.model, transformed = OptimizedMoETransform.apply(
```
quic-mamta: This _apply_pre_export_pytorch_transforms() still passes enable_chunking and moe_prefill_packed_chunk_size into OptimizedMoETransform.apply(), but that method does not use those kwargs.

QEfficient/base/modeling_qeff.py
```diff
@@ -350,6 +361,38 @@ def compile(self, *args, **kwargs) -> Path:
...
+            self.model,
+            prefill_only=bool(prefill_only),
+            enable_chunking=bool(enable_chunking),
+            num_cores=num_cores,
+            moe_prefill_packed_chunk_size=moe_prefill_packed_chunk_size,
+            qaic_config=qaic_config,
```
quic-mamta: here why are we passing moe_prefill_packed_chunk_size, qaic_config should have expert_parallel_chunk_size ?

QEfficient/cloud/infer.py
```diff
@@ -19,6 +20,20 @@
...
 from QEfficient.utils.logging_utils import logger
 
+_DEPRECATED_CLOUD_API_WARNING = (
+    "QEfficient.cloud.infer is deprecated. These APIs will no longer be usable and will be removed after the "
+    "1.23.0 release."
+)
+_DEPRECATION_WARNING_EMITTED = False
+
+
+def _warn_deprecated_cloud_api() -> None:
```
quic-mamta: this function can be defined once and reused in infer/compile/export/finetune.

QEfficient/transformers/models/deepseek_v3/modeling_deepseek.py
```diff
@@ -782,174 +788,173 @@ def forward(
...
 
-        # Get common parameters from first expert
-        first_expert = self.experts[0]
-        self._qeff_quantized_experts = hasattr(first_expert.gate_proj, "bits")
-        if not self._qeff_quantized_experts:
+    def _transform_quantized_expert_weights(self) -> None:
+        if hasattr(self, "all_gate_qweight"):
             return
 
+        first_expert = self.experts[0]
```
quic-mamta: _deepseek_first_unfused_expert can be used here.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -2028,10 +2019,12 @@ def compile(
         self.transform(
             ctx_len=ctx_len,
             seq_len=prefill_seq_len,
-            batch_size=batch_size,
+            bs=batch_size,
```
quic-mamta, line 2025: why is it changed?

QEfficient/transformers/moe/block.py
```diff
@@ -0,0 +1,168 @@
...
+    def __qeff_init__(self) -> None:
+        if "weights_transformed" not in self.__dict__:
+            self.weights_transformed = False
+
+    @property
+    def expert_blocking_num_packed_chunks(self) -> int:
+        return self.expert_parallel_num_packed_chunks
+
+    @expert_blocking_num_packed_chunks.setter
+    def expert_blocking_num_packed_chunks(self, value: int) -> None:
```
quic-mamta: should we rename this to expert_parallel_x instead of expert_blocking_x? same for line 89

QEfficient/base/modeling_qeff.py
ochougul: yeah its okay to pass them anyway.

QEfficient/transformers/models/modeling_auto.py
ochougul: I think this will break the vllm infra, we need to to send a mail to them before deprecating this. So for now we should raise a warning but still accept it. Good catch

QEfficient/transformers/models/modeling_auto.py
ochougul: we don't support the OptimizedMoeTransform via export path, its only supported via compile. and vllm uses compile always so this is okay.

QEfficient/transformers/models/deepseek_v3/modeling_deepseek.py
```diff
@@ -1122,361 +1078,16 @@ def moe_waa_unpack(self, hidden_states, topk_indices, topk_weights):
...
-
-        self.bits = first_expert.gate_proj.bits
-        self.group_size = first_expert.gate_proj.group_size
-        self.act_fn = first_expert.act_fn
-        assert first_expert.gate_proj.act_order == first_expert.up_proj.act_order == first_expert.down_proj.act_order, (
-            "act_order mismatch"
-        )
-        self.act_order = first_expert.gate_proj.act_order
+    def forward(self, hidden_states: torch.Tensor):
+        if not getattr(self, "_qeff_quantized_experts", False):
```
quic-mamta: This if condition needs to be reverted, should be if getattr(self, "_qeff_quantized_experts", False):, otherwise it will always go to moe_waa_unpack only.

quic-mamta (changes requested):

QEfficient/transformers/moe/block.py
abhishek-singh591: done.

ochougul:
CI-Ready

QEfficient/transformers/moe/quantized.py
```diff
@@ -0,0 +1,244 @@
...
+        up_proj = _dequantize_projection(up_qweight, up_scales, up_qzeros, group_size)
+        down_proj = _dequantize_projection(down_qweight, down_scales, down_qzeros, group_size)
+
+        gate_out = torch.bmm(x_chunk, gate_proj.transpose(1, 2).to(x_chunk.dtype))
+        up_out = torch.bmm(x_chunk, up_proj.transpose(1, 2).to(x_chunk.dtype))
+        down_out = torch.bmm(act_fn(gate_out) * up_out, down_proj.transpose(1, 2).to(x_chunk.dtype))
+
+        routing_chunk = CtxGatherFunc3DGeneralized.apply(routing_weight, chunk_indices)
+        previous_out = CtxGatherFunc3DGeneralized.apply(expert_out, chunk_indices)
+        chunk_valid_rows = torch.clamp(
```
quic-mamta: Have you checked if expert parallel prefill only kimi model using subfunction compiles with torch.clamp?
you should use torch.where instead of torch.clamp, 
            valid_rows_delta = valid_rows - packed_start
            chunk_valid_rows = torch.where(
                valid_rows_delta < 0,
                torch.zeros_like(valid_rows_delta),
                torch.where(
                    valid_rows_delta > packed_chunk_size,
                    torch.full_like(valid_rows_delta, packed_chunk_size),
                    valid_rows_delta,
                ),
            )
https://github.com/quic/efficient-transformers/blob/main/QEfficient/transformers/models/deepseek_v3/modeling_deepseek.py#L1323

quic-mamta (changes requested):

QEfficient/transformers/moe/quantized.py
abhishek-singh591: Do we have a test for this in the CI?

QEfficient/transformers/moe/quantized.py
```diff
@@ -0,0 +1,244 @@
...
+    @property
+    def num_experts(self) -> int:
+        return self.gate_qweight.shape[0]
+
+    @property
+    def hidden_size(self) -> int:
+        return self.gate_qweight.shape[-1] * 2
+
+
+def _build_matched_idx_from_cumsum(token_to_expert: torch.Tensor) -> torch.Tensor:
```
quic-mamta, line 62: this can be reused from moe/flavours.py

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -1478,3 +1409,384 @@ def apply(cls, model: nn.Module, attn_blocking_config) -> Tuple[nn.Module, bool]
...
+            chunk_size_key in moe_config and moe_config.get(chunk_size_key) is not None
+        )
+        expert_parallel_chunk_size = moe_config.get(chunk_size_key, MOE_PREFILL_PACKED_CHUNK_SIZE)
+        if expert_parallel_chunk_size is None:
+            expert_parallel_chunk_size = MOE_PREFILL_PACKED_CHUNK_SIZE
+        expert_parallel_chunk_size = int(expert_parallel_chunk_size)
+        if expert_parallel_chunk_size <= 0:
+            raise ValueError("moe expert_parallel_chunk_size must be greater than zero")
+        compile_seq_len = prefill_seq_len or ONNX_EXPORT_EXAMPLE_SEQ_LEN
+        num_packed_chunks = max(1, -(-compile_seq_len // expert_parallel_chunk_size))
```
quic-mamta, line 1624: Can we put a check here that compile_seq_len must be divisble by expert_parallel_chunk_size and error out early?

QEfficient/transformers/moe/quantized.py
ochougul: Yes makes sense

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3903,35 +3888,55 @@ def export(
...
+        #################
+        # Handle OptimizedMoeTransform, BlockingTransform introduced changes
+        ################
+        # TODO: We should change the ctx_len i.e. the past_key_values shape based on num_kv_blocks which is not yet handled.
+        expert_parallel_is_triggered = self.hash_params.get("moe_prefill_flavour") == MoEFlavour.EXPERT_PARALLEL
+        blocking_enabled = self.hash_params.get("blocking_kwargs", None) is not None
+        for_loop_number_forced_by_blocking = (
+            self.hash_params.get("blocking_kwargs", None).num_q_blocks if blocking_enabled else None
+        )
+        for_loop_number_forced_by_expert_parallel = self.hash_params.get("moe_prefill_expert_parallel_chunk_size", None)
```
quic-mamta: moe_prefill_num_packed_chunks should be used here.

ochougul: yes that is fixed. Thanks

QEfficient/transformers/models/pytorch_transforms.py
ochougul: No, if lets say user is compiling for seq_len=32, but default block size is 256. we should go to 1. I think right thing is if its not divisible just tell user that we are defaulting to 1. I will add the warning.

ochougul:
CI-Ready

ochougul:
CI-Ready

QEfficient/transformers/moe/qflavours.py
```diff
@@ -0,0 +1,283 @@
...
+        else:
+            packed_stop = packed_start + packed_chunk_size
+
+        chunk_matched_idx = matched_idx[:, packed_start:packed_stop]
+
+        x_chunk = CtxGatherFunc3DGeneralized.apply(x_expanded, chunk_matched_idx)
+
+        gate_proj_unpacked = CastToUInt4Func.apply(gate_qweight)
+        gate_zeros_unpacked = CastToUInt4Func.apply(gate_qzeros)
+        gate_proj_dq = DequantizeLinearFunc.apply(gate_proj_unpacked, gate_scales, gate_zeros_unpacked, group_size)
```
quic-mamta: _dequantize_projection can be used here similar to moe_quantized_decode_bmm.

PR #982, opened 2026-05-13
Adding PagedAttention support for CausalLM models

vaibverm:
This PR adds the PagedAttention (https://arxiv.org/pdf/2309.06180) support for all CausalLM models in QEfficient. 
The major change is that KV cache is not treated as a contiguous memory under this implementation but rather a collection of blocks which can reside in a non-contiguous fashion inside the memory. This forces cache scatter and gather operations to happen per KV block.

### Summary of changes compared to BlockedKV:
1) The **cache shape** changes from [BS, num_kv_heads, CL, dh] to [total_num_kv_blocks,  num_kv_heads, kv_block_size, dh].
2) num_kv_blocks = -(-ctx_len // kv_block_size) = physical blocks required for 1 batch element in K cache.
3) Total_num_kv_blocks = BS (kv_batch_size) * num_kv_blocks = total physical blocks available for K cache.
4) 2 new inputs **block_table** [BS, num_kv_blocks] and **slot_id** [BS] are passed as inputs to the ONNX.
    4) a) block_id is each entry in the block_table and points to the physical K/V block that needs to be read/written corresponding to (position_id // kv_block_size)th entry in block_table. ‘-1’ signifies invalid/unallocated block.
    4) b) slot_id tells how many entries are already filled in currently active block => read up to / write after (slot_id – 1)
5) _Limitation_ - Cache writes to only 1 block at a time per batch element => CPL = kv_block_size. Hence, cache writes should not cross the block boundary.
6) vLLM provides KV Cache Manager implementation which maintains the KV cache block_table with logical to physical block mapping and slot_id for location mapping within the active block.

PR #1209, opened 2026-07-23
Adding PagedAttention support for CausalLM models

vaibverm:
This PR is a clone of https://github.com/quic/efficient-transformers/pull/982. This PR adds the PagedAttention (https://arxiv.org/pdf/2309.06180) support for all CausalLM models in QEfficient.
The major change is that KV cache is not treated as a contiguous memory under this implementation but rather a collection of blocks which can reside in a non-contiguous fashion inside the memory. This forces cache scatter and gather operations to happen per KV block.

### Summary of changes compared to BlockedKV:

1. The cache shape changes from [BS, num_kv_heads, CL, dh] to [total_num_kv_blocks, num_kv_heads, kv_block_size, dh].
2. num_kv_blocks = -(-ctx_len // kv_block_size) = physical blocks required for 1 batch element in K cache.
3. Total_num_kv_blocks = BS (<kv_batch_size>) * num_kv_blocks = total physical blocks available for K cache.
4. 2 new inputs **block_table [BS, num_kv_blocks]** and **slot_id [BS]** are passed as inputs to the ONNX.
    4) a) block_id is each entry in the block_table and points to the physical K/V block that needs to be read/written corresponding to (position_id // kv_block_size)th entry in block_table. ‘-1’ signifies invalid/unallocated block.
    4) b) slot_id denotes how many entries are already filled in currently active block => read up to / write after (slot_id – 1)
5. Limitation - Cache writes to only 1 block at a time per batch element => CPL is less than or equal to kv_block_size. Hence, cache writes should not cross the block boundary.
6. vLLM provides KV Cache Manager implementation which maintains the KV cache block_table with logical to physical block mapping and slot_id for location mapping within the active block.


QEfficient/blocking/attention_blocking.py
```diff
@@ -16,26 +16,35 @@
...
     blocked_q_attention_forward,
     blocked_qkv_attention_forward,
+    blocked_qkv_paged_attention_forward,
 )
 
 
 class BlockingMode(str, Enum):
     NONE = ""
     KV = "kv"
+    KV_PAGED = "kv_paged"
```
kdulla, line 36: Would be better to not changed BlockingMode and instead add a paged_attention_flag to AttentionBlockingConfig

QEfficient/blocking/attention_blocking.py
```diff
@@ -52,15 +61,24 @@ def supports_blocked_kv(past_key_value: Optional[Cache]) -> bool:
     return past_key_value is not None and hasattr(past_key_value, "read_only_blockedKV")
 
 
+def supports_paged_attention_blocked_kv(past_key_value: Optional[Cache]) -> bool:
+    return past_key_value is not None and hasattr(past_key_value, "read_only_pagedAttention")
+
+
 _STRATEGIES: Dict[BlockingMode, Callable] = {
```
kdulla, line 68: Can create a second dictionary _STRATEGIES_PAGED that is checked if AttentionBlockingConfig has paged_attention flag True

QEfficient/blocking/attention_blocking.py
kdulla, line 58: see above

QEfficient/blocking/attention_blocking.py
```diff
@@ -147,6 +195,8 @@ def generic_blocked_attention_interface(
                 comp_ctx_lengths=comp_ctx_lengths,
                 batch_index=batch_index,
                 position_ids=position_ids,
+                block_table=block_table,
```
kdulla: Is this needed if we are not doing paged attention in this else condition?

QEfficient/blocking/attention_blocking.py
kdulla, line 197: Can change this to check different strategies dictionary based on paged_attention flag or not

QEfficient/blocking/blocking_configurator.py
```diff
@@ -42,17 +42,25 @@ def _infer_data_bytes(compile_config: Dict[str, Any]) -> int:
 
 def _normalize_attention_mode(raw_mode: str) -> str:
     mode = raw_mode.lower()
-    if "h" in mode and "q" in mode and "kv" in mode:
+    if "h" in mode and "q" in mode and "kv" in mode and "paged" in mode:
```
kdulla, line 45: See comments in attention_blocking, if we change paged_attention to a flag instead of additional BlockingModes we can avoid extra if conditions here

QEfficient/blocking/blocking_configurator.py
```diff
@@ -347,9 +354,22 @@ def build_transformer_blocking_config_for_transform(
...
-        if qaic_config.get("num_kv_blocks", False) and enable_blocking and "kv" in blocking_mode:
+        if (
+            qaic_config.get("num_kv_blocks", False)
+            and enable_blocking
+            and "kv" in blocking_mode
+            and "paged" not in blocking_mode
+        ):
             mode_from_config = "kv" + mode_from_config
             blocking_config.num_kv_blocks = _get_valid_num_blocks(qaic_config, "num_kv_blocks")
+        if (
```
kdulla, line 375: Only kv_paged attention is possible from qaic_config from this code

QEfficient/blocking/attention_blocking.py
```diff
@@ -79,10 +97,17 @@ def past_key_value_update(
     comp_ctx_lengths: Optional[torch.LongTensor] = None,
     batch_index: Optional[torch.LongTensor] = None,
     position_ids: Optional[torch.LongTensor] = None,
+    block_table: Optional[torch.LongTensor] = None,
```
kdulla: We only ever call this function when use_paged_kv_blocked is disabled, is this needed?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3864,6 +3886,22 @@ def export(
...
 
+        if self.supports_paged_attention:
+            batch, num_kv_heads, CL, dh = kv_cache_shape
+            total_num_kv_blocks = batch * num_kv_blocks
+            kv_cache_shape = [total_num_kv_blocks, num_kv_heads, kv_block_size, dh]
+            example_inputs["block_table"] = torch.arange((bs * num_kv_blocks), dtype=torch.int64).view(
+                bs, num_kv_blocks
+            )
+            example_inputs["slot_id"] = torch.zeros(bs, dtype=torch.int64)
+            dynamic_axes["block_table"] = {0: "batch_size", 1: "num_kv_blocks"}
```
kdulla, line 3970: Num kv blocks can't be a proper dynamic axis since it is fixed at export time based on the number of loops the attention forward goes through

QEfficient/transformers/cache_utils.py
```diff
@@ -246,6 +248,92 @@ def read_only_blockedKV(self, start_index, end_index, cache_kwargs):
...
+        """
+        # Gather
+        k_out, v_out = self.keys, self.values
+        if k_out is not None:
+            self._mark_initialized(k_out)
+        position_ids = cache_kwargs.get("position_ids")
+        batch, seq_len = position_ids.shape
+        num_kv_blocks, num_kv_heads, block_size, dh = k_out.shape
+        ctx_indices = torch.arange(block_size)[None, ...]
+        block_fill_len = position_ids.max(1, keepdim=True).values % block_size
```
anujgupt-github, line 283: Why don't you use slot_id here?
position_id.max % block_size should be same as slot_id, isnt it?
write** function is operating on slot_ids

QEfficient/transformers/cache_utils.py
```diff
@@ -246,6 +248,92 @@ def read_only_blockedKV(self, start_index, end_index, cache_kwargs):
...
+            num_kv_blocks, num_kv_heads, block_size, dh = self.keys.shape
+            block_index = position_ids.max(1).values // block_size  # Assuming only 1 block is written at max
+            invalid_scatter_index = torch.iinfo(torch.int32).max
+            ctx_indices = torch.arange(seq_len) + slot_id.unsqueeze(-1)
+
+            rows = torch.arange(batch)
+            block_id = block_table[rows, block_index].unsqueeze(-1)
+            ctx_indices = torch.where(block_id < 0, invalid_scatter_index, ctx_indices)
+            block_id = block_id.unsqueeze(-1)
+            self.keys = CtxScatterFuncPagedAttention.apply(self.keys, block_id, ctx_indices, key_states)
```
anujgupt-github, line 335: On a CB decode step, key_states have shape [1, num_kv_heads, 1, dh] (single request) and block_table is [full_batch_size, num_kv_blocks]. The code therefore:

Reads block_table[torch.arange(1), block_index] — only row 0 of the block table, regardless of which request slot is active.
Scatters into self.keys at block_id = block_table[0, position_ids.max // block_size] 
This means all continuously‑batched requests write into request‑0’s KV blocks.
Compare with mainline _QEffDynamicLayer.update() and _QEffDynamicLayer.write_only() in the same file. Both explicitly consume batch_index for the CB path. Paged write doesn't seem to do this

QEfficient/generation/text_generation_inference.py
```diff
@@ -195,7 +195,9 @@ def get_compilation_dims(qpc_path: str) -> Tuple[int, int, Optional[int]]:
     compilation_ctx_len = int(spec["ctx_len"])
     if compilation_fbs := spec.get("full_batch_size", None):
         compilation_fbs = int(compilation_fbs)
-    return compilation_batch_size, compilation_ctx_len, compilation_fbs
+    if compilation_num_kv_blocks := spec.get("num_kv_blocks", None):
+        compilation_num_kv_blocks = int(compilation_num_kv_blocks)
+    return compilation_batch_size, compilation_ctx_len, compilation_fbs, compilation_num_kv_blocks
```
anujgupt-github, line 200: Every caller in the codebase that unpacks 3 values will break with this change. You are updating only two call sites: cloud_ai_100_exec_kv and kv_offload_generate. 
The remaining get_compilation_dims callers on the vision‑language single‑QPC generate paths and in some tests were not updated

QEfficient/generation/text_generation_inference.py
```diff
@@ -195,7 +195,9 @@ def get_compilation_dims(qpc_path: str) -> Tuple[int, int, Optional[int]]:
     compilation_ctx_len = int(spec["ctx_len"])
     if compilation_fbs := spec.get("full_batch_size", None):
         compilation_fbs = int(compilation_fbs)
-    return compilation_batch_size, compilation_ctx_len, compilation_fbs
+    if compilation_num_kv_blocks := spec.get("num_kv_blocks", None):
+        compilation_num_kv_blocks = int(compilation_num_kv_blocks)
```
anujgupt-github, line 199: what happens without blocking?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3773,16 +3790,20 @@ def export(
...
             max_blocks = -1
             for num_blocks in self.hash_params.get("blocking_kwargs").__dict__.values():
                 if isinstance(num_blocks, int):
                     max_blocks = max(max_blocks, num_blocks)
             block_size = -(-seq_len // max_blocks)
             seq_len = block_size * max_blocks
-        fbs: int = constants.ONNX_EXPORT_EXAMPLE_FBS
+            num_kv_blocks = self.hash_params["blocking_kwargs"].num_kv_blocks
+            self.supports_paged_attention = "paged" in self.hash_params["blocking_kwargs"].mode
+            seq_len = kv_block_size = -(-seq_len // num_kv_blocks) if self.supports_paged_attention else seq_len
```
anujgupt-github, line 3867: if paged attention is false, we don't want block_size to be seq_len

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3864,6 +3886,22 @@ def export(
...
+            )
+            example_inputs["slot_id"] = torch.zeros(bs, dtype=torch.int64)
+            dynamic_axes["block_table"] = {0: "batch_size", 1: "num_kv_blocks"}
+            dynamic_axes["slot_id"] = {0: "batch_size"}
+            # Assuming 4d pkv, might have to recheck for GPTBigCode with 3d pkv
+            pkv_dynamic_axes = {
+                0: "total_num_kv_blocks",
+                2: "kv_block_size",
+            }
+
```
anujgupt-github, line 3977: what happens when PL > block size?
How does ONNX Export work?

QEfficient/blocking/blocking_configurator.py
```diff
@@ -74,13 +80,13 @@ def _resolve_effective_blocking_mode(attention_cfg: Dict[str, Any], requested_mo
     if head_block_size > 1 and num_kv_blocks > 1:
         return "hkv"
     if head_block_size > 1:
-        return "hqkv"
+        return "h" + mode
```
anujgupt-github: What if mode is already hqkv? this will return hhkqv

QEfficient/blocking/blocking_configurator.py
```diff
@@ -74,13 +80,13 @@ def _resolve_effective_blocking_mode(attention_cfg: Dict[str, Any], requested_mo
...
     if num_q_blocks > 1 and num_kv_blocks > 1:
-        return "qkv"
+        return mode
     if num_q_blocks > 1:
-        return "q"
+        return mode
```
anujgupt-github: _normalize_attention_mode can return kv_paged.
if num_q_blocks > 1 and num_kv_blocks == 1, this still returns "kv_paged"
Is that ok?

QEfficient/blocking/blocking_configurator.py
```diff
@@ -61,8 +69,6 @@ def _normalize_attention_mode(raw_mode: str) -> str:
 
 def _resolve_effective_blocking_mode(attention_cfg: Dict[str, Any], requested_mode: str) -> str:
     mode = _normalize_attention_mode(requested_mode)
-    if mode == "":
```
anujgupt-github, line 64: Why did you remove this?
If user didn't specify blocking, why should below code get executed?

QEfficient/blocking/blocked_attention_forwards.py
```diff
@@ -208,6 +208,121 @@ def blocked_kv_attention_forward(
     return attn_output, attn_weights
```
anujgupt-github, line 210: Six new attention forward functions are copy‑paste of the corresponding non‑paged ones.
The four paged functions differ only in the KV‑block fetch step (contiguous slice vs paged gather). Core functionality is identical. Instead of doubling the code, share a helper _fetch_kv_block(cache, layer_idx, j, num_kv_blocks, cache_kwargs, paged: bool) and keep one set of forwards

QEfficient/blocking/attention_blocking.py
```diff
@@ -122,8 +149,29 @@ def generic_blocked_attention_interface(
...
     if past_key_value is not None:
-        if use_kv_blocked and sliding_window is None:
+        if use_paged_kv_blocked and sliding_window is None:
+            cache_kwargs = {
+                "batch_index": batch_index,
+                "position_ids": position_ids,
+                "block_table": block_table,
+                "slot_id": slot_id,
+            }
+            if sliding_window is not None:
```
anujgupt-github, line 162: Why separate handling? are you assuming that block size will always be > SWA size?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1231,6 +1231,11 @@ def __init__(self, model, qaic_config: Optional[dict] = None, **kwargs):
         super().__init__(model, **kwargs)
         self.model = model.get_qeff_language_decoder()
         self.model.qaic_config = qaic_config
+        # Below is to pass qaic_config downstream
+        if hasattr(self.model, "model"):
+            self.model.model.qaic_config = qaic_config
+            if hasattr(self.model.model, "model"):
+                self.model.model.model.qaic_config = qaic_config
```
anujgupt-github, line 1239: Prefer passing qaic_config explicitly through constructors, or attach it once on the top wrapper and always read via a helper that walks up.

anujgupt-github:
"Limitation - Cache writes to only 1 block at a time per batch element => CPL is less than or equal to kv_block_size. Hence, cache writes should not cross the block boundary." 

Where is this enforced in the code?

anujgupt-github (changes requested):

kdulla (changes requested):
The approach of adding new blocking modes adds a lot of repeated unnecessary code, changing that should help streamline many parts of this

QEfficient/blocking/blocking_configurator.py
```diff
@@ -61,8 +69,6 @@ def _normalize_attention_mode(raw_mode: str) -> str:
 
 def _resolve_effective_blocking_mode(attention_cfg: Dict[str, Any], requested_mode: str) -> str:
```
anujgupt-github, line 70: When paged attention is not requested, why should blocking mode resolution differ from mainline?

QEfficient/transformers/cache_utils.py
```diff
@@ -246,6 +248,92 @@ def read_only_blockedKV(self, start_index, end_index, cache_kwargs):
         v_out = torch.where(invalid_mask.unsqueeze(-1), torch.tensor(0.0, dtype=torch.float32), v_out)
         return k_out, v_out
 
+    def read_only_pagedAttention(self, block_index, updated, cache_kwargs):
```
anujgupt-github, line 258: Do we want mixed snake_case/camelCase cache APIs like read_only_pagedAttention?

QEfficient/transformers/cache_utils.py
```diff
@@ -246,6 +248,92 @@ def read_only_blockedKV(self, start_index, end_index, cache_kwargs):
...
+            ctx_indices = torch.arange(seq_len) + slot_id.unsqueeze(-1)
+
+            rows = torch.arange(batch)
+            block_id = block_table[rows, block_index].unsqueeze(-1)
+            ctx_indices = torch.where(block_id < 0, invalid_scatter_index, ctx_indices)
+            block_id = block_id.unsqueeze(-1)
+            self.keys = CtxScatterFuncPagedAttention.apply(self.keys, block_id, ctx_indices, key_states)
+            self.values = CtxScatterFuncPagedAttention.apply(self.values, block_id, ctx_indices, value_states)
+
+    def get_seq_lengthPagedAttention(self, cache_position=None) -> int:
```
anujgupt-github, line 338: Should this return allocated paged-cache capacity or the active sequence length?

QEfficient/customop/ctx_scatter_gather.py
```diff
@@ -56,6 +56,53 @@ def symbolic(g: torch.Graph, data: torch.Value, position_ids: torch.Value, updat
...
+    return ops.ScatterND(data, indices, updates)
+
+
+class CtxScatterFuncPagedAttention(torch.autograd.Function):
+    """
+    Function to scatter the current key values into KV-cache.
+    """
+
+    @staticmethod
+    def forward(data: torch.Tensor, block_index: torch.Tensor, position_ids: torch.Tensor, updates: torch.Tensor):
```
anujgupt-github, line 89: Is this in-place scatter safe if the same cache tensor is reused across eager parity checks?

QEfficient/utils/generate_inputs.py
```diff
@@ -131,8 +158,21 @@ def prepare_pytorch_inputs(self):
             inputs["batch_index"] = torch.arange(self.full_batch_size).view(-1, 1)
 
         past_key_values = []
+        num_kv_blocks = self._get_num_kv_blocks()
+        kv_block_size = self._get_kv_block_size()
+
+        if self._is_paged_attention and num_kv_blocks and kv_block_size:
```
anujgupt-github, line 164: Is this intended to check the _is_paged_attention method object instead of calling it?

QEfficient/transformers/models/llama/modeling_llama.py
```diff
@@ -112,6 +112,8 @@ def forward(
         hidden_states: torch.Tensor,
```
anujgupt-github, line 111: Do we want every model wrapper signature to carry block_table and slot_id when most paths won't use paged attention?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3773,16 +3790,20 @@ def export(
         bs: int = constants.ONNX_EXPORT_EXAMPLE_BATCH_SIZE
         seq_len: int = constants.ONNX_EXPORT_EXAMPLE_SEQ_LEN
 
-        # increase seq_len if using a larger number of blocks
+        self.supports_paged_attention = False
```
anujgupt-github, line 3856: Should export input construction mutate self.supports_paged_attention as model state?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -3773,16 +3790,20 @@ def export(
...
+        # increase seq_len if using a larger number of blocks and set PagedAttention params if required
         if self.hash_params.get("blocking_kwargs", None):
             max_blocks = -1
             for num_blocks in self.hash_params.get("blocking_kwargs").__dict__.values():
```
anujgupt-github, line 3860: Do we want max_blocks derived from every int in blocking_kwargs.__dict__, including unrelated fields?

QEfficient/blocking/attention_blocking.py
```diff
@@ -52,15 +61,24 @@ def supports_blocked_kv(past_key_value: Optional[Cache]) -> bool:
     return past_key_value is not None and hasattr(past_key_value, "read_only_blockedKV")
 
 
+def supports_paged_attention_blocked_kv(past_key_value: Optional[Cache]) -> bool:
```
anujgupt-github, line 64: Should paged behavior be selected by hasattr on the cache object rather than an explicit config signal?

QEfficient/generation/text_generation_inference.py
```diff
@@ -482,6 +486,8 @@ def __init__(
         self.full_batch_size = (
             full_batch_size if full_batch_size else self._fetch_full_batch_size()
         )  # Check and fetch full batch size if CB is enabled
+        self.num_kv_blocks = num_kv_blocks if num_kv_blocks else None
```
quic-rishinr, line 489: nit: since the default value is None we need not require the if condition here

vaibverm:
> "Limitation - Cache writes to only 1 block at a time per batch element => CPL is less than or equal to kv_block_size. Hence, cache writes should not cross the block boundary."
> 
> Where is this enforced in the code?

This is not enforced in code right now since this is not a hard limitation. This limitation was introduced for current implementation based on the decision that CPL and kv_block_size should have same granularity. This is a "To do" for future implementation and probably the only one.

If needed, a warning/status message can be added to alert the user of this limitation when blocking_mode="*kv_paged" and CPL > kv_block_size.

QEfficient/transformers/cache_utils.py
vaibverm: This was switched from using (slot_id + seq_len) to (position_ids.max(1, keepdim=True).values % block_size) because of PyTorch and ONNX runtime generating NaN logits due to padding tokens even though it generated correct logits with QAIC compiled QPC run. 

While reading blocks, we have 3 scenarios:
1) Block is invalid => do no read this block => invalid index assigned => gather_limit is 0
2) Block is fully valid => read the full block => gather_limit is kv_block_size
3) Block is partially valid => only some entries/tokens written in current block => read the block till valid tokens only => gather_limit = valid tokens length.

Since VLLM assigns slot_id at the start of prefill/decode run, so at the start of prefill slot_id is 0. If we write/scatter CPL = 4 tokens in a block with kv_block_size = 8, then we must read/gather only 4 tokens which can be achieved using (slot_id + seq_len) for recently "updated" block. Now, for QAIC runs padding tokens are not counted in seq_len while for PyTorch/ ONNX runtime runs padding tokens are part of seq_len. This leads to reading more tokens than needed for PyTorch and ONNX runtime cases and ends up with NaN logits. Using (position_ids.max(1, keepdim=True).values % block_size) for gather limit avoids this issue for all cases.

QEfficient/transformers/cache_utils.py
vaibverm: Block_table has shape [batch_size, num_kv_blocks] => if key_states have shape [1, num_kv_heads, 1, dh] then block_table will have shape [1, num_kv_blocks]. Hence, in this case we need to read K/V cache from physical location pointed to by block_index'th entry in the block_table and block table has only 1 row. The accuracy of the block_table is ensured by VLLM.  

In CB case, each batch_index does not have separate allocated KV cache space. E.g. if 2 requests in the continuous batch share the same prefix, then only one of them will write to cache (or both of them can write to same location with same data, this is VLLM level optimization) and VLLM will ensure that KV cache for both those requests be read from the same physical KV blocks. 

Let's take an example of num_kv_blocks=4 and BS=2. Let's say prefill wrote to 1 block only, then in decode case the block_table will look like [[0,1,-1,-1], [2,3,-1,-1]] in case no prefix is shared where -1 denotes invalid block which has not been written to yet. If for the same case, we share a prefix, the block table will look like [[0,1,-1,-1], [0,2,-1,-1]]. So, there might be cases where we read the same block for 2 different requests or different blocks and all this is managed by VLLM before the block_table for the current request is sent. 

To tie this back to your question, if batch_size is 1 and num_kv_blocks is 4, then for first request the block_table can be [0,1,2,17] and for second request the block_table can be [4,15,22,80]. With block_index 0, we will read KV cache at physical block 0 for request 1 and physical block 4 for request 2 even though we read row 0 for both. Similarly, scatter will happen at physical block pointed to by (position_ids.max // kv_block_size)th entry in the block_table for that particular request. 

The key idea here is that block_table is not fixed and it updates with input_ids and position_ids updates.

QEfficient/generation/text_generation_inference.py
vaibverm: Thanks for catching this. I have updated QEfficient/generation/vlm_generation.py and QEfficient/transformers/models/modeling_auto.py files. The following file was unpacking only 2 values, have updated it as well: QEfficient/transformers/models/modeling_auto.py

I was expecting to catch such issues in tests.

QEfficient/generation/text_generation_inference.py
vaibverm: Without kv_blocking, num_kv_blocks gets None value and none of the downstream "if num_kv_blocks ..." branches are triggered.

QEfficient/transformers/models/modeling_auto.py
vaibverm: kv_block_Size is not used anywhere if self.supports_paged_attention is False. Hence the fallback to default of kv_block_size = seq would not effect any non-pagedAttention path here. That fallback to default is to ensure seq_len = seq_len if self.supports_paged_attention is False.

Please let me know if you see any additional issues with this.

QEfficient/transformers/models/modeling_auto.py
vaibverm: CPL > kv_block_size is a functional issue, infrastructure wise there will be no issues and export should happen normally. Just that the cache write will not happen correctly due to breaking the assumption on line 320 of QEfficient/transformers/cache_utils.py. Hence, generated output would be wrong in that case.

QEfficient/blocking/blocking_configurator.py
vaibverm: There was an issue with older version of _normalize_attention_mode() in mainline where "h" and "b" blocking modes were dropped. The current code was written to fix that. Since, that issue has been fixed, I have updated the conditions as well to reflect the correct modes.

But there is still an issue in _resolve_effective_blocking_mode() function in current mainline. This function will never resolve to "hqkv" blocking mode because if head_block_size > 1 and num_q_blocks > 1 and num_kv_blocks > 1, then the current mainline function returns "hq" which is incorrect.

I have corrected the mainline function and also added paged version of the modes to the corrected code. Please review to check if any issues still remain.

QEfficient/blocking/blocking_configurator.py
vaibverm: Same as above. Corrected.

QEfficient/blocking/blocking_configurator.py
vaibverm: This has been corrected in the latest commit. Details in the comment above.

QEfficient/utils/generate_inputs.py
quic-hemagnih, line 164: Shouldn't it be self._is_paged_attention()? What's the intention here?

QEfficient/blocking/attention_blocking.py
quic-hemagnih, line 162: This peice of code will never execute, as at line#159, sliding _window should be none to enter in the code leg.
Now at this line you are checking that sliding window is not NONE.

QEfficient/transformers/cache_utils.py
```diff
@@ -246,6 +248,92 @@ def read_only_blockedKV(self, start_index, end_index, cache_kwargs):
...
+        else:
+            invalid_idx_value = 0
+
+        ctx_indices = torch.where(invalid_mask, invalid_idx_value, ctx_indices)
+
+        k_out = CtxGatherFuncPagedAttention.apply(k_out, block_indices, ctx_indices)
+        v_out = CtxGatherFuncPagedAttention.apply(v_out, block_indices, ctx_indices)
+
+        v_out = torch.where((invalid_mask.unsqueeze(1)).unsqueeze(-1), torch.tensor(0.0, dtype=torch.float32), v_out)
+        return k_out, v_out
```
quic-hemagnih, line 301: Any reasons why k_out is not masked here? If we don't mask k_out then it might set a wrong current_max in the online softmax accumulator, which corrupts the numerical normalization of all subsequent valid blocks, producing a near-zero output instead of the correct attention-weighted value?
What do you think?

QEfficient/transformers/cache_utils.py
```diff
@@ -246,6 +248,92 @@ def read_only_blockedKV(self, start_index, end_index, cache_kwargs):
...
+
+    def write_only_pagedAttention(self, key_states, value_states, cache_kwargs):
```
quic-hemagnih, line 303: Align function naming conventions with existing code of QEFF

QEfficient/blocking/blocked_attention_forwards.py
```diff
@@ -208,6 +208,121 @@ def blocked_kv_attention_forward(
...
+        device=query.device,
+    )
+    current_denominator = torch.zeros(batch_size, num_heads, seq_len, device=query.device)
+
+    if torch.onnx.is_in_onnx_export():
+        attention_mask = None
+        use_causal_mask = True
+    position_ids = cache_kwargs.get("position_ids")
+    block_table = cache_kwargs.get("block_table")  # [BS, num_kv_blocks] -> each entry is block_id value
+    kv_block_size = past_key_value.get_seq_length() if past_key_value is not None else 0
```
quic-hemagnih, line 247: Why don't we use get_seq_lengthPagedAttention() in place of get_seq_length()?

For paged attention, keys has shape [total_num_kv_blocks, num_kv_heads, kv_block_size, dh], so keys.shape[-2] is kv_block_size — which happens to be
correct. But this is fragile as based on the paged layout. The newly added get_seq_lengthPagedAttention() method exists precisely for this purpose.

quic-hemagnih: This comment applies for other instances too

anujgupt-github:
Can you add tests for the cases below?

- `get_compilation_dims()` should still support existing callers that unpack 3 values by default.
- `num_kv_blocks` should be covered through an explicit opt-in path, not by changing the default return shape.
- Non-paged generation should behave unchanged when `num_kv_blocks` is absent.
- `CPL > kv_block_size` should not export into a wrong-output path.
- A paged cache write where `slot_id + seq_len > kv_block_size` should be covered.
- CB should be covered with two active slots using different physical KV blocks in `block_table`.
- CB prefix sharing should be covered where two slots share some physical blocks and then diverge.
- Non-paged blocking mode resolution should be covered for `kv`, `qkv`, `hq`, `hkv`, and `hqkv`.
- Paged attention disabled + blocking enabled should be covered to prove old blocked attention behavior is unchanged.
- Unsupported paged configs should be covered so we fail clearly instead of exporting a graph that can generate wrong output.

QEfficient/transformers/cache_utils.py
```diff
@@ -644,6 +732,53 @@ def read_only_blockedKV(self, start_index, end_index, layer_idx, cache_kwargs):
         """
         return self.layers[layer_idx].read_only_blockedKV(start_index, end_index, cache_kwargs)
 
+    def read_only_pagedAttention(self, block_index, updated, layer_idx, cache_kwargs):
+        # def read_only_pagedAttention(self, start_index, end_index, layer_idx, cache_kwargs):
```
quic-hemagnih, line 727: remove this line

quic-hemagnih (changes requested):
1) Add unit test cases for  read_only_pagedAttention / write_only_pagedAttention on QEffLayerCache

**Suggested Optimizations to reduce unnecessary loops-**
1) read_only_pagedAttention() is being called redundantly inside the head/q/batch loops, even though its inputs only depend on KV block index j.

For each j:
- block_index = block_table[:, j]
- updated = (position_ids.max(...) // kv_block_size) == j
- cache_kwargs is unchanged

These do not depend on head_block_idx, q_block_idx, or b_block_idx. As a result, the same expensive CtxGatherFuncPagedAttention / GatherND over the full KV cache is repeated multiple times, and only cheap head/batch slicing differs afterward.

For example:
- blocked_hqkv: calls = num_head_blocks × num_q_blocks × num_kv_blocks, but only num_kv_blocks unique gathers are needed.
- blocked_bhqkv: calls = num_head_blocks × num_q_blocks × num_batch_blocks × num_kv_blocks, but again only num_kv_blocks unique gathers are needed.

Suggest hoisting the gather outside the outer loops:
1. Gather k/v once per KV block j.
2. Cache the gathered k/v states.
3. Reuse them inside head/q/batch loops and apply only the required slicing there.

This should reduce redundant CtxGatherFuncPagedAttention calls significantly, e.g. from 64 to 8 in hqkv and from 128 to 8 in bhqkv for the given configs, with the tradeoff of holding gathered KV blocks in memory temporarily.

2) Same applies to blocked_bhqkv_paged_attention_forward also.

QEfficient/blocking/blocked_attention_forwards.py
vaibverm: This was done to keep the paged and non-paged paths completely isolated from each other while developing the PagedAttention implementation for the first time and also to allow different attention calculation optimizations for each path, if needed. This separation was also done to follow QEfficient infrastructure style where different forwards are picked for different blocking methods. 

Once PagedAttention is reviewed, I can share helper functions to keep one set of forward.

QEfficient/blocking/attention_blocking.py
vaibverm: In the current implementation, PagedAttention hasn't been evaluated with sliding window attention yet. The reason is that in sliding window attention, different layers will require different max num_kv_blocks and how VLLM handles this heterogeneity hasn't been studied yet. This will be done in a future implementation.

QEfficient/blocking/blocked_attention_forwards.py
anujgupt-github: It would help if you can update this now. Will be easier to review

QEfficient/blocking/attention_blocking.py
anujgupt-github: Better to assert the path which is not supported instead of letting it fall through

QEfficient/blocking/attention_blocking.py
vaibverm: I was trying to stay consistent with the current QEfficient blocking infrastructure where various blocking modes each have a separate forward and a separate mode. But I agree, we can probably introduce another qaic_config flag like "supports_pagedAttention" to switch between the 2 KV blocking modes. That would be an added consolidation of the existing code but wouldn't change functionality.

QEfficient/blocking/attention_blocking.py
vaibverm: This is needed to propagate the block_table and slot_id values form ONNX inputs down to attention forwards. In the else condition, these will just take the None values. In my opinion, the placeholders are still needed in the case where PagedAttention is supported. I would love to know if any alternative can work.

QEfficient/blocking/blocking_configurator.py
vaibverm: I am not able to fully understand the reason for this observation. We have both "paged" in blocking_mode and "paged" not in blocking_mode if statements. Why do you think only kv_paged is possible?

QEfficient/blocking/attention_blocking.py
vaibverm: Updated the function to remove the unwanted args. Thanks for the good catch.

QEfficient/transformers/cache_utils.py
vaibverm: I will leave the final decision to the QEff team regarding the case to use. My reasoning was that PagedAttention is a single word in the original paper: https://arxiv.org/pdf/2309.06180, so probably shouldn't be broken into 2 words like paged_attention. But please let me know if we need a clean snake_case for names involving PagedAttention and I can do that.

QEfficient/generation/text_generation_inference.py
```diff
@@ -800,6 +815,10 @@ def run_prefill(self, prompt, generation_len, prefill_logit_bs=1, decode_batch_i
         inputs["position_ids"] = np.where(inputs.pop("attention_mask"), np.arange(padded_len), -1)
         inputs.pop("token_type_ids", None)
 
+        if block_table is not None:
+            inputs["block_table"] = block_table
+            inputs["slot_id"] = np.zeros((prefill_logit_bs), dtype=np.int64)
```
quic-rishinr, line 820: Slot_id is initialized and never updated. Shouldnt we update the slot ID after each chunk similar to input ids and position ids in loop 851 to 875?

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -2461,6 +2473,8 @@ def kv_offload_generate(
             lang_inputs["mm_token_type_ids"] = np.zeros_like(
                 lang_inputs["input_ids"], dtype=lang_inputs["mm_token_type_ids"].dtype
             )
+        if num_kv_blocks:
+            lang_inputs["slot_id"] = (np.max(lang_inputs["position_ids"]) % kv_block_size).reshape(batch_size)
```
quic-rishinr, line 2537: np.max(lang_inputs["position_ids"]) would fail if the bs > 1. this will return a scalar. You would need to fetch the max value for all batch ids.

QEfficient/utils/generate_inputs.py
quic-rishinr: it should be self._is_paged_attention()

QEfficient/blocking/attention_blocking.py
quic-rishinr: better to make the paged as a bool on the blocking config at the KV read/write step. that way we can remove a lot of duplicate code.

quic-rishinr:
@vaibverm can you please rebase the PR against main?

QEfficient/transformers/models/llama/modeling_llama.py
vaibverm: Only the causalLM and ImageTextToText models which support KV Blocking should carry block_table and slot_id. Non causalLM models which do not use KV cache should not need to carry these inputs.

vaibverm:
> @vaibverm can you please rebase the PR against main?

Have done the rebase. But the changes in cache_utils.py for paged gather and scatter functions need to be done in line with mainline changes for the dynamo path coming from QEfficient/customop/utils.py. I need to understand in more detail about dynamo related changes.

quic-rishinr:
@vaibverm can you resolve the comments which you have already addressed? this will make it easier for reviewers while revieing the PR. Also please fix the lint issues.

quic-rishinr:
ci_ready

QEfficient/blocking/blocking_configurator.py
athavale-shivani: Resolved.

QEfficient/generation/text_generation_inference.py
athavale-shivani: Resolved.

QEfficient/generation/text_generation_inference.py
athavale-shivani: Resolved.

QEfficient/transformers/cache_utils.py
athavale-shivani: Resolved.

QEfficient/transformers/cache_utils.py
athavale-shivani: Resolved.

QEfficient/transformers/models/modeling_auto.py
aditjadh: Change self.paged_attention to a local variable? going through Blocking_kwargs as the single source of truth?

QEfficient/blocking/attention_blocking.py
aditjadh: Resolved paged attention through a single config flag

QEfficient/blocking/attention_blocking.py
aditjadh: Resolved

QEfficient/blocking/attention_blocking.py
aditjadh: Resolved

QEfficient/generation/text_generation_inference.py
athavale-shivani: Resolved — slot_id is now recomputed per prefill chunk as chunk_start_position_id % kv_block_size (same formula decode already uses), instead of staying fixed at 0 for every chunk. This was causing later chunks to silently overwrite earlier chunks' rows whenever kv_block_size spans more than one prefill chunk.

QEfficient/customop/ctx_scatter_gather.py
athavale-shivani: Resolved — added data = data.clone() at the top of both CtxScatterFunc.forward and CtxScatterFuncPagedAttention.forward, matching the 3D scatter variants (CtxScatterFunc3D, CtxScatterFunc3DGeneralized, CtxScatterFunc3DInt) which already clone before their in-place scatter.

QEfficient/transformers/cache_utils.py
athavale-shivani: Based on my analysis, k_out doesn't need masking because the code always masks the scores computed from it before those scores are used for anything. Any garbage slot in k_out is, by definition, a slot the sequence hasn't written to yet. The causal mask always blocks future positions, so that garbage score gets overwritten before it can affect current_max or the final output. 
We can resolve this if you agree, or I can look into it if you have any suggestions.

QEfficient/transformers/models/modeling_auto.py
aditjadh: Resolved

QEfficient/transformers/models/modeling_auto.py
aditjadh: Resolved

QEfficient/transformers/models/modeling_auto.py
aditjadh: Resolved

QEfficient/transformers/models/modeling_auto.py
quic-rishinr: we should raise an expectation with NotImplementedError in this case.

quic-rishinr:
> Can you add tests for the cases below?
> 
> * `get_compilation_dims()` should still support existing callers that unpack 3 values by default.
> * `num_kv_blocks` should be covered through an explicit opt-in path, not by changing the default return shape.
> * Non-paged generation should behave unchanged when `num_kv_blocks` is absent.
> * `CPL > kv_block_size` should not export into a wrong-output path.
> * A paged cache write where `slot_id + seq_len > kv_block_size` should be covered.
> * CB should be covered with two active slots using different physical KV blocks in `block_table`.
> * CB prefix sharing should be covered where two slots share some physical blocks and then diverge.
> * Non-paged blocking mode resolution should be covered for `kv`, `qkv`, `hq`, `hkv`, and `hqkv`.
> * Paged attention disabled + blocking enabled should be covered to prove old blocked attention behavior is unchanged.
> * Unsupported paged configs should be covered so we fail clearly instead of exporting a graph that can generate wrong output.

@athavale-shivani please check if all these are checked and addressed.

QEfficient/transformers/models/modeling_auto.py
quic-rishinr: this code is failing in the unit test. please check

QEfficient/transformers/cache_utils.py
quic-rishinr: Please add eager tests for two CB slots using different physical blocks and shared prefix.

For example, as you mentioned

kv_block_size = 4
block_table = [[3, 7], [3, 12]]
position_ids = [[4], [4]]
slot_id = [0, 0]

Both requests should read the shared prefix from block 3, while their next writes must go separately to blocks 7 and 12. Also verify replacement prefill passes the correct block_table row. Once these pass, we can resolve this comment.

QEfficient/blocking/attention_blocking.py
quic-rishinr: Please add a NotImplemented exception in this case.

QEfficient/transformers/models/modeling_auto.py
quic-rishinr: The generic CausalLM path was fixed, but Qwen2.5-VL, Qwen3-VL, and Qwen3-VL-MoE still declare num_kv_blocks dynamic. Dual QPC VLM prefill keeps slot_id at zero for all chunks.

quic-rishinr:
> 1. Add unit test cases for  read_only_pagedAttention / write_only_pagedAttention on QEffLayerCache
> 
> **Suggested Optimizations to reduce unnecessary loops-**
> 
> 1. read_only_pagedAttention() is being called redundantly inside the head/q/batch loops, even though its inputs only depend on KV block index j.
> 
> For each j:
> 
> * block_index = block_table[:, j]
> * updated = (position_ids.max(...) // kv_block_size) == j
> * cache_kwargs is unchanged
> 
> These do not depend on head_block_idx, q_block_idx, or b_block_idx. As a result, the same expensive CtxGatherFuncPagedAttention / GatherND over the full KV cache is repeated multiple times, and only cheap head/batch slicing differs afterward.
> 
> For example:
> 
> * blocked_hqkv: calls = num_head_blocks × num_q_blocks × num_kv_blocks, but only num_kv_blocks unique gathers are needed.
> * blocked_bhqkv: calls = num_head_blocks × num_q_blocks × num_batch_blocks × num_kv_blocks, but again only num_kv_blocks unique gathers are needed.
> 
> Suggest hoisting the gather outside the outer loops:
> 
> 1. Gather k/v once per KV block j.
> 2. Cache the gathered k/v states.
> 3. Reuse them inside head/q/batch loops and apply only the required slicing there.
> 
> This should reduce redundant CtxGatherFuncPagedAttention calls significantly, e.g. from 64 to 8 in hqkv and from 128 to 8 in bhqkv for the given configs, with the tradeoff of holding gathered KV blocks in memory temporarily.
> 
> 2. Same applies to blocked_bhqkv_paged_attention_forward also.

@athavale-shivani please check this

quic-rishinr:
> changes for the dynamo path coming from QEfficient/customop/utils.py. I need to understand in more detail about dynamo related changes.

@vaibverm/@athavale-shivani we will take up the dynamo related changes in the separate PR. We will restrict scope for this PR to get the paged attention working on Causal models.

QEfficient/transformers/cache_utils.py
quic-rishinr: We will stick with the snake_case. Please update.

QEfficient/blocking/blocking_configurator.py
quic-rishinr: @kdulla Can you clarify?

QEfficient/transformers/models/modeling_auto.py
aditjadh: Needs confirmation for no remaining changes

QEfficient/blocking/blocking_configurator.py
aditjadh: Resolved

QEfficient/blocking/blocked_attention_forwards.py
aditjadh: Resolved. A single blocked unified attention forward method for each strategy that implements both paged and non-paged behavior

QEfficient/blocking/attention_blocking.py
aditjadh: Resolved

QEfficient/blocking/attention_blocking.py
aditjadh: Resolved

QEfficient/blocking/attention_blocking.py
aditjadh: Comment pending for review

QEfficient/blocking/blocking_configurator.py
aditjadh: Resolved, blocking mode to match mainline

QEfficient/transformers/cache_utils.py
aditjadh: Resolved

QEfficient/utils/generate_inputs.py
aditjadh: Resolved

QEfficient/transformers/cache_utils.py
aditjadh: Resolved to match snake case

quic-rishinr:
ci_ready

quic-rishinr:
CI-Ready

ochougul:
CI-Ready

QEfficient/blocking/blocking_configurator.py
athavale-shivani: Marked resolved

QEfficient/transformers/models/llama/modeling_llama.py
athavale-shivani: Need further clarification on how to proceed/ to resolve this comment.

QEfficient/utils/generate_inputs.py
athavale-shivani: Marked resolved.

QEfficient/blocking/attention_blocking.py
athavale-shivani: Marked resolved.

QEfficient/transformers/cache_utils.py
athavale-shivani: Need further clarification on how to proceed/ to resolve this comment.

QEfficient/generation/text_generation_inference.py
athavale-shivani: Need further clarification on how to proceed/ to resolve this comment.

QEfficient/transformers/cache_utils.py
athavale-shivani: Today it returns allocated capacity: self.keys.shape[-2] * self.keys.shape[0] (block size × number of allocated blocks) — fixed at allocation time, doesn't shrink/grow with how many tokens have actually been written for a given request. Right now, this method has zero callers anywhere outside its own definition. Do you see a use case for returning active sequence length? Would be glad to receive further direction on this.

QEfficient/blocking/blocked_attention_forwards.py
athavale-shivani: Checked this — get_seq_length() and get_seq_length_paged_attention() return different quantities. get_seq_length() returns self.keys.shape[-2], which for the paged layout is kv_block_size (per-block size). get_seq_length_paged_attention() returns self.keys.shape[-2] * self.keys.shape[0] — total capacity across all blocks.
get_seq_lengt() is used to get kv_block_size to size each block read — swapping in the paged-specific method would change that to total capacity and break block indexing. So current usage looks intentional rather than fragile. 

Please let me know your thoughts/ further direction on this.

QEfficient/transformers/models/modeling_auto.py
aditjadh: Resolved

QEfficient/transformers/cache_utils.py
athavale-shivani: Added — TestPagedAttentionBlockTableRegression in tests/unit_test/models/test_cache_correctness.py, using exactly the example you gave: kv_block_size=4, block_table=[[3,7],[3,12]], shared logical block 0 → physical block 3, divergent logical block 1 → physical blocks 7/12.

Four eager tests, all passing:
  - test_shared_prefix_written_to_common_physical_block — both rows prefill and land in physical block 3.
  - test_divergent_decode_writes_go_to_separate_physical_blocks — after the shared prefix, a decode step writes row 0's new token to block 7 and row 1's to block 12, confirms they don't bleed into each other or into block 3, and confirms untouched blocks stay zero.
  - test_read_back_gathers_shared_prefix_and_divergent_tail_per_row — reads back logical block 0 and confirms both rows see the identical shared prefix, then logical block 1 and confirms each row sees its own divergent tail.
  - test_replacement_prefill_uses_request_specific_block_table_row — covers your "verify replacement prefill passes the correct block_table row" ask specifically: two rows whose logical block 0 maps to two different physical blocks (3 and 9, not sharing a prefix), confirms each row's prefill write lands in its own physical block rather than both collapsing onto row 0's.**

QEfficient/transformers/models/modeling_auto.py
athavale-shivani: Done — write_only_paged_attention (QEfficient/transformers/cache_utils.py:327-338) now raises NotImplementedError in exactly this case, before any cache write happens:                                                                                     

  if seq_len > block_size:
      raise NotImplementedError(                                                                                                 
          f"write_only_paged_attention only supports writing within a single KV block per call "
          f"(seq_len={seq_len} > block_size={block_size}). Prefill/decode chunks larger than "
          f"block_size must be split by the caller before writing to a paged KV cache."
      )                                                          
  if int((slot_id + seq_len).max()) > block_size:
      raise NotImplementedError(                                 
          f"write_only_paged_attention write would cross a KV block boundary "
          f"(slot_id + seq_len > block_size={block_size}). Writes must stay within a single "
          f"physical block; split the write across blocks before calling this method."
      )                                                          

CPL > kv_block_size write now fails loudly instead of silently landing in the wrong physical block and producing wrong generated output. Covered by TestPagedAttentionSingleBlockWriteLimit in tests/unit_test/models/test_cache_correctness.py (eager, CPU-only): within-block write succeeds, seq_len > block_size raises, slot_id + seq_len > block_size raises, and the exact-boundary case (slot_id + seq_len == block_size) still succeeds.

QEfficient/transformers/models/modeling_auto.py
aditjadh: Resolved

>

QEfficient/blocking/blocking_configurator.py
aditjadh: Resolved

QEfficient/transformers/cache_utils.py
athavale-shivani: Removed this function as it was unused.

QEfficient/blocking/blocked_attention_forwards.py
athavale-shivani: Removed get_seq_length_paged_attention function as it was unused.

QEfficient/transformers/models/modeling_auto.py
athavale-shivani: Good catch — 41beb26 only fixed the generic CausalLM export path; qwen2_5_vl.py, qwen3_vl.py, and qwen3_vl_moe.py still declared {0: "batch_size", 1: "num_kv_blocks"} on block_table. Fixed in 3496aa1 — mirrored the same one-line change (drop the 1: "num_kv_blocks" entry, since the block count is fixed at export/trace time, not a real ONNX dynamic axis) into all three Qwen files. Pushed to PR_branch.

On the VLM slot_id bug (never built in _execute_chunked_prefill): slot_id was never populated at all for any of the 3 VLM prefill entry points in vlm_generation.py, same root mechanism as the CausalLM bug you flagged on text_generation_inference.py:820. Fixed in 4f4ce2e: slot_id is now recomputed per prefill chunk instead of staying frozen/unset before the chunk loop.

While fixing that, I noticed VisionLanguageGeneration had no way to actually populate block_table/num_kv_blocks at all — the dual-QPC constructor never wired it through, so paged attention on the language QPC was unreachable at runtime even though the export-side plumbing already supported it for qwen2_5_vl/qwen3_vl/qwen3_vl_moe. Added that wiring in 425c8e8: num_kv_blocks is now an optional constructor param (mirrors the non-VLM TextGeneration path), and self.block_table stays None/inert when it's not supplied — no behavior change for existing non-paged VLM callers.

athavale-shivani:
ci_ready

quic-rishinr:
CI-Ready

quic-rishinr:
CI-Ready

quic-rishinr:
CI-Ready

QEfficient/blocking/blocked_attention_forwards.py
```diff
@@ -123,6 +123,31 @@ def update_running_softmax_prefill(
...
 
+def _read_kv_block(
+    *,
+    past_key_value: Cache,
+    start_index: int,
+    end_index: int,
+    layer_idx: int,
+    kv_block_size: int,
+    paged_attention: bool,
+    j: int,
```
quic-rishinr: nit : use proper variable naming for J something like kv_block_idx.

QEfficient/blocking/blocked_attention_forwards.py
```diff
@@ -1074,6 +1126,39 @@ def blocked_qkv_attention_forward(
     if sinks is not None:
         sinks = sinks.reshape(1, -1, 1, 1).expand(batch_size, -1, seq_len, -1)
 
+    # Gather each KV block once: block_index/updated/cache_kwargs only depend on `j`,
+    # not on q_block_idx, so hoist the read out of the q-block loop below.
+    kv_blocks = []
+    for j in range(num_kv_blocks):
```
quic-rishinr: rename the variable name j to kv_block_idx

QEfficient/blocking/blocked_attention_forwards.py
```diff
@@ -1217,6 +1291,39 @@ def blocked_hqkv_attention_forward(
     if sinks is not None:
         sinks = sinks.reshape(1, -1, 1, 1).expand(batch_size, -1, seq_len, -1)
 
+    # Gather each KV block once: block_index/updated/cache_kwargs only depend on `j`,
+    # not on head_block_idx/q_block_idx, so hoist the read out of the loops below.
+    kv_blocks = []
+    for j in range(num_kv_blocks):
```
quic-rishinr: use the kv_block_idx as variable name instead of j

tests/transformers/models/causal_lm_models/test_causal_lm_blocking_hqkv.py
```diff
@@ -57,6 +57,18 @@ def test_full_causal_all_blocking_pytorch_vs_kv_vs_ort_vs_ai100(model_name, bloc
         num_devices=4,
     )
 
+    # head qkv_paged_attention blocking
```
quic-rishinr: we should not be running paged attention for all the models. this would add another 1 hour to the current CI time which is not acceptable, lets reduce the configs and keep the testing limited for few priority models.

quic-rishinr (changes requested):

quic-rishinr:
@athavale-shivani please fix the lint and Unit test failures

QEfficient/customop/ctx_scatter_gather.py
```diff
@@ -42,6 +42,12 @@ class CtxScatterFunc(torch.autograd.Function):
 
     @staticmethod
     def forward(data: torch.Tensor, position_ids: torch.Tensor, updates: torch.Tensor):
+        if not torch.onnx.is_in_onnx_export():
```
quic-rishinr, line 45: @athavale-shivani why do you want this change in regular scatter method? this could increse the latency even if the paged attention is disabled. can we remove this?

quic-rishinr (changes requested):

ochougul:
CI-Ready

qraniumcitest:
CI-Ready

QEfficient/blocking/blocked_attention_forwards.py
aditjadh: Resolved

QEfficient/blocking/blocked_attention_forwards.py
aditjadh: Resolved

QEfficient/blocking/blocked_attention_forwards.py
aditjadh: Resolved

QEfficient/customop/ctx_scatter_gather.py
athavale-shivani: It was added per @anujgupt-github earlier review comment flagging that the in-place scatter here mutates data directly, so a caller holding a reference to the pre-scatter tensor (e.g. for an eager-parity comparison) would see it silently change underneath them. It was applied to both CtxScatterFunc.forward and CtxScatterFuncPagedAttention.forward to match the 3D scatter variants (CtxScatterFunc3D, CtxScatterFunc3DGeneralized, CtxScatterFunc3DInt), which already clone before scattering. If you do not want CtxScatterFunc to change, I can remove it.

quic-rishinr:
CI-Ready

QEfficient/customop/ctx_scatter_gather.py
quic-rishinr: This will add additional latency lets remove the clone for regular scatter

quic-rishinr:
CI-Ready

quic-rishinr:
CI-Ready

quic-rishinr (approved):
LGTM!! Thanks Shivani, Aditya and Vaibhav

quic-rishinr:
CI-Ready

quic-rishinr:
CI-Ready

quic-rishinr:
CI-Ready

PR #1185, opened 2026-07-15
Qwen3.6_Tiling 

mohiso22:
This PR is for performance related changes on qwen3.5 family of models

PR #1212, opened 2026-07-24
feat: Replace num_replicate_kv_heads with boolean replicate_kv_heads flag

quic-dhirajku:
- Replace user-facing num_replicate_kv_heads int param in qaic_configwith a boolean replicate_kv_heads flag; value computation is now fully internal and never exposed to the user.
  - maybe_apply_replicate_kv_transform reads only replicate_kv_headsfrom qaic_config; computed repeat factor is kept in hash_params and never written back into qaic_config.
  - Resolve model_config.text_config before calling calculate_num_replicate_kv_heads so VLM top-level configs correctly surface text-side head counts.

### DIFF
--- a/QEfficient/base/modeling_qeff.py
+++ b/QEfficient/base/modeling_qeff.py
@@ -125,15 +125,19 @@ class QEFFBaseModel(ABC):
         if model_config is None or qaic_config is None or "EncoderWrapper" in self.model.__class__.__name__:
             return 1
 
-        num_replicate_kv_heads = qaic_config.get("num_replicate_kv_heads")
-        if num_replicate_kv_heads is None:
-            num_replicate_kv_heads = calculate_num_replicate_kv_heads(
-                num_devices=num_devices,
-                text_model_config=model_config,
-            )
-        if num_replicate_kv_heads is not None:
-            num_replicate_kv_heads = int(num_replicate_kv_heads)
-            qaic_config["num_replicate_kv_heads"] = num_replicate_kv_heads
+        replicate_kv_heads = qaic_config.get("replicate_kv_heads", False)
+
+        if not replicate_kv_heads:
+            return 1
+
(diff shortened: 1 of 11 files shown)


QEfficient/transformers/models/modeling_auto.py
```diff
@@ -1612,10 +1612,11 @@ def export(
 
         # TODO: remove the current pt weight offload capability once CustomLoader is in place
         if offload_pt_weights is None:
-            if prefill_only and prefill_seq_len > 1:
+            if prefill_only and prefill_seq_len is not None and prefill_seq_len > 1:
                 offload_pt_weights = False  # keep weights resident for the decode export
             else:
-                offload_pt_weights = True
+                num_replicate_kv_heads = self.lang_model.hash_params.get("num_replicate_kv_heads", 1)
+                offload_pt_weights = num_replicate_kv_heads <= 1
```
ochougul: This change is not required.

QEfficient/base/modeling_qeff.py
```diff
@@ -125,15 +125,19 @@ def maybe_apply_replicate_kv_transform(self, model_config, num_devices: int, qai
...
+        if not replicate_kv_heads:
+            return 1
+
+        text_config = getattr(model_config, "text_config", None)
+        effective_config = text_config if text_config is not None else model_config
+
+        num_replicate_kv_heads = calculate_num_replicate_kv_heads(
+            num_devices=num_devices,
+            text_model_config=effective_config,
+        )
```
ochougul, line 139: Also update this logic such that, we replicate only when `num_devices>num_kv_heads and num_devices%num_kv_heads==0`

ochougul (changes requested):

ochougul (approved):

ochougul:
being addressed in #1185

PR #1220, opened 2026-07-27
feat(weight-free): weight-free ONNX export for causal LMs

amarquic:
## Summary

  Adds weight-free ONNX export for causal LM models on top of the dynamo export infrastructure.

  Weight-free export builds the model on meta device (no weights in RAM), exports ONNX graph structure only, and lets the QAIC compiler load
  weights directly from the original safetensors checkpoint at compile time via `weight_spec.json`.

  ## New files

  - `QEfficient/exporter/weight_free/` — core weight-free export, weight spec format, checkpoint prep pipeline (MoE expert stacking, dtype
  conversion, .bin→safetensors auto-convert)
  - `examples/text_generation/weight_free/` — single-model smoke test, CB weight-free smoke test

  ## Key changes

  - `modeling_qeff.py` — `use_weight_free_export` flag; reorder `example_inputs`/`dynamic_shapes` by `forward()` signature so `torch.export`
  binds them correctly; skip `SplitTensorsTransform`; embed `weight_spec.json` as ONNX metadata; symlink prepared checkpoint next to ONNX
  - `modeling_auto.py` — `use_weight_free_export` wired through `export()` and `compile()`; auto-enable `dynamo=True`
  - `onnx_transforms.py` — disable unsupported transforms for dynamo path
  - `export_utils.py` — separate dynamo/TorchScript subfunction setup paths
  - `torch_patches.py` — `temporarily_enable/disable_nested_compile_regions()`
  - `core.py` — dtype suffix in prepared checkpoint dir; prune fake initializers after meta-device export

  ## Model fixes

  - `codegen` — fix `embed_positions` device/dtype for weight-free subfunction tracing
  - `glm4_moe` — `logits.float()` to fix runtime buffer mismatch; rotary cos/sin device sync; `_rotary_dim` attribute; MoE expert placeholder
  dtype
  - `granitemoe`, `mixtral`, `phi3`, `qwen3_moe` — rotary cos/sin device/dtype sync for meta-device tracing; MoE expert placeholder dtype
  - `continuous_batching.py` — fix `use_dynamo=True` → `dynamo=True`

  ## Tests

  ```bash
  pip install -e .[test]
  pytest tests/weight_free/ -m "not on_qaic" -n auto -v

  ```


quic-amitraj:
In PR description, could you please add what model architecture, and their features are enabled with this PR? @amarquic

QEfficient/base/modeling_qeff.py
```diff
@@ -429,6 +431,68 @@ def _export_via_dynamo(
...
+        tmp_onnx_path: Path,
+        example_inputs: Dict[str, torch.Tensor],
+        input_names: List[str],
+        output_names: List[str],
+        dynamic_axes: Dict,
+        export_kwargs: Dict,
+        onnx_transform_kwargs: Optional[Dict] = None,
+    ):
+        """Export via weight-free dynamo path: meta-device model, no embedded weights."""
+        from QEfficient.customop.dynamo_ops import DYNAMO_CUSTOM_OP_TABLE
```
quic-amitraj: `DYNAMO_CUSTOM_OP_TABLE` is already imported at the module level (line 39). This local import is redundant and gives the impression that this is a lazy/conditional load when it isn't. Remove it and use the top-level import.

QEfficient/base/modeling_qeff.py
```diff
@@ -429,6 +431,68 @@ def _export_via_dynamo(
...
+        sig_keys = list(inspect.signature(self.model.forward).parameters.keys())
+        sig_key_set = set(sig_keys)
+        ordered_inputs, ordered_shapes = {}, {}
+        for k in sig_keys:
+            if k in example_inputs:
+                ordered_inputs[k] = example_inputs[k]
+            if k in dynamic_shapes:
+                ordered_shapes[k] = dynamic_shapes[k]
+        example_inputs = {**ordered_inputs, **{k: v for k, v in example_inputs.items() if k not in sig_key_set}}
+        dynamic_shapes = {**ordered_shapes, **{k: v for k, v in dynamic_shapes.items() if k not in sig_key_set}}
```
quic-amitraj: This input/dynamic_shapes reordering block (lines 454–463) is copy-pasted verbatim from `_export_via_dynamo` (lines 388–398). Same for the dynamo kwargs setup just below (lines 465–473 here, lines 400–408 there) and the `TORCH_INVOKE_ALLOW_CREATE_FALLBACK` save/restore (lines 475–492 here, lines 409–432 there).

Since both methods are being added/modified in this PR, this is the right time to factor the shared logic out rather than carry the duplication forward. Suggestions:
- `_reorder_inputs_by_signature(self, inputs, shapes)` for the ordering logic.
- A `@contextmanager` for the env var patch (fits well in `torch_patches.py` alongside the others).
- `_build_dynamo_export_kwargs(self, export_kwargs)` for the kwargs setup.

This keeps the two export paths clearly parallel and avoids a future where they drift apart.

QEfficient/base/modeling_qeff.py
```diff
@@ -554,7 +634,31 @@ def _resolve_pkv_names(layer_idx, layer_state):
...
+                    dynamic_axes=dynamic_axes,
+                    export_kwargs=export_kwargs,
+                    onnx_transform_kwargs=onnx_transform_kwargs,
+                )
+                shutil.move(str(tmp_onnx_path), str(onnx_path))
+                tmp_weight_spec = tmp_onnx_dir / "weight_spec.json"
+                final_weight_spec = export_dir / "weight_spec.json"
+                if tmp_weight_spec.exists():
+                    shutil.move(str(tmp_weight_spec), str(final_weight_spec))
+                    self.weight_spec_path = str(final_weight_spec)
```
quic-amitraj: `self.weight_spec_path` is assigned two different types across two paths:
- Line 551 (early-return): `_weight_spec_path` — a `Path` object
- Line 655 (here): `str(final_weight_spec)` — a `str`

The annotation on line 161 says `Optional[str]`. Both assignments should be consistent. Since `onnx_path` and `qpc_path` follow the `Optional[str]` convention in this class, cast the line 551 assignment with `str(...)` as well.

QEfficient/base/modeling_qeff.py
```diff
@@ -600,6 +713,18 @@ def _resolve_pkv_names(layer_idx, layer_state):
             model.metadata_props.append(
                 onnx.StringStringEntryProto(key="qeff_transforms", value=",".join(self._transform_names()))
             )
+            # For weight-free export, embed weight_spec.json as ONNX metadata so the
+            # QAIC compiler can locate and load the external checkpoint weights at compile time.
+            if use_weight_free_export and self.weight_spec_path is not None:
+                import json as _json
+
+                from QEfficient.exporter.weight_free.core import _upsert_metadata_prop
```
quic-amitraj: Two things here:

1. `_upsert_metadata_prop` is a private function (underscore prefix) being imported from another module. Importing private symbols across module boundaries breaks encapsulation. Consider exposing a public helper in `core.py` — e.g. `embed_weight_spec_as_metadata(model, weight_spec_path)` — and calling that instead. The base class shouldn't need to know about the `com.qti.aisw.extdata` key.

2. `import json as _json` (line 719) is unnecessary — there's no `json` name at module scope to collide with. Move `import json` to the top-level imports.

QEfficient/base/modeling_qeff.py
```diff
@@ -614,6 +739,18 @@ def _resolve_pkv_names(layer_idx, layer_state):
...
+        # For weight-free export, symlink the prepared checkpoint directory next to
+        # the ONNX so relative paths in weight_spec.json resolve correctly at compile time.
+        if use_weight_free_export and self.weight_spec_path is not None:
+            from QEfficient.exporter.weight_free.spec import load_weight_spec
+
+            spec = load_weight_spec(Path(self.weight_spec_path))
+            prepared_out = Path(spec.model_id)
+            symlink = onnx_path.parent / prepared_out.name
+            if prepared_out.exists() and not symlink.exists():
+                symlink.symlink_to(prepared_out)
```
quic-amitraj: If `symlink_to` raises for any reason like permissions, NFS mount, unsupported filesystem, the exception will propagate as an unexpected failure at the end of an otherwise successful export. Worse, if this is ever caught somewhere higher up and swallowed, the checkpoint path won't resolve at compile time and the error will appear far away from here with no obvious cause.

Please wrap this in `try/except OSError` and emit a `logger.warning` with both the source (`prepared_out`) and target (`symlink`) paths so the developer has something actionable.

QEfficient/base/modeling_qeff.py
```diff
@@ -1006,6 +1145,7 @@ def _compile(
 
         layerwise_cache_probe = compiler_options.pop("_layerwise_cache_probe", False)
         moe_prefill_packed_chunk_size = compiler_options.pop("moe_prefill_packed_chunk_size", None)
+        use_weight_free_export = compiler_options.pop("use_weight_free_export", False)
```
quic-amitraj: `use_weight_free_export` is a proper named parameter in the public `compile()` in `modeling_auto.py`, but here in `_compile` it's buried inside `**compiler_options` and popped out manually. A typo in the key name would silently default to `False` with no error or warning. Promote it to a named parameter in `_compile`'s signature, consistent with how `dynamo`, `use_onnx_subfunctions`, and others are handled.

QEfficient/base/modeling_qeff.py
```diff
@@ -429,6 +431,68 @@ def _export_via_dynamo(
             else:
                 os.environ["TORCH_INVOKE_ALLOW_CREATE_FALLBACK"] = prev_invoke_fallback
 
+    def _export_via_weightfree(
```
quic-amitraj: [Nit] `_export_via_legacy` and `_export_via_dynamo` both annotate `-> None`. `_export_via_weightfree` returns a tuple but has no return type annotation. Please add one (e.g. `-> Tuple[Dict, Callable]`) for consistency and IDE discoverability.

QEfficient/base/modeling_qeff.py
quic-amitraj: Keep this in export_utils.py and reuse for both.

quic-amitraj (commented):
The base class currently contains the full implementations of export backends (`_export_via_legacy`, `_export_via_dynamo`, `layer_wise_export` and `_export_via_weightfree`). 

**Suggestion:** create `QEfficient/exporter/onnx_exporter.py` and move these all functions there:

- `export_via_legacy`
- `export_via_dynamo`
- `_export_via_weightfree`
- `layer_wise_export`

Keep all the common codes(mentioned above) across the dynamo and weightfree export in export_utils.

examples/text_generation/basic_inference.py
```diff
@@ -20,6 +20,7 @@ def main():
     parser.add_argument("--prefill-seq-len", type=int, default=32, help="Prefill sequence length")
     parser.add_argument("--ctx-len", type=int, default=128, help="Context length")
     parser.add_argument("--dynamo", action="store_true", help="Export via dynamo")
+    parser.add_argument("--use_weight_free_export", action="store_true", help="Export via Weight-free")
     parser.add_argument("--use-onnx-subfunctions", action="store_true", help="Use subfunctions while exporting")
```
quic-amitraj, line 23: Please update the script to handle flag `use_weight_free_export`.

scripts/weight_free_runner.py
```diff
@@ -0,0 +1,312 @@
...
+    "compile_peak_rss_mb": 29900.1,
+    "compile_duration_seconds": 26.8,
+    "onnx_size_gb": 0.932,
+    "qpc_dir_size_gb": 13.473,
+    "completions": ["generated text 1", ...],
+    "error": null
+}
+"""
+
+from __future__ import annotations
```
quic-amitraj: This script a benchmarking and profiling tool, not an example. Keep it in the scripts with proper folder/file name.

scripts/weight_free_runner.py
```diff
@@ -0,0 +1,312 @@
...
+# ---------------------------------------------------------------------------
+# CLI
+# ---------------------------------------------------------------------------
+
+
+def _parse_args() -> argparse.Namespace:
+    p = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
+    p.add_argument("--model_name", required=True)
+    p.add_argument("--mode", default="basic", choices=["basic", "cb", "ccl", "cb_ccl"], help="Test mode.")
+    p.add_argument("--weight_free", action="store_true", help="Use weight-free (meta-device) export.")
```
quic-amitraj: Make flags uniform across.

scripts/weight_free_runner.py
```diff
@@ -0,0 +1,312 @@
...
+            )
+
+        # ── Phase 2: Export (PyTorch → ONNX) ──────────────────────────────
+        profiler.mark_operation("Model Export")
+        t_export = time.perf_counter()
+
+        onnx_path = Path(
+            qeff_model.export(
+                export_dir=str(export_dir),
+                use_dynamo=True,
```
quic-amitraj: Its update with `dynamo`, please check once and make it alligned with dynamo changes.

quic-amitraj (changes requested):

examples/text_generation/weight_free/export_compile_infer.py
```diff
@@ -0,0 +1,200 @@
...
+from pathlib import Path
+
+import torch
+from accelerate import init_empty_weights
+from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
+
+from QEfficient.transformers.models.modeling_auto import QEFFAutoModelForCausalLM
+
+
+def parse_args():
```
quic-amitraj: 1. Remove _runner.py from examples/ and move it to scripts/ where it belongs (along with compare_weightfree.py if need, which it references  but is missing)
  2. Simplify both example scripts strip out the --no_weight_free dual-mode handling and make each show weight-free cleanly, similar to how examples/dynamo/causal_lm/basic_dynamo_inference.py is written
  3. Fix the negative flag names to positive (--weight_free default true, or just remove the non-weight-free path)
  4. Fix use_dynamo=True bug in _runner.py
  5. Rename the example scripts by following our pattern.

quic-amitraj:
@amarquic As `layerwise` export planned to be deprecated, let's add warnings, logs as deprecation warnings for `layerwise`.

quic-amitraj:
Regarding the example scripts you have added-

- export_compile_infer.py: the basic weight-free flow is already covered by basic_inference.py with --use_weight_free_export. If a dedicated example is needed, it belongs at examples/dynamo/causal_lm/weight_free_inference.py since dynamo is a prerequisite.
-  continuous_batching.py — should be a flag on the existing CB examples, not a new file in a new folder.
-  No need of init in the example folder, please remove that as well.

examples/text_generation/weight_free/export_compile_infer.py
```diff
@@ -0,0 +1,200 @@
...
+  2. compile() — calls export + QAIC compile in one step
+  3. QPC inference on device
+
+Timing is printed for compile and inference.
+ORT and PyTorch baseline are intentionally omitted (use compare.py for parity checks).
+
+Usage
+-----
+# Weight-free export (default)
+python examples/text_generation/weight_free/export_compile_infer.py \\
```
quic-amitraj: This is also not working, model name is incorrect. please fix it.

examples/text_generation/basic_inference.py
amarquic: done

QEfficient/base/modeling_qeff.py
amarquic: done

QEfficient/base/modeling_qeff.py
amarquic: done

QEfficient/base/modeling_qeff.py
amarquic: done

QEfficient/base/modeling_qeff.py
amarquic: done

QEfficient/base/modeling_qeff.py
amarquic: done

QEfficient/base/modeling_qeff.py
amarquic: Done

QEfficient/base/modeling_qeff.py
amarquic: done

examples/text_generation/weight_free/export_compile_infer.py
amarquic: done

quic-amitraj:
`sin_cached`/`cos_cached` can remain on the device where the model was initialized, while `hidden_states` may be materialized on the runtime/export device in the weight-free path. Move only the indexed rotary slices to `hidden_states.device`; this keeps the fix local and avoids changing QEff init/loading behavior.

Suggested change in the affected forward paths:

```python
# before
sin = self.sin_cached[position_ids].unsqueeze(1)
cos = self.cos_cached[position_ids].unsqueeze(1)

# after
sin = self.sin_cached[position_ids].unsqueeze(1).to(device=hidden_states.device)
cos = self.cos_cached[position_ids].unsqueeze(1).to(device=hidden_states.device)
```

QEfficient/transformers/models/gptj/modeling_gptj.py
```diff
@@ -33,6 +33,7 @@
 
 def apply_rotary_pos_emb(tensor: torch.Tensor, sin: torch.Tensor, cos: torch.Tensor) -> torch.Tensor:
     # Sin Cos are also fixated on fp32 here
+
```
quic-amitraj, line 36: Remove this.

.gitignore
```diff
@@ -93,3 +93,6 @@ qeff_models
 # Agent tool-specific skill links (generated by `make codex` / `make claude`)
 .agents/
 .claude/
+env1/
```
quic-amitraj, line 96: Why?

QEfficient/exporter/weight_free/checkpoint.py
```diff
@@ -0,0 +1,171 @@
...
+    Returns
+    -------
+    Path
+        Directory containing the downloaded or local checkpoint files.
+    """
+    candidate = Path(model_id_or_path).expanduser()
+    if candidate.exists():
+        return candidate
+
+    snapshot_dir = snapshot_download(
```
quic-amitraj, line 35: Please use hf_download api from qeff.

QEfficient/exporter/weight_free/checkpoint.py
```diff
@@ -0,0 +1,171 @@
...
+    model_id_or_path : str
+        Local model path or Hugging Face Hub model id.
+
+    Returns
+    -------
+    Path
+        Directory containing the downloaded or local checkpoint files.
+    """
+    candidate = Path(model_id_or_path).expanduser()
+    if candidate.exists():
```
quic-amitraj, line 32: Not making sense, plese check and remove.

quic-amitraj (changes requested):

PR #1230, opened 2026-07-31
feat: add w_sub ONNX semantic weight + node rename transform

vishwasdivakar:
Adds opt-in semantic renaming for w_sub ONNX exports. When rename_onnx_nodes=True is passed alongside use_onnx_subfunctions=True, opaque ONNX names (e.g., onnx::MatMul_5898) are renamed to their HF equivalents (e.g., model.layers.0.self_attn.q_proj.weight) using 256-byte fingerprint matching against HF safetensors. This makes exported traces self-sufficient for llm_breakdown without a separate rename step.

What it does:

1. RenameWsubTransform in onnx_transforms.py (+310 lines) - fingerprint-matches ONNX external weights to HF params, renames weights + anchored MatMul/RMSNorm nodes, stamps metadata. Handles both individual files and consolidated .onnx.data with offsets. Supports BF16/FP16/FP32 auto-detection.
2. Integration in modeling_qeff.py (+74 lines) - adds rename_onnx_nodes param (default False), resolves HF cache path, calls the transform after existing ONNX transforms. Warns if rename_onnx_nodes=True without use_onnx_subfunctions=True.
3. Kwarg threading in modeling_auto.py (+2 lines) - passes rename_onnx_nodes through to the export call.

Safety: Default is False - zero impact on existing behavior unless explicitly opted in.

Validated Models: LLaMA-3-8B, GPT-OSS 120B and Kimi-K2.5 on AI100

### DIFF
--- a/QEfficient/base/modeling_qeff.py
+++ b/QEfficient/base/modeling_qeff.py
@@ -26,6 +26,7 @@ from QEfficient.base.onnx_transforms import (
     OnnxTransformPipeline,
     PruneFakeInitializersTransform,
     RenameFunctionOutputsTransform,
+    RenameWsubTransform,
     SplitTensorsTransform,
 )
 from QEfficient.base.pytorch_transforms import PytorchTransform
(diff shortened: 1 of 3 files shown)


vishwasdivakar:
@vbaddi, please review!
Thank you.

vbaddi:
++ @quic-rishinr @ochougul

QEfficient/base/modeling_qeff.py
```diff
@@ -602,6 +642,40 @@ def _resolve_pkv_names(layer_idx, layer_state):
...
             logger.info("ONNX transforms applied")
 
+            # W_Sub semantic rename: when enabled, rename opaque weights/nodes
+            # to semantic HF names using fingerprint matching.
+            if rename_onnx_nodes and not export_kwargs.get("use_onnx_subfunctions", False):
+                logger.warning(
+                    "rename_onnx_nodes=True requires use_onnx_subfunctions=True. "
+                    "Skipping semantic rename."
+                )
+            elif rename_onnx_nodes and export_kwargs.get("use_onnx_subfunctions", False):
```
vbaddi, line 652: `export_wrapper` removes `use_onnx_subfunctions` before `_export`, but here we check `export_kwargs["use_onnx_subfunctions"]` inside `_export`. That means `rename_onnx_nodes=True` would usually skip the actual rename.

QEfficient/transformers/models/modeling_auto.py
```diff
@@ -4050,6 +4050,8 @@ def _legacyify_cache(obj):
                 dynamo=dynamo,
                 offload_pt_weights=kwargs.get("offload_pt_weights", True),
                 prefill_only=prefill_only,
+                rename_onnx_nodes=kwargs.get("rename_onnx_nodes", False),
```
vbaddi, line 4053: `rename_onnx_nodes` was not part of the export hash. A renamed and non-renamed ONNX could collide in the same cache directory.

QEfficient/base/modeling_qeff.py
```diff
@@ -62,6 +63,44 @@
 logger = logging.getLogger(__name__)
 
 
+def _resolve_hf_cache_path(pretrained_model_name_or_path: str) -> Optional[str]:
```
vbaddi, line 66: `_resolve_hf_cache_path` duplicated Hugging Face Hub cache resolution and could pick the wrong snapshot/revision. Export code should not infer model revision by scanning cache multiple times.

QEfficient/base/onnx_transforms.py
```diff
@@ -667,3 +667,313 @@ def _set_external_data(tensor, file_name):
...
+    HF safetensors to build a rename map, then applies weight rename + node rename.
+
+    Handles both individual external files (onnx__MatMul_NNN) and consolidated
+    data files (model.onnx.data with offset+length metadata).
+    """
+
+    FINGERPRINT_BYTES = 256
+
+    # Regex: model.layers.N.<role>.weight -> extract <role> for node rename
+    _SEMANTIC_PARAM_RE = re.compile(
```
vbaddi, line 685: we already have this bunch of data from model. Could you pls re-use.

QEfficient/base/onnx_transforms.py
```diff
@@ -667,3 +667,313 @@ def _set_external_data(tensor, file_name):
...
+            if not os.path.exists(shard_path):
+                continue
+            fp = cls._read_hf_fingerprint(shard_path, param_name, target_dtype)
+            if fp:
+                fp_map[fp] = param_name
+
+        return fp_map
+
+    @classmethod
+    def _read_hf_fingerprint(cls, shard_path: str, param_name: str, target_dtype) -> Optional[bytes]:
```
vbaddi, line 860: **Generally:**
- Pls remove `_resolve_hf_cache_path`, `rename_onnx_nodes`, `hf_model_path`, and `RenameWsubTransform` code which is causing the bloat.
- Add a parameter identity renaming to `QEfficient/utils/torch_patches.py`.
- Add a minimal `RenameWsubNodesTransform` for local-function `MatMul`, `Gemm`, and `CustomRMSNorm` nodes.
- We can register the node rename automatically for legacy `use_onnx_subfunctions=True`.
- Also, Add unit tests.

vbaddi (changes requested):

PR #1231, opened 2026-08-02
GPT-OSS: Hoist Layer-Invariant Attention Mask Preparation Outside Decoder Subfunctions

abhishek-singh591:
# GPT-OSS: Hoist Layer-Invariant Attention Mask Preparation Outside Decoder Loop

  ## Summary

  This change hoists GPT-OSS attention-mask preparation out of `QEffGptOssDecoderLayer` and out of the
  decoder-layer loop. The model now computes the reusable full-attention and sliding-window additive
  masks once before iterating over decoder layers, then passes the already-prepared mask matching each
  layer’s attention type.

  ## Problem

  For GPT-OSS 120B decode, enabling ONNX subfunctions showed an approximately 13% inference performance
  drop compared with non-subfunction compilation. Inspection of the exported decoder function showed
  redundant, layer-invariant mask work inside `QEffGptOssDecoderLayer`; for example, nodes such as
  `Gather_604` appeared inside the repeated decoder subfunction.

  GPT-OSS uses both `full_attention` and `sliding_attention`, but the actual full/sliding masks are
  layer-invariant for a given forward call. Only the choice of which mask to pass depends on
  `config.layer_types[layer_idx]` / `self_attn.sliding_window`.

  ## Approach

  The model forward now builds the base causal mask and sliding-window causal mask once, before the
  decoder-layer loop.

  It then prepares reusable additive masks outside the loop:

  - `full_attention_mask` for full-attention layers.
  - `sliding_attention_mask` for sliding-attention layers.
  - For chunked prefill, the same sliding-window index slice is applied once before the loop.

  Inside the decoder-layer loop, the code only selects the already-prepared mask based on the layer’s
  attention type.

  Old in-layer behavior:

  ```python
  torch.where(mask, MIN_MASKED_ATTENTION_VALUE, scores)

  New prepared-mask behavior:

  scores + prepared_mask

  These are equivalent because prepared_mask contains 0 for allowed positions and
  MIN_MASKED_ATTENTION_VALUE for masked positions.

  This follows the same idea as the Qwen2.5 vision optimization: compute reusable masks outside the
  repeated layer/block loop, then pass the selected prepared mask into each layer.


QEfficient/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -681,11 +681,7 @@ def eager_attention_forward(
     attn_weights = torch.matmul(query, key_states.transpose(2, 3)) * scaling
 
     if attention_mask is not None:
-        attn_weights = torch.where(
-            attention_mask,
-            torch.full_like(attn_weights, MIN_MASKED_ATTENTION_VALUE, dtype=attn_weights.dtype),
-            attn_weights,
-        )
+        attn_weights = attn_weights + attention_mask
```
ochougul, line 684: why are we removing torch.where conditions? Not sure if doing sum is numerically stable.

abhishek-singh591: We are not removing the masking condition semantically, we are just moving it out of the repeated attention path.

  Earlier, we did this inside attention:

  ```python
  torch.where(mask, MIN_MASKED_ATTENTION_VALUE, scores)

  Now, we precompute the mask once as an additive mask:

  prepared_mask = torch.where(mask, MIN_MASKED_ATTENTION_VALUE, 0)

  and attention does:

  scores + prepared_mask

  So the behavior is equivalent:

  - Unmasked positions: scores + 0 == scores
  - Masked positions: scores + MIN_MASKED_ATTENTION_VALUE == MIN_MASKED_ATTENTION_VALUE

abhishek-singh591:
CI-Ready

QEfficient/transformers/models/gpt_oss/modeling_gpt_oss.py
ochougul: The `torch.where` condition with -inf values is required by compiler for the softmax to work correctly. This might impact performance, can you put it back.

ochougul: If you want to change this. Please prove performance and accuracy on full model in 20B, 120B models for large context lengths

ochougul (changes requested):
Keep where condition inside attention block, only the boolean mask creation can be outside of the block

ochougul:
CI-Ready

ochougul (approved):

PR #1235, opened 2026-08-04
Fix KV cache handling for Gemma4 26B

nsivalin:
- Fixes sliding-window KV cache corruption in `QEffGemma4DynamicLayer.update()` where physical-slot validity was computed from the wrapped/physical position instead of the logical position, causing attention to read stale/garbage KV data (leading to infinite repetition and non-deterministic output) once a request's sequence length crossed the `sliding_window` boundary. 

- The fix is scoped to `google/gemma-4-26B-A4B-it` only (via `config.num_experts` as the differentiator, leaving the original code path unchanged for all other Gemma4 variants (31B/E2B/E4B).


ochougul:
Please keep this PR in draft unless a maintainer approves it. Then you can make it ready for review.

quic-hemagnih:
Can you fix the LINT issues

quic-rishinr:
closing the PR as change will be added as part of #1250

PR #1242, opened 2026-08-05
Fix: derive dummy_seq_len from the prefill_seq_len kwarg

prabeesh-1:
Change-Id: I4dae49b60a59e5929ba3cce675895caa792d7dfd


ochougul:
Why is this needed? can you please point to the failure case that this code is fixing?
The code LGTM!

prabeesh-1:
image

Observing improper output while running with prefill_seq_len=128 for Qwen3.6-35B-A3B model.

PR #1245, opened 2026-08-06
Added subfunction fix for gemma 4 with CB

abhishek-singh591:
## Summary

  This PR fixes Gemma4 vision continuous-batching compilation when use_onnx_subfunctions=True.

  The QAIC compiler previously failed on the Gemma4 decoder ONNX graph with:

  Range: limit input must be a constant tensor

  The failing Range operation originated in the Gemma4 sliding-window KV-cache update path.

  ## Root Cause

  The sliding-cache context length was derived from the runtime KV-cache tensor shape:

  layer_ctx_len = self.keys.shape[2]

  That value is used to build cache indices with torch.arange(...). With ONNX subfunctions enabled,
  the shape-derived value became a dynamic decoder-layer subfunction input:

  Shape(past_key) -> Gather -> Range limit

  QAIC requires the Range limit inside subfunctions to be a constant, so compilation failed.

  ## Fix

  - Pass cache_kwargs["sliding_window"] = self.sliding_window from Gemma4 sliding-attention layers
    into the cache update path.
  - Store each sliding layer’s configured window length in the Gemma4 dynamic cache layer.
  - Use the static configured sliding-window length for sliding-cache indexing instead of
    self.keys.shape[2].
  - Create sliding-layer dummy KV caches with config.sliding_window for continuous batching, while
    preserving the input-length-bounded cache size for regular batching.
  - Export custom-op context-length inputs as ONNX constants when they originate as Python integers.

  ## Validation

  Validated with:

  /home/abhishek/dynamo/bin/python \
    examples/image_text_to_text/models/gemma_vision/gemma4/gemma4_cb.py

  The continuous-batching example completed successfully with use_onnx_subfunctions=True.


quic-hemagnih:
@abhishek-singh591  Can you run gemma4_diss.py script and ensure that perf for Gemma4-26B-A4B-IT model is not impacted with these changes.

QEfficient/customop/ctx_scatter_gather.py
```diff
@@ -247,6 +247,8 @@ def setup_context(ctx, inputs, outputs):
 
     @staticmethod
     def symbolic(g: torch.Graph, data: torch.Value, ctx_indices: torch.Value, comp_ctx_len: int) -> torch.Value:
+        if not isinstance(comp_ctx_len, torch.Value):
+            comp_ctx_len = g.op("Constant", value_t=torch.tensor(comp_ctx_len, dtype=torch.int64))
```
ochougul, line 251: why are you changing this, seems unreleated.

QEfficient/customop/ctx_scatter_gather_cb.py
```diff
@@ -144,6 +144,8 @@ def setup_context(ctx, inputs, outputs):
     def symbolic(
         g: torch.Graph, data: torch.Value, batch_index: torch.Value, ctx_indices: torch.Value, comp_ctx_len: int
     ) -> torch.Value:
+        if not isinstance(comp_ctx_len, torch.Value):
+            comp_ctx_len = g.op("Constant", value_t=torch.tensor(comp_ctx_len, dtype=torch.int64))
```
ochougul, line 148: same comment

QEfficient/transformers/models/gemma4/modeling_gemma4.py
```diff
@@ -584,6 +584,8 @@ def forward(
         input_shape = hidden_states.shape[:-1]
         hidden_shape = (*input_shape, -1, self.head_dim)
         cache_kwargs = {"position_ids": position_ids, "batch_index": batch_index}
+        if self.sliding_window is not None:
+            cache_kwargs["sliding_window"] = self.sliding_window
```
ochougul: No need to add it here, if we have already saved it as constant in the cache class.

ochougul (changes requested):

ochougul:
CI-Ready

QEfficient/customop/ctx_scatter_gather.py
abhishek-singh591: This is needed because the static sliding_window now reaches the symbolic as a Python int (for example, 512) rather than a torch.Value. CtxGatherFuncCB expects an ONNX INT64 graph input, and without materializing it as a Constant, export fails with Node.z_() receiving value=512.

QEfficient/customop/ctx_scatter_gather_cb.py
abhishek-singh591: Same as above.

QEfficient/transformers/models/gemma4/modeling_gemma4.py
abhishek-singh591: Yes not needed, will remove it.

QEfficient/customop/ctx_scatter_gather.py
abhishek-singh591: We could also handle this in cache_utils.py by converting the static sliding-window value to a 1-D torch.Tensor before calling min(), so it reaches the custom gather op as a graph value.

ochougul: yeah I would prefer it to have fixed the second way you suggested than changing the custom op implementation. Unless you tell me that the ONNX graph doesn't change at all with this change.

ochougul (changes requested):
Lets always keep sliding_window_len to be a constant and keep same in get_dummy_cache.
And verify if we test CB, non-cb with subfunc in our tests for this model. if not lets add the test

abhishek-singh591:
> Lets always keep sliding_window_len to be a constant and keep same in get_dummy_cache. And verify if we test CB, non-cb with subfunc in our tests for this model. if not lets add the test

Fixed the issue.

abhishek-singh591:
Without ONNX subfunctions, the decoder is exported as one graph, so `self.keys.shape[2]` stays within that complete graph.
  With ONNX subfunctions, each decoder layer receives `past_key` as an input; `layer_ctx_len = self.keys.shape[2]` becomes `Shape(past_key) → Gather(axis=2)`.
  Using `layer_ctx_len` in `torch.arange(...)` produces `Shape(past_key) → Gather → Range(limit)`, and QAIC rejects this dynamic `Range` limit.

ochougul (approved):

ochougul:
CI-Ready

ochougul:
CI-Ready

PR #1249, opened 2026-08-07
Gemma4 Moe and Prefill changes

tchawada:
Optimized the moe MLP block by processing only the top-K selected experts for each token. The implementation gathers expert-specific projection weights (gate, up, and down), performs batched matrix multiplications across all (token, expert) pairs, applies the gated activation (act_fn(gate) * up), and combines expert outputs using router-provided top-K weights.

NSP-parallel expert-blocked dispatch to the chunked prefill MoE path, replacing the sequential per-expert loop with a batched packed-prefix approach.


quic-hemagnih:
This code is already merged in the release branch and now we can close this PR.

PR #1229, opened 2026-07-31
Head parallel and batch fold blocking for Qwen3-VL MOE

kdulla:
# Overview
This PR adds support for new blocking modes, including head parallel blocking, prefill blocking, decode MOE blocking and batch-fold blocking. Head parallel blocking is introduced for causal models while the other approaches are introuced for Qwen3-VL MOE. Expert Parallelism is enabled via expert_parallel compile option while new flags in qaic_config can be passed for different attention blocking configurations. Additionally, this PR contains updated tests for the new blocking modes introduced. 

# Modes Added
- Head parallel softmax kv blocking:
  - split heads across cores, enabled for both causal models and Qwen3VL-moe
  - Enabled by passing kv_blocking_headpar_split (Default: 0 or 16) in qaic config with kv blocking mode
  - invokes newly created blocked_kv_attention_forward_headpar_offline attention forward
- Batch fold + head parallel kv blocking:
  - fold batch size and head dimensions before applying head parallel dimension, intended for high batch size use cases, enabled for Qwen3-VL-moe
  - Enabled by passing batch_fold: True in qaic config with kv blocking mode
  - invokes newly created blocked_kv_attention_forward_decode_headpar_batch attention forward
  - Alters cache layout to be [1, BS * NH, SL, D]
- Online softmax prefill:
  - Applies online softmax to chunked prefill blocks, intended for long-context prefill efficiency
  - Enabled by passing prefill_blocking_mode: "online", prefill_block_chunks, and prefill_n_rep_chunk in qaic_config
  - Invokes newly created blocked_qkv_attention_forward_prefill_online attention forward

Additionally, pre-existing KV blocking modes have been updated to correctly use ctx_len from compile_time for computing block sizes instead of past_seen-tokens. 

# Scripts
E2E run on Qwen3-VL moe with batch fold and prefill blocking can be ran via:
`python examples/image_text_to_text/models/qwen3_vl_moe/qwen3_vl_disagg_mode.py`

Unit tests on Qwen3-VL moe can be ran via:
`pytest tests/transformers/models/image_text_to_text/test_qwen3_vl_moe_disagg_blocked.py`

Unit tests for head parallel softmax on causal models added to tests/transformers/models/causal_lm_models/test_causal_lm_blocking_hqkv.py

PR #1251, opened 2026-08-08
Gemma 4 blocking support

tv-karthikeya:
To enabling Gemma4 blocked disaggregated prefill/decode example

 **Summary**
  This PR updates the Gemma4 vision blocked-disaggregated example to exercise separate optimized prefill and decode QPCs with long-context blocked attention. The script now uses QKV blocked prefill, head-parallel KV blocked decode, retained KV handoff between prefill and decode, and avoids copying large retained decode state back to host every token.

  **Blocking Changes**

  - Enables blocked attention with:

    ENABLE_BLOCKING = True
    ENABLE_HEAPAR = True

  - Uses separate KV block counts for prefill and decode:

    PREFILL_NUM_KV_BLOCKS = 16
    DECODE_NUM_KV_BLOCKS = 8

  - Uses QKV prefill blocking:

    PREFILL_MODE = "qkv"
    PREFILL_BLOCK_CHUNKS = ceil(PREFILL_SEQ_LEN / 128)
    PREFILL_N_REP_CHUNK = 1

    This enables head-parallel/offline QKV prefill with query chunks of 128 tokens.

  - Uses head-parallel split auto-resolution:

    KV_BLOCKING_HEADPAR_SPLIT = 0 # 0 lets the compile path resolve the split from available cores.

  - Enables KV skip optimization:

    "skip_kv": True

    This allows blocked attention to skip future KV blocks that cannot contribute to the current token.

  **Runtime Optimization**

  - Skips retained decode outputs:

    lang_decode_session.skip_buffers(decode_retained_outputs)
    This avoids copying large *_RetainedState buffers, especially past_key and past_value, back to host every decode token.
    This avoids copying large *_RetainedState buffers, especially past_key and past_value, back to host every decode token.

  - Decode loop no longer feeds updated KV back each step. After the first decode call, it only updates:

    input_ids
    position_ids

    The retained state is expected to remain resident in the QAIC session.


vbaddi:
is there a need to add batch split/fold blocking to these models? @quic-hemagnih

quic-hemagnih:
Batch fold is not needed for this model

tv-karthikeya:
CI-Ready

QEfficient/blocking/blocked_attention_forwards.py
```diff
@@ -208,6 +247,776 @@ def blocked_kv_attention_forward(
...
+            start_index=start_index,
+        )
+        causal_mask = (
+            causal_mask.expand(batch_size, num_kv_heads, seq_len, kv_len_block)
+            .reshape(1, BH, seq_len, kv_len_block)
+            .unsqueeze(3)
+            .expand(1, BH, seq_len, num_kv_groups, kv_len_block)
+            .reshape(1, BH, seq_len * num_kv_groups, kv_len_block)
+        )
+        attn_weights_block = attn_weights_block.masked_fill(causal_mask, -3.0e4)
```
ochougul, line 325: Please use `MIN_MASKED_ATTENTION_VALUE` here instead of the hardcoded `-3.0e4`. Same for lines 330, 433, 454, 572, 882 and 884. The rest of this file (including the forwards right above and below the new ones) already uses the constant everywhere.

Apart from consistency, mixing the two values is a correctness problem. The running max is seeded with `MIN_MASKED_ATTENTION_VALUE` i.e. `-inf`, but the masked logits get `-3.0e4`. So for a row where the whole block is masked, the block max ends up at `-3.0e4` instead of `-inf`, `exp(logit - max)` becomes 1 for the masked positions instead of 0, and those ones get accumulated into `split_sum`/`split_out` in the two-stage merge.

Also at line 572 please pick up `mask_dtype` the way `blocked_kv_attention_forward` does rather than `query.dtype`, and keep the accumulator seed and the mask fill the same value within each forward.

If `-3.0e4` was picked deliberately because `-inf` was producing nan/inf on device, let me know - in that case it should be a named constant and the seeds should move to match it, not left as a literal.

(Same comment as https://github.com/quic/efficient-transformers/pull/1229#discussion_r3761595669 - reposting here since this is the branch being merged.)

QEfficient/base/modeling_qeff.py
```diff
@@ -667,13 +667,12 @@ def get_onnx_path(
             kwargs.update(
                 {
                     "prefill_only": prefill_only,
-                    "prefill_seq_len": constants.ONNX_EXPORT_EXAMPLE_SEQ_LEN,
+                    "prefill_seq_len": specializations[0].get("seq_len"),
```
vbaddi, line 670: nit: why do we want to do this? this might spike up the RAM, when we want compile for CPL 1K and all right?

vbaddi (changes requested):

ochougul:
@quic-hemagnih Pls do not merge this PR directly to main. We want to merge this directly into release branch, and merge to main after proper review.

quic-hemagnih:
Blocking code is merged in the mainline in the PR: https://github.com/quic/efficient-transformers/pull/1308/ closing this PR now.

PR #1253, opened 2026-08-09
chore: update container setup and development environment

talnufai:
- Switch to an Ubuntu version compatible with Python 3.12
- Add Vim
- Add torchvision installation
- Add `sleep infinity` to allow exec access into the container
- Add `docker-compose.yml` with a synced local volume
- Configure `.env` as the environment file
- Add `example.env`

---------


quic-rishinr:
@talnufai why you want this change to be merged on Qeff?

talnufai:
> @talnufai why you want this change to be merged on Qeff?

Hi @quic-rishinr , the current dockerfile doesn’t work. The dockerfile can help with setting up the environment quickly

PR #1256, opened 2026-08-10
Consolidate text_generation examples into a single flag-driven basic_inference.py

quic-rishinr:
Replace the per-feature example scripts (gguf_models, moe_inference, continuous_batching, blocked_attention_inference, batch_blocking_example, gpt_oss_blocked_example, cli_examples) with one canonical entry point, examples/text_generation/basic_inference.py, where every recipe is a flag over a real QEFFAutoModelForCausalLM.from_pretrained / .compile / .generate kwarg. Shared argument parsing lives in examples/_common/args.py.

Move the two bespoke recipes (GLM4 KV-head surgery, Kimi-K2 MLA absorption) under examples/text_generation/advanced/ with a README describing their scratchpad nature.

Fix parser/wiring defects found in review:
- emit num_kv_heads_repeat at the top level of build_qaic_config so it is not silently dropped without --enable-blocking
- reject --num-devices/--device-group mismatch and empty --device-group at the CLI instead of failing deep in the runtime
- surface --include-sampler in basic --help
- wire --max-seq-len-cached, --layerwise, and --layerwise-window-size, forwarding --layerwise to from_pretrained for the meta-device load path
- evaluate advanced-help visibility lazily at parser-build time

Repoint examples/README.md and the docs pages that linked to the deleted scripts at basic_inference.py.


ochougul:
CI-Ready

PR #1259, opened 2026-08-10
feat(0810): Add continuous batching support to batch-folded decode attn implementation

vbaddi:


vbaddi:
Closing, this is not the version, we will be working on for `batch_split` mode.

PR #1260, opened 2026-08-10
fix(qwen3_5_moe): restore original chunk-attn solver to recover accuracy

prabeesh-1:
Revert chunk_gated_delta_solver to "original" in QEffQwen3_5MoeGatedDeltaNet. __qeff_init__ to restore row-by-row forward substitution, which isolates rounding to one row at a time and does not propagate errors across iterations.

Change-Id: I7521a80b6171258f94decfd91a67145bba744a72


vbaddi:
@prabeesh-1 whats the reason for this change? also pls remove change ID for qeff github pushes.

prabeesh-1:
> @prabeesh-1 whats the reason for this change? also pls remove change ID for qeff github pushes.

We see accuracy deviation of 1.8% with the new solver. I tried with original solver and recover back the accuracy. Hence suggesting the change for merge.

anujgupt-github:
Not planned to be merged soon. Original path with matrix inversion being optimized

PR #1263, opened 2026-08-11
Fix weight-free ONNX cache reuse and add CCL test coverage; enable dynamo export for phi3/gpt-oss

amarquic:
When use_weight_free_export=True the model has no loaded weights, so re-using a previously stored onnx_path would silently produce a mismatched graph. Guard the early-return with the new flag so a fresh export is always triggered for weight-free compilations.


QEfficient/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -1061,9 +1061,7 @@ def forward(
 
         if output_attentions:
             outputs += (self_attn_weights,)
-
-        if use_cache:
-            outputs += (present_key_value,)
+        # not returning present_key_value as it is not used
```
vbaddi, line 1065: nit: why is this required? any issues if we keep this same?

QEfficient/transformers/cache_utils.py
```diff
@@ -1206,12 +1206,8 @@ def from_legacy_cache(
         cache = cls(
             config,
             batch_size=past_key_values[0][0].shape[0],
-            max_cache_len=int(past_key_values[1][0].shape[2])
-            if torch._dynamo.is_compiling()
-            else past_key_values[1][0].shape[2],
-            sliding_window_len=int(past_key_values[0][0].shape[2])
-            if torch._dynamo.is_compiling()
-            else past_key_values[0][0].shape[2],
+            max_cache_len=past_key_values[1][0].shape[2],
```
vbaddi: nit: why is this changed?

tests/dynamo/_helpers.py
```diff
@@ -61,7 +61,7 @@
     "mpt": "hf-internal-testing/tiny-random-MptForCausalLM",
     "olmo2": "hf-internal-testing/tiny-random-Olmo2ForCausalLM",
     "phi": "hf-internal-testing/tiny-random-PhiForCausalLM",
-    # "phi3": "tiny-random/phi-4", #TODO: need to fix the SplitToSequence issue
+    "phi3": "tiny-random/phi-4",
```
quic-amitraj, line 64: Is this issue fixed now?

tests/dynamo/test_on_qaic.py
```diff
@@ -73,10 +82,12 @@ def test_dynamo_fp16_compile(model_type, model_id, tmp_export_dir):
         num_cores=16,
         batch_size=BATCH_SIZE,
         use_onnx_subfunctions=True,
+        dynamo=True,
```
quic-amitraj, line 84: Is this was not using the dynamo in test previously?

quic-amitraj (changes requested):
Rename the PR, its not refecting the code changes.

QEfficient/transformers/models/gpt_oss/modeling_gpt_oss.py
smedhe: these changes should have been part of the TF upgrade and also because the returned value is a Cache object we are getting error in dynamo export. This change is required.

QEfficient/transformers/cache_utils.py
smedhe: This change was added to make dynamo work in the last commit but this int() causes sliding_window and ctx_len to get baked into the graph and so the resulting graph was static in nature.

tests/dynamo/_helpers.py
smedhe: Yes, enabled dynamo support also, the previous modeling changes are not present in this code.

tests/dynamo/test_on_qaic.py
smedhe: just making the flag explicit, as we have exported with dynamo and we are passing the same onnx path to compile, it was using dynamo earlier as well

QEfficient/base/modeling_qeff.py
```diff
@@ -987,7 +987,7 @@ def _compile(
             # ONNX because re-exporting is no longer possible. Otherwise export for
             # the current compile mode, e.g. decode vs. disaggregated prefill.
             weights_offloaded = self._is_weights_offloaded or any(param.is_meta for param in self.model.parameters())
-            if self.onnx_path is not None and weights_offloaded:
+            if self.onnx_path is not None and weights_offloaded and not use_weight_free_export:
```
quic-amitraj, line 991: Add a TODO here that, we have to utilize all the transforms like `PrefillOnlyExternalModuleMapperTransform` in hash creation so that it will never pick same onnx file for prefill and decode.

quic-amitraj (changes requested):

QEfficient/base/modeling_qeff.py
amarquic: Done

PR #1284, opened 2026-08-25
Weight-free export comments addressed 

quic-amitraj:
## Summary
  - Moves CausalLM weight-free export enablement to `QEFFAutoModelForCausalLM.from_pretrained(..., weight_free=True)`.
  - Removes the need to pass `use_weight_free_export` through `export()` and `compile()`; those paths now derive weight-free mode from the QEff model instance.
  - Builds weight-free models directly with `accelerate.init_empty_weights()`, avoiding an unnecessary real-weight load before meta export.
  - Preserves `attn_implementation` and keeps `config.dtype` / `config.torch_dtype` aligned with the resolved dtype for meta model construction.
  - Adds `QEFF_CHECKPOINT_HOME` support for prepared weight-free checkpoints and shortens the prepared checkpoint suffix to `-wf-<dtype>`.
  - Deduplicates `TORCH_INVOKE_ALLOW_CREATE_FALLBACK` handling with a shared context manager.
  - Updates examples and weight-free tests to use the new API.


QEfficient/exporter/weight_free/export.py
```diff
@@ -49,8 +37,24 @@ def _to_meta(value: Any) -> Any:
     return value
 
 
+def _resolve_weight_free_target_dtype(model_dtype: Optional[torch.dtype]) -> torch.dtype:
```
quic-amitraj: No need

quic-amitraj:
Addressed in #1290.

PR #1293, opened 2026-08-31
Fix Gemma4 RoPE handling

nsivalin:
Gemma4 was the only QEff model still using HF’s runtime rotary embeddings, recomputing RoPE cos/sin via `Cos`/`Sin` ops on every forward pass.  This caused the computations to be repeatedly traced into the exported ONNX graph across decoder and vision-encoder layers, triggering a compiler tensor-aliasing issue.  
The fix introduces `QEffGemma4TextRotaryEmbedding` and `QEffGemma4VisionRotaryEmbedding`, which precompute static cos/sin tables during initialization.  As a result, `forward()` now performs a simple `Gather`, aligning Gemma4 with the implementation pattern used by all other QEff models and eliminating the decode corruption issue.


vbaddi:
@ochougul @quic-rishinr do we need this?

PR #1295, opened 2026-08-31
fix(gemma3): Fix OCR regression caused by RoPE ONNX aliasing

prabeesh-1:


ochougul:
being fixed at #1296

quic-hemagnih (approved):
@ochougul  Can we prioritize this for CI, once its passed we can merge it

PR #1302, opened 2026-09-02
Fix GPT-OSS blocked prefill cache reads and head-par prefill blocking

smedhe:
## Summary

  - Fix GPT-OSS blocked cache read helpers to use the owning layer cache tensors directly.
  - Avoid duplicate cache updates during blocked prefill for non-sliding GPT-OSS attention.
  - Default n_rep_chunk to 1 in attention blocking config.
  - Update the GPT-OSS blocked prefill head-par example to handle logits shaped [B, V], [B, 1, V], or [B, S, V].
  - Pass MoE prefill packed chunk sizing through the attention blocking config.
  - Add unit coverage for blocked K/V cache read dispatch.


examples/text_generation/gpt_oss_blocked_prefill_headpar_example.py
```diff
@@ -208,6 +223,7 @@ def main():
         "num_kv_blocks": 2,
         "num_q_blocks": 2,
         "ctx_len": args.ctx_len,
+        "moe_config": {"expert_prefill_chunk_size": args.moe_prefill_packed_chunk_size},
```
kdulla, line 226: should be expert_parallel_chunk_size

QEfficient/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -645,6 +645,9 @@ def forward(
         value_states = self.v_proj(hidden_states).view(hidden_shape).transpose(1, 2)
         query_states, key_states = qeff_apply_rotary_pos_emb(query_states, key_states, cos_cached, sin_cached)
 
+        blocking_config = getattr(self, "attn_blocking_config", AttentionBlockingConfig())
+        use_blocking = blocking_config is not None and blocking_config.mode.is_prefill and (self.sliding_window is None)
```
kdulla, line 649: we should raise an error if we do a blocking mode that is not prefill specific, not just skip it like this code does.

kdulla (changes requested):

examples/text_generation/batch_blocking_example.py
```diff
@@ -68,19 +93,19 @@ def main():
     # setup qaic config to enable blocking, ensure 4 or more device ids are passed
     qaic_config = {
         "blocking_mode": args.blocking_mode,
-        "num_batch_blocks": 2,
-        "num_kv_blocks": 2,
-        "num_q_blocks": 2,
-        "head_block_size": 8,
+        "num_batch_blocks": args.num_batch_blocks,
```
kdulla: with the most recent updates to the interface, this would fail for modes except hqkv and bhqkv - when passing number of blocks via qaic_config, it needs to exactly match the blocking mod expected

examples/text_generation/gpt_oss_blocked_example.py
```diff
@@ -8,11 +8,41 @@
...
+import numpy as np
+from transformers import AutoConfig, AutoTokenizer
 
 from QEfficient import QEFFAutoModelForCausalLM
 
 
+def build_blocking_config(args):
+    blocking_mode = args.blocking_mode.lower()
+    qaic_config = {
+        "blocking_mode": blocking_mode,
```
kdulla: see comment above

kdulla (changes requested):
example files need an update but otherwise looks good

kdulla:
gpt-oss updated logic looks right and the tests also look correct so if those pass, the changes are functionally correct

examples/text_generation/batch_blocking_example.py
smedhe: Done

examples/text_generation/gpt_oss_blocked_example.py
smedhe: Done

smedhe:
CI-Ready

PR #1312, opened 2026-09-08
Wire QEffGptOssMLP into shared MoEBlockMixin

Shrubabati7:
- Aligned gpt_oss MoE with MoE ubench implemetation
- Using common MoE transforms class now in gpt_oss
- TTFT for 2L, ctx-len 8192, prefill-seq-len 512, moe-prefill-packed-chunk-size 256: 1.10sec in SDK 1.23.0.41 assert


QEfficient/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -286,164 +112,12 @@ def moe_profile(self) -> MoEProfile:
...
-            torch.nn.functional.one_hot(selected_experts[:, 0], num_classes=self.experts.num_experts)
-            .bool()
-            .T.unsqueeze(-1)
-        )
-        awesome_experts_mask_2 = (
-            torch.nn.functional.one_hot(selected_experts[:, 1], num_classes=self.experts.num_experts)
-            .bool()
-            .T.unsqueeze(-1)
-        )
-        awesome_experts_mask_3 = (
```
abhishek-singh591, line 374: This removes the only GPT-OSS runtime use of NUM_FFN_BLOCKS: the shared MoE path does not consume it, and the new MoE config options are not a compatibility mapping for its token-blocking behavior.
 
If removing this legacy control is intentional, please also remove its stale handling from modeling_auto.py (export sizing/hash) and update examples/disagg_serving/README.md and if backward compatibility is needed put the warning as well.

QEfficient/transformers/models/pytorch_transforms.py
```diff
@@ -1661,6 +1661,10 @@ def apply(
                 module.tree_reduce = tree_reduce
                 module.expert_parallel_num_packed_chunks = num_packed_chunks
                 module.expert_blocking_packed_chunk_size = expert_parallel_chunk_size
+                expert_intermediate_block_size = moe_config.get("expert_intermediate_block_size", None)
+                if expert_intermediate_block_size is not None:
+                    expert_intermediate_block_size = int(expert_intermediate_block_size)
```
abhishek-singh591, line 1666: We can put a check for negative values and reject non-positive values, as this value is passed through int() before being used as the tile-loop step. This is mainly for robustness.

QEfficient/transformers/moe/flavours.py
```diff
@@ -114,16 +114,24 @@ def cumsum_scatter_gather_update_expert_blocked(
...
     batch_size, seq_len = T2Ei.shape
     num_packed_chunks = max(1, int(num_packed_chunks))
     assert seq_len % num_packed_chunks == 0, (
         f"seq_len={seq_len} must be divisible by num_packed_chunks={num_packed_chunks}"
     )
-    packed_chunk_size = seq_len // num_packed_chunks
+    packed_chunk_size = max(1, min(seq_len // num_packed_chunks, seq_len))
     matched_idx = build_matched_idx_from_cumsum(T2Ei)
-    valid_rows = torch.einsum("ij->i", T2Ei.to(torch.int32)).unsqueeze(1)
+    valid_rows = T2Ei.to(torch.int32).sum(dim=1, keepdim=True)
```
abhishek-singh591: .sum() will fail on current mainline, did you validate this?

QEfficient/transformers/moe/flavours.py
```diff
@@ -114,16 +114,24 @@ def cumsum_scatter_gather_update_expert_blocked(
...
+    to ``[N, chunk, expert_intermediate_block_size]`` — matching the microbench
+    ``--ffn-weight-block-size`` behaviour.
+    """
     batch_size, seq_len = T2Ei.shape
     num_packed_chunks = max(1, int(num_packed_chunks))
     assert seq_len % num_packed_chunks == 0, (
         f"seq_len={seq_len} must be divisible by num_packed_chunks={num_packed_chunks}"
     )
-    packed_chunk_size = seq_len // num_packed_chunks
+    packed_chunk_size = max(1, min(seq_len // num_packed_chunks, seq_len))
```
abhishek-singh591: We don't need this pls remove, since num_packed_chunks is computed during the MoE transform and the export sequence length is subsequently rounded to a multiple of it in modeling_auto.py.

QEfficient/transformers/models/gpt_oss/modeling_gpt_oss.py
Shrubabati7: Restored: QEffGptOssMLP.execute_moe_flavour now overrides the flavour dispatch unconditionally when NUM_FFN_BLOCKS is set (matching the old mixin behaviour), modeling_auto.py hash/sizing logic retained with docstring, and README.md updated to reflect the token-blocking semantics.

QEfficient/transformers/models/pytorch_transforms.py
Shrubabati7: Change done: added if expert_intermediate_block_size <= 0: raise  ValueError(...) immediately after the int() cast in OptimizedMoEExportConfigTransform.

abhishek-singh591 (approved):

vbaddi:
CI-Ready

PR #1314, opened 2026-09-08
Enable Dynamo KV-blocking Loop export for kv_headpar

quic-abhamidi:
Integrate a Dynamo-only torch.while_loop implementation for blocked_kv_attention_forward_headpar_offline so KV-blocked decode can export a structured ONNX Loop instead of unrolling the Python for-loop per KV block.

  Key changes:
  - Enable the KV Loop path by default for blocking_mode=kv_headpar when dynamo=True.
  - Preserve the legacy Python-loop path for dynamo=False and allow explicit opt-out with use_kv_loop_op=False.
  - Add StaticLoopInputsTransform to rewrite ONNX Loop trip count and initial condition to static constants required by QAIC compiler unrolling.
  - Add InlineTorchSubgraphFunctionsTransform for dynamo subfunction export, inlining torch-generated subgraph functions that contain Loop nodes so QAIC does not see Loop control inputs inside FunctionProto bodies.
  - Update Dynamo causal LM examples and README with KV-blocking CLI support.
  - Add Dynamo/unit tests covering Loop export, config defaults, static Loop
    input rewriting, and function-body Loop handling.

  Verified:
  - tests/dynamo/test_kv_blocking_loop.py: 4 passed
  - tests/unit_test/transforms/test_blocking_transform.py::TestDynamoKVLoopBlockingConfig: 3 passed
  - Qwen2-1.5B kv_headpar Dynamo compile succeeds with and without
    --enable-onnx-subfunctions.

### DIFF
--- a/QEfficient/base/modeling_qeff.py
+++ b/QEfficient/base/modeling_qeff.py
@@ -25,8 +25,10 @@ from QEfficient.base.onnx_transforms import (
     BaseOnnxTransform,
     CustomOpTransform,
     FP16ClipTransform,
+    InlineTorchSubgraphFunctionsTransform,
     OnnxTransformPipeline,
     RenameFunctionOutputsTransform,
+    StaticLoopInputsTransform,
     SplitTensorsTransform,
 )
 from QEfficient.base.pytorch_transforms import PytorchTransform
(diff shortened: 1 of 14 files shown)


jinkasaisagar:
Rebase required?

PR #1294, opened 2026-08-31
feat(0831): Localize constant ReduceSum axes inside ONNX subfunctions

vbaddi:
## Summary

  - Added `LocalizeFunctionReduceSumAxesTransform` for ONNX subfunction exports.
  - Replaced the previous `torch.einsum(... reduction ...)` workarounds back to equivalent `.sum(...)` forms across QEff modeling/MoE/blocking code.
  - Enabled the ReduceSum axes localization only when `use_onnx_subfunctions=True`.

  We previously used `einsum` in several reduction patterns to avoid ONNX subfunction export promoting constant `ReduceSum` axes into `FunctionProto` inputs. That workaround avoided compiler failures, but it could lower into heavier performance dip's due to lowering of BMM instead of the intended elementwise multiply plus ReduceSum path.

  The desired PyTorch source should be able to stay as:

  ```python
  (query * query).sum(dim=-1, keepdim=True)
  and export as:
  Mul -> ReduceSum <- Constant([-1]) -- inside the ONNX function body.
  ```

  ## ONNX Transform Details

  LocalizeFunctionReduceSumAxesTransform it rewrites a function input when:

  - a node inside an ONNX FunctionProto is ReduceSum
  - the ReduceSum axes input is one of the function formal inputs
  - every top-level call site passes a compile-time constant for that formal input
  - the value is a valid integer scalar or 1-D axes tensor

  When eligible, the transform:

  - inserts a local Constant node inside the function body
  - rewires each matching ReduceSum to use that local constant
  - removes the axes formal input from the function signature

PR #1328, opened 2026-09-11
[Fix] removing continues_batching  check for qpc key selection in modeling_auto

tchawada:
Issue: Decode Compilation Fails with Missing ONNX External Data File Error, current behavior picks qpc_key = "lang_qpc_path" in place of "lang_decode_qpc_path" under continuous batching, which breaks vllm_qaic


vbaddi:
@tchawada this should be addressed in #1294 Can you check?
cc: @quic-hemagnih

tchawada:
Addressed the same in PR 1337

PR #1338, opened 2026-09-14
Qwen3 vl moe causal mask subfunction

quic-rishinr:
Moves Qwen3-VL MoE causal-mask construction into each decoder attention subfunction.


vbaddi:
@quic-rishinr 
The PR moves the causal-mask computation inside the decoder attention subfunction, which matches the compiler recommendation. However, for the CB + batch-folded path, we should re-export and compile once to confirm that:

- [ ] The extra ONNX function inputs/outputs are removed.
- [ ] Decode performance returns close to the non-subfunction baseline.

cc: @tv-karthikeya

PR #1339, opened 2026-09-15
Add ONNX subfunction dedupe DeduplicateRepeatedSubgraphTransform for repeated Dynamo decoder layers

smedhe:
## Summary
  - Add `DeduplicateRepeatedSubgraphTransform` to collapse structurally identical Dynamo-exported
  decoder layer subfunctions.
  - Wire the dedupe pass into the ONNX transform pipeline after repeated subgraph renaming.
  - Canonicalize layer-local static single assignment names and single-dimension `Shape`/`Squeeze` outputs so
  equivalent decoder layers with symbolic shape temp-name differences can reuse one FunctionProto.
  - Preserve distinct subfunctions for structurally different layer groups, such as mixed sparse/
  dense decoder layers.

  ## Motivation
  Dynamo repeated-subgraph export can emit one ONNX local function per decoder layer even when
  multiple layers have identical bodies. This bloats the exported ONNX, and defeats the intended reusable-subfunction path. For mixed-layer models, the desired output is one subfunction per unique decoder-layer structure
  rather than one subfunction per layer.


smedhe:
check the hashing tests in the legacy test suite, add this transform if they fail

QEfficient/base/onnx_transforms.py
```diff
@@ -517,6 +522,293 @@ def apply(cls, model: ModelProto, target_classnames: Optional[List[str]] = None,
...
+    position, so duplicate call nodes can safely target the first structurally
+    equivalent function without changing graph-level tensor names.
+    """
+
+    _NUMERIC_SUFFIX_RE = re.compile(r"^(?P<base>.+)_(?P<idx>\d+)$")
+
+    @classmethod
+    def apply(cls, model: ModelProto, target_classnames: Optional[List[str]] = None, **kwargs) -> bool:
+        target_classnames = [name for name in (target_classnames or []) if name]
+        candidates = cls._collect_candidate_functions(model, target_classnames)
```
jinkasaisagar: _collect_candidate_functions() has recursive calls. Can this be optimized?

PR #1349, opened 2026-09-17
fix(dynamo): preserve subfunction source metadata

vbaddi:
PyTorch loses original FX source metadata when Dynamo retraces GraphModule subfunctions through invoke_subgraph. This causes ONNX function nodes to reference generated tracing locations instead of the original QEff model files and source lines.

Add a process-scoped monkey patch in torch_patches that retraces **GraphModule** subfunctions through torch.fx.Interpreter under preserve_node_meta(). The original PyTorch function is restored after export, and concurrent exports are serialized because the patched API is process global.

Apply the workaround to both regular Dynamo export and weight-free Dynamo export paths.

**Validated with:**
- tests/dynamo/test_transforms.py: 13 passed
- Normal Dynamo tiny Qwen2 export, compile, and inference
- Weight-free Dynamo tiny Qwen2 export, compile, and inference
- ONNX function metadata containing original QEff source locations


smedhe (approved):
LGTM

quic-amitraj (approved):
LGTM

